PR #303, opened 2014-08-19
workaround for intermittent failures of tcp_shutdown test on Windows

TTimo:
This gross hack avoids the tcp_shutdown test failure on Windows. It also fixes an abort during shutdown that we encountered in our internal product while using the ipc transport (our internal bug had lower repro, but wasn't as pathological as the tcp_shutdown test - still, it was the same issue).

I'm not suggesting that this be merged, but I'm hoping someone can provide some insight as to what is going on (e.g. what is a non NN_SOCK_SRC_EP doing here) and point towards a better fix.

### DIFF
--- a/src/core/sock.c
+++ b/src/core/sock.c
@@ -825,11 +825,20 @@ static void nn_sock_shutdown (struct nn_fsm *self, int src, int type,
     if (nn_slow (sock->state == NN_SOCK_STATE_STOPPING_EPS)) {
 
         /*  Endpoint is stopped. Now we can safely deallocate it. */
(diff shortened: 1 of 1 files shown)


TTimo:
I may have closed this one by mistake. It's still an issue / fix which we have in our prod code.

wirebirdlabs:
This PR might effectively be handled by https://github.com/nanomsg/nanomsg/commit/7555f49e83ff414133f77e46389f8e5632eeea1f -- @TTimo , what do you think?

gdamore:
Yep, I've resolved this a different way altogether.  Closing this one.

Issue #283, opened 2014-07-15
Windows IPC pub/sub: messages silently dropped if sent too fast


TTimo:
I've built a test case in my repository for now:
https://github.com/TTimo/nanomsg/blob/0.4-beta/tests/ipc_flood.c

Sending 10 messages in quick succession, then trying to read them back.

I've put a wait before the last one, reading back will see the first message and the last one, but nothing in between.

Started tracing this a little bit. After the first message, nn_dist_send will early out in the "In the specific case when there are no outbound pipes. There's nowhere to send the message to. Deallocate it" part. No error is reported, message is silently dropped but API reports a good send.

I ran it on OSX as well, no messages are dropped and the test passes.

Any suggestions on what to investigate next?

gdamore:
This is a non-issue, except perhaps a misunderstanding of PUB/SUB.  This is a best effort delivery protocol, and if you flood pub faster than sub can consume, you're going to drop messages.  Frankly, its not possible to "fix" this without creating a situation where one bad / slow subscriber could create stalls for everyone else in the network.

gdamore:
CLosing this as its a misunderstanding/mis-expectation of PUB/SUB.  If you flood, you overrun queues, and messages get dropped.  Either don't flood, or use some other out of band protocol to cope with this.  PUB/SUB cannot, and will not.

TTimo:
So that makes the protocol 'unreliable' then. It should at least return error codes when messages are getting dropped. I'm not sure how people hope to ever use this in production as is.

gdamore:
Indeed, the protocol is *unreliable*.  Just like  UDP.  Its entirely best effort.  If you have a requirement for reliable delivery, you need to use REQ/REP, or you have to build another sideband protocol.

Most people use PUB/SUB as a broadcast delivery of updates, and at fairly moderate rates.  PUB/SUB is not for multicasting at high bandwidth, and indeed there is no such thing (really) as reliable multicast or broadcast either.

PR #320, opened 2014-10-22
test and fix for issue 283 - windows ipc pub/sub dropping messages

TTimo:
test and fix for issue 283 - windows ipc pub/sub dropping messages if sent to fast
see https://github.com/nanomsg/nanomsg/issues/283


djc:
Any progress/further insight here?

TTimo:
That was the last response from @sustrik via email:

" Let's add "int txbuf" member to nn_usock. It will be initialised to 0 and incremented with every send operation by the number of bytes sent. Also, when send completes (usock_win.inc:852) it will be decremented appropriately. That way we'll have precise idea of how much outbound data is buffered at any given moment.

Afterwards, when sending, we can disable more sends once certain buffer limit was exceeded."

Non trivial work, but it would lead to a correct fix. My patch forces a wait on send and forces the API to be synchronous.

djc:
Okay, thanks. Are you interested in working on that suggestion?

TTimo:
No immediate plans no.

Over the past year our group has submitted a number of pull request to correct a few significant deficiencies in the Windows support.

Most of these patches are still sitting in the queue; I can't justify putting resources towards more nanomsg contributions in the current state of things.

djc:
I'm sorry to hear that, and sorry for the delay in processing your work.

I will try to work my way through your pull requests in the near future.

src/protocols/utils/dist.c
```diff
@@ -71,14 +71,13 @@ int nn_dist_send (struct nn_dist *self, struct nn_msg *msg,
     struct nn_dist_data *data;
     struct nn_msg copy;
 
-    /*  TODO: We can optimise for the case when there's only one outbound
+    /*  TODO: We can optimize for the case when there's only one outbound
```
gdamore: euro vs. us english.  even though i'm in the us, please don't bother to change these.

src/protocols/utils/dist.c
```diff
@@ -71,14 +71,13 @@ int nn_dist_send (struct nn_dist *self, struct nn_msg *msg,
...
         pipe here. No message copying is needed in such case. */
 
     /*  In the specific case when there are no outbound pipes. There's nowhere
-        to send the message to. Deallocate it. */
+        to send the message to. */
     if (nn_slow (self->count) == 0) {
-        nn_msg_term (msg);
-        return 0;
+        return -EAGAIN;
```
gdamore: This fix seems unrelated to the original bug?

TTimo: I will check, but I think it is related. Windows would silently drop messages otherwise, and we need a way to tell when it wasn't sent due to a socket with a full buffer.

gdamore: I object to this change in principle, because you're trying to achieve something that makes no sense with pub/sub.

PUB/SUB is *best effort* delivery.  It doesn't tell you that the receiver(s) received it -- and indeed there could be some that receive it, and some that don't based on buffering state in the individual subscribers.

Basically, if you're expecting to be able to flood without dropping, then you can't use a multicast protocol like this on its own.  You can augment the protocol with a backchannel that is per subscriber to notice missed messages, and replay just those messages that are missed.

But if you need reliable delivery *DON'T USE PUB/SUB*.  The only SP protocol that is "reliable" is REQ/REP.

The "silent drop" behavior of the client here is a design *feature*, not a bug.  Think of this like radio broadcast -- you put the bits out on the wire (or maybe not, if the transmit queue is too full), but you have no idea if anyone was listening or received.

And frankly, if you try to fix *this* case, you'll quickly find that the other cases, where messages are dropped by individual pipes due to flow control for example, becomes an issue as well.  It's fundamentally not a solveable problem with just a single "multicast" channel.

gdamore:
I'm closing this PR as I feel it violates the basic design principles of nanomsg's SP protocol -- which is best effort delivery.  See my comments above for more detail.

TTimo:
Is there something in the nanomsg docs that clearly explains that PUB/SUB
is unreliable? I don't think I ever ran into silent unreliable behavior
when doing PUB/SUB with say, zeromq

gdamore:
I think the docs are deficient.

But yes, zeromq suffers from the same behavior.  If you've not run into, it may be that you've been lucky, or perhaps the data rates you push are lower.  Also, zeromq has additional buffering, and eliminates a level of system calls, so its probably somewhat less likely to drop under load.

None of that changes the fact that architecturally, for both nanomsg and zeromq, PUB/SUB is unreliable.

Indeed, much discussion around the "slow-start" problem for both zeromq and nanomsg is a direct consequence/allusion to this problem.  If the protocols were reliable, or gave notification back to the application, then slow-start would be a non-issue. :-)

TTimo:
There's a pretty good explanation there:
http://zguide.zeromq.org/php:chapter5#Pros-and-Cons-of-Pub-Sub

Alright I learned something today. Thanks for going through through those
old PRs :)

TTimo

gdamore:
Great.  And you're welcome. :-)

PR #339, opened 2014-11-20
WebSocket transport now uses CMSG headers.

wirebirdlabs:
This patch enables the ws:// protocol to have full SP capability. It is submitted under the MIT license.

@sustrik This is intended as the final commit for opcode-aware WebSocket messages, to create the point from which we begin paring down to binary-only messages.


djc:
Closing as requested.

wirebirdlabs:
@djc thank you!

PR #340, opened 2014-11-20
WebSocket Stress test using Autobahn TestSuite

wirebirdlabs:
This stress test is currently designed to test for full RFC 6455 compliance. However, it's likely we will move to binary-only payloads, at which point we will modify this test accordingly. Currently, this test is disabled by default (via `#define NN_WS_STRESS_SKIP_PERF 1` in `tests/ws_stress.c `), since this test requires Autobahn Testsuite to be installed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -71,6 +71,7 @@ add_libnanomsg_test (ipc_stress)
 add_libnanomsg_test (tcp)
 add_libnanomsg_test (tcp_shutdown)
 add_libnanomsg_test (ws)
+add_libnanomsg_test (ws_stress)
 add_libnanomsg_test (tcpmux)
 
 #  Protocol tests.
--- a/Makefile.am
+++ b/Makefile.am
@@ -473,6 +473,7 @@ TRANSPORT_TESTS = \
     tests/tcp \
     tests/tcp_shutdown \
     tests/ws \
+    tests/ws_stress \
     tests/tcpmux
 
 PROTOCOL_TESTS = \
--- a/tests/ws_stress.c
+++ b/tests/ws_stress.c
@@ -1,0 +1,368 @@
+/*
+    Copyright (c) 2012 Martin Sustrik.  All rights reserved.
+    Copyright (c) 2014 Wirebird Labs LLC.  All rights reserved.
+
+    Permission is hereby granted, free of charge, to any person obtaining a copy
+    of this software and associated documentation files (the "Software"),
+    to deal in the Software without restriction, including without limitation
+    the rights to use, copy, modify, merge, publish, distribute, sublicense,
+    and/or sell copies of the Software, and to permit persons to whom
+    the Software is furnished to do so, subject to the following conditions:
+
+    The above copyright notice and this permission notice shall be included
+    in all copies or substantial portions of the Software.
+
+    THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+    IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+    THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+    LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+    FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+    IN THE SOFTWARE.
+*/
+
+#include "../src/nn.h"
+#include "../src/pair.h"
+#include "../src/ws.h"
+
+#include "../src/utils/int.h"
+
+#include "testutil.h"
+
+/*****************************************************************************/
+/*  Stress tests the WebSocket transport using Autobahn Testsuite.           */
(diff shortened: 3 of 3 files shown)


wirebirdlabs:
I don't suggest merging this PR, but can we discuss strategy moving forward in order to close out this old PR?

Autobahn TestSuite was chosen as the industry-standard WebSocket implementation tester when I had envisioned the scope of the WebSocket transport in nanomsg to be wider than it is, but after discussions with @sustrik and @gdamore, I now agree that the WebSocket transport for nanomsg only needs to reasonably implement a subset of RFC 6455 (not nearly as complete a set as originally designed). For the most part, it should not be much more "special" than, say, TCP.

Though, even after scaling back, the WebSocket transport remains "special" for a couple reasons:
1) A few transport-specific features -- such as PING/PONG and chunked transfers -- are effectively handled by the transport layer in libnanomsg, but not exposed to the end-user of libnanomsg. Why do we even support these, if not exposed to the library user? Because it's unclear if browsers/user agents have the prerogative to use these features, even if not explicitly requested by client-side app code running in the browser (nominally talking about just Javascript here). By implementing these features, we can be reasonably sure that the library does not crap-out on what are reasonably modest expectations laid out by RFC-6455.
2) The one feature that is exposed, which makes the WebSocket transport special from other transport types (TCP, inproc, ipc, etc...) is that a new socket-level option that is transport-specific enables the transport to verify that the payload of the message is valid UTF-8 -- it's been contested whether we should keep this, but I would strongly urge we do for now, since that's the encoding of the web.

All said -- this is just a brief history of why the Autobahn TestSuite was originally used as a departure from the established strategy of just tests written in C -- since the WebSocket transport for nanomsg was sufficiently different from other transports. The cost of this was introducing a new big dependency -- Python, in addition to Autobahn TestSuite, in addition to the other things *it* required. Yuck.

Though, one new thing that nanomsg has gained since this was first developed is CI servers Travis and Appveyor, which to my understanding are easier to set up, automate, and maintain based on files in the repo.

All leading to a single question -- **do we keep Autobahn TestSuite?**

If yes -- we would configure Autobahn to turn off all the tests that are irrelevant (e.g., per-message compression). Additionally, probably only configure this test to run within "official" CI repositories, but be off by default for anyone cloning the repo and building locally (since it is non-trivial to set up, and pretty much annihilates the the spirit of minimal dependencies)

If no -- well, I don't have any reasons to suggest we ditch it, other than, "it will take effort" :-)

All said -- I'd like to get feedback on whether we want to keep it, then appropriately close or modify this PR so we can get it off the backlog.

gdamore:
So I need to read the PR but here are my high level thoughts.  

First the socket option isn't just an option.  Since it can impact the wire protocol we must retain the ability to set that. In fact mangos lacks this and I should look at ways to add it. 

I also believe there is no reason why we should settle for less than full rfc6455 compliance in the protocol.  We don't express the details to applications but we can and should do the right thing where the wire protocol is concerned. 

I'm opposed to adding the test suite and Python as a stock build dependency for sure.  

But I think if we can find a way to use this work to test in CI that would be useful. 

Until I go over more carefully it's unclear to me whether this should be done in this repo or perhaps with a supplementary repo containing the extra stuff. 

I will try to take a look at this soon.

gdamore:
So your test code looks mostly good, although it uses the removed APIs.

I'm not sure what impact that has on the test suite.  But it clearly seems to me to depend on autobahn, hence not appropriate for a default build.

I didn't understand where the JSON config files were?

wirebirdlabs:
> I didn't understand where the JSON config files were?

Created dynamically as part of the test (link to code): https://github.com/wirebirdlabs/featherweight-nanomsg/blob/65b8630c11fd620d50e8e0a4f2e6db9523a633fc/tests/ws_stress.c#L60-L76

wirebirdlabs:
> So your test code looks mostly good, although it uses the removed APIs.

Good point. Now, re-glancing at the code myself for a first time in a while, it would need to change a non-trivial amount.

For instance, previously, ws:// sockets could parse TEXT or BINARY message types per-message, yet we have now decided (for the better, I think we all agree) that one of two protocols is set per-socket.

Additionally, I'm not certain where the current implementation stands on PING/PONG delivery to the application level (i.e., does libnanomsg expose these any longer? or did we encapsulate them?). But looking at the test, it clearly expects them (in addition to NN_WS_MSG_TYPE_GONE, which I know no longer exists: https://github.com/wirebirdlabs/featherweight-nanomsg/blob/65b8630c11fd620d50e8e0a4f2e6db9523a633fc/tests/ws_stress.c#L202-L207)

gdamore:
Actually, the decision is per transport, and can be overridden via cmsg.
Sadly, we don’t really have a good way to establish different options on
different transport *instances*; the options get applied to all instances
of the transport.  That’s a limitation of libnanomsg’s design (and btw, yet
another limitation I removed in mangos.  When dealing with e.g. TLS you
really need to have separate “dialers” and “listeners” with different
options.)

Libnanomsg handles PINGS by sending PONGS under the covers.  It never
initiates a PING, and does not expose PONGS to users.  Users *could* cheat
and initiate a PING (or forcibly deliver a PONG) by messing with the cmsg
data, but that’s not documented.

On Sun, Oct 25, 2015 at 5:16 PM, Wirebird Labs LLC <notifications@github.com

wirebirdlabs:
@gdamore I'm poking around with this again, bringing it up to speed with recent changes in WebSocket. Especially, in the context of ensuring shutdown behavior is clean.

I have a hunch this test could be easy to integrate into the automated build on Travis and AppVeyor. I can do the work to make that happen.

wirebirdlabs:
This PR is superseded by https://github.com/nanomsg/nanomsg/pull/563 and will be closed once #563 is acknowledged.

wirebirdlabs:
Superseded by #565. Closing.

Issue #182, opened 2013-11-08
SUB socket doesn't automatically re-connect after server restarts


brunoqc:
I'm using pubsub with tcp and if I stop and restart the server, the client stops receiving messages.

I'm using the latest commit 99131593aed8eab2795ba4bfaaea32b0a64b14c8 on Windows 7 (64-bit).

I built nanomsg using MSVC 2012 express.

Here's Go code to reproduce.

server
```go
package main

import (
	"fmt"
	"time"

	"github.com/op/go-nanomsg"
)

func main() {
	pub, errNew := nanomsg.NewPubSocket()
	if errNew != nil {
		panic(errNew)
	}
	defer pub.Close()

	if _, err := pub.Bind("tcp://0.0.0.0:8861"); err != nil {
		panic(err)
	}

	c := time.Tick(1 * time.Second)
	for now := range c {
		msg := now.String()
		fmt.Println("->", msg)
		pub.Send([]byte("chan "+msg), 0)
	}
}
```

client
```go
package main

import (
	"fmt"

	"github.com/op/go-nanomsg"
)

func main() {
	sub, errNew := nanomsg.NewSubSocket()
	if errNew != nil {
		panic(errNew)
	}
	defer sub.Close()

	if err := sub.Subscribe("chan"); err != nil {
		panic(err)
	}

	if _, err := sub.Connect("tcp://127.0.0.1:8861"); err != nil {
		panic(err)
	}

	for {
		buf, err := sub.Recv(0)
		if err != nil {
			panic(err)
		}

		fmt.Printf("<- %s\n", buf)
	}
}
```

gdamore:
So, I feel that this change goes against the intended design of nanomsg, which is that apps need never notice the arrival and departure of connections.

Ultimately this problem is caused by another issue -- which is that on many platforms you won't notice a lost connection unless you have an outstanding read().  We need to make sure we always have outstanding blocking reads going on; which is tricky given the non-threaded design of the library.

I'm going to have to think about this some more, but this PR's approach is wrong (the API impact to the client), so I'm closing the PR.   The bug will remain open.

gdamore:
Doe we know, is this still a problem with the current head bits?  I have a sneaking suspicion that I've fixed the root problem here.

brunoqc:
I ran the two nanocat commands at the same time. If I restart the server one, the client begin again to get the 'hello' messages.

I'll try the Go code if the Go binding module is not too out-of-date.

[...]

PR #345, opened 2014-12-05
test case and proposed fix for issue #182, sub sockets not reconnecting

TTimo:
test case and proposed fix for issue #182, sub sockets not reconnecting
refreshed and tested against latest master


djc:
Also, it would be nice if you can rewrite the commit message to have a moree concise first line (e.g. max 70 characters or so) and more explanation in subsequent lines.

src/core/sock.c
```diff
@@ -658,6 +658,12 @@ int nn_sock_recv (struct nn_sock *self, struct nn_msg *msg, int flags)
             return rc;
         }
 
+		/* Have to check if one broken connection needs a reconnection at all cases */
+		if(nn_fast (rc == -EAGAIN) && nn_fast(self->statistics.broken_connections > 0)) {
```
gdamore: nn_fast feels wrong here.  For two reasons:

1. We're already in an error path -- this should never be hot -- feels like premature optimization
2. Is it the case that when in this error path, this is really the most  likely case?

There are indent and spacing issues as Dirkjan already noted, as well.

tests/ipc_pub_disconnect.c
```diff
@@ -0,0 +1,97 @@
+/*
+    Copyright (c) 2012 250bpm s.r.o.  All rights reserved.
```
gdamore: This copyright is wrong.

gdamore:
As I indicated in #182 the approach here is wrong -- we need to rethink this and solve it without kicking the error back to the application.  Maybe hard to do without bringing in threads...

PR #348, opened 2014-12-08
Support setting security attributes on Windows

TTimo:
This patch adds support for setting the security attributes when creating the sockets. We use this for IPC named pipes, where we have a priviledged service that needs communicating with processes in other sessions. Comes with a little demo / test.

### DIFF
--- a/src/aio/usock_win.h
+++ b/src/aio/usock_win.h
@@ -75,6 +75,13 @@ struct nn_usock {
     /*  For now we allocate a new buffer for each write to a named pipe. */
     void *pipesendbuf;
 
+    /* Pointer to security attribute*/
+    SECURITY_ATTRIBUTES sec_attr;
+
+    /* Out Buffer and In Buffer size */
+    int outbuffersz;
+    int inbuffersz;
+
     /*  Errno remembered in NN_USOCK_ERROR state  */
     int errnum;
 };
--- a/src/aio/usock_win.inc
+++ b/src/aio/usock_win.inc
@@ -93,6 +93,11 @@ void nn_usock_init (struct nn_usock *self, int src, struct nn_fsm *owner)
     /* NamedPipe-related stuff. */
     memset (&self->pipename, 0, sizeof (self->pipename));
     self->pipesendbuf = NULL;
+    memset (&self->sec_attr, 0, sizeof (SECURITY_ATTRIBUTES));
+
+    /* default size for both in and out buffers is 4096*/
+    self->outbuffersz = 4096;
+    self->inbuffersz = 4096;
 }
 
 void nn_usock_term (struct nn_usock *self)
@@ -501,7 +506,6 @@ static void nn_usock_create_io_completion (struct nn_usock *self)
 static void nn_usock_create_pipe (struct nn_usock *self, const char *name)
 {
     char fullname [256];
-
     /*  First, create a fully qualified name for the named pipe. */
     _snprintf(fullname, sizeof (fullname), "\\\\.\\pipe\\%s", name);
 
@@ -513,10 +517,10 @@ static void nn_usock_create_pipe (struct nn_usock *self, const char *name)
         PIPE_TYPE_BYTE | PIPE_READMODE_BYTE |
             PIPE_WAIT | PIPE_REJECT_REMOTE_CLIENTS,
         PIPE_UNLIMITED_INSTANCES,
-        4096,
-        4096,
+        self->outbuffersz,
+        self->inbuffersz,
         0,
-        NULL);
+        &self->sec_attr);
 
     /* TODO: How to properly handle self->p == INVALID_HANDLE_VALUE? */
     win_assert (self->p != INVALID_HANDLE_VALUE);
@@ -543,7 +547,7 @@ DWORD nn_usock_open_pipe (struct nn_usock *self, const char *name)
         NULL,
         OPEN_ALWAYS,
         FILE_FLAG_OVERLAPPED,
-        NULL);
+        &self->sec_attr);
 
     if (self->p == INVALID_HANDLE_VALUE)
         return GetLastError ();
--- a/src/aio/usock_win.inc
+++ b/src/aio/usock_win.inc
@@ -509,8 +509,6 @@ static void nn_usock_create_pipe (struct nn_usock *self, const char *name)
     /*  First, create a fully qualified name for the named pipe. */
     _snprintf(fullname, sizeof (fullname), "\\\\.\\pipe\\%s", name);
 
-    /* TODO: Expose custom nOutBufferSize, nInBufferSize, nDefaultTimeOut,
-             lpSecurityAttributes */
     self->p = CreateNamedPipeA (
         (char*) fullname,
         PIPE_ACCESS_DUPLEX | FILE_FLAG_OVERLAPPED,
@@ -522,7 +520,6 @@ static void nn_usock_create_pipe (struct nn_usock *self, const char *name)
         0,
         &self->sec_attr);
 
-    /* TODO: How to properly handle self->p == INVALID_HANDLE_VALUE? */
     win_assert (self->p != INVALID_HANDLE_VALUE);
 
     self->isaccepted = 1;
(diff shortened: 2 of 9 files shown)


src/aio/usock_win.inc
```diff
@@ -93,6 +93,11 @@ void nn_usock_init (struct nn_usock *self, int src, struct nn_fsm *owner)
     /* NamedPipe-related stuff. */
     memset (&self->pipename, 0, sizeof (self->pipename));
     self->pipesendbuf = NULL;
+    memset (&self->sec_attr, 0, sizeof (SECURITY_ATTRIBUTES));
+
+    /* default size for both in and out buffers is 4096*/
```
gdamore: spacing nit.

src/aio/usock_win.inc
```diff
@@ -501,24 +506,20 @@ static void nn_usock_create_io_completion (struct nn_usock *self)
...
         PIPE_UNLIMITED_INSTANCES,
-        4096,
-        4096,
+        self->outbuffersz,
+        self->inbuffersz,
         0,
-        NULL);
+        &self->sec_attr);
 
-    /* TODO: How to properly handle self->p == INVALID_HANDLE_VALUE? */
```
gdamore: This was an assertion.  I'm not a windows dev... should this TODO still be dealt with?  Is the assertion truly always valid?  (If so then I agree with removing the comment.)

src/core/sock.c
```diff
@@ -152,6 +152,12 @@ int nn_sock_init (struct nn_sock *self, struct nn_socktype *socktype, int fd)
     /*  Should be pretty much enough space for just the number  */
     sprintf(self->socket_name, "%d", fd);
 
+    /* Security attribute*/
+    self->sec_attr = NULL;
+    self->sec_attr_size = 0;
+    self->inbuffersz = 4096;
+    self->outbuffersz = 4096;
```
gdamore: Should these buffer sizes be expressed as compile time macros somewhere (to make it easier to tune)?

src/core/sock.c
```diff
@@ -152,6 +152,12 @@ int nn_sock_init (struct nn_sock *self, struct nn_socktype *socktype, int fd)
     /*  Should be pretty much enough space for just the number  */
     sprintf(self->socket_name, "%d", fd);
 
+    /* Security attribute*/
```
gdamore: closing comment space nit.

src/core/sock.c
```diff
@@ -292,6 +298,12 @@ static int nn_sock_setopt_inner (struct nn_sock *self, int level,
         return 0;
     }
 
+    if (level == NN_SOL_SOCKET && option == NN_SEC_ATTR) {
+        self->sec_attr = (void*)optval;
```
gdamore: space nit.  void *

src/core/sock.c
```diff
@@ -438,6 +456,16 @@ int nn_sock_getopt_inner (struct nn_sock *self, int level,
             strncpy (optval, self->socket_name, *optvallen);
             *optvallen = strlen(self->socket_name);
             return 0;
+        case NN_SEC_ATTR:
+            memcpy (optval, self->sec_attr, self->sec_attr_size);
```
gdamore: I'm not sure I understand the style rule here.. we have strncpy with a space after it above, but strlen without.  Do functions get a space between the name and open paren, or not?  (I prefer without.  But I prefer consistency even more.)

src/core/sock.h
```diff
@@ -136,6 +136,15 @@ struct nn_sock
 
     /*  The socket name for statistics  */
     char socket_name[64];
+
+    /* Win32 Security Attribute */
+    void*	sec_attr;
```
gdamore: again void *

src/core/sock.h
```diff
@@ -136,6 +136,15 @@ struct nn_sock
...
+    size_t	sec_attr_size;
+
+    /* Out Buffer and In Buffer size */
```
gdamore: This comment adds no practical value -- the variable names are clear enough.

tests/win_sec_attr.c
```diff
@@ -0,0 +1,96 @@
+/*
+    Copyright (c) 2012 Martin Sustrik  All rights reserved.
```
gdamore: Did you copy Martin's code?  I'm surprised to see *only* his copyright here, rather than the actual author's.

gdamore:
Modulo a few stylistic nits, the code looks good to me.  I'm not a Windows guy, so I can't verify the *correctness* for that platform, but it seems reasonable.  I think it would be nice to compress these changes into a single commit (git rebase) before merging, and use that chance to fix the few minor nits I pointed out.

tests/win_sec_attr.c
TTimo: Oh yeah probably a copy paste from another test case. I'll go over your code review comments and clean this up further sometimes in the next few days.

This code is only added to the cmake project files, so it should only be compiled and tested on Windows.

src/core/sock.c
```diff
@@ -292,6 +298,12 @@ static int nn_sock_setopt_inner (struct nn_sock *self, int level,
         return 0;
     }
 
+    if (level == NN_SOL_SOCKET && option == NN_SEC_ATTR) {
```
gdamore: NN_SOL_SOCKET feels like the wrong socket option for this.  Its a transport specific option, so you should define one.  Its kind of unfortunate that we need this at all, as Windows is quite different from UNIX in this regard.

For UNIX we're punting, requiring users to change the permissions of any socket after creation.  Is this possible with Windows?

src/aio/usock_win.h
```diff
@@ -75,6 +75,13 @@ struct nn_usock {
     /*  For now we allocate a new buffer for each write to a named pipe. */
     void *pipesendbuf;
 
+    /* Pointer to security attribute*/
```
gdamore: spacing nit -- space before comment closer please.

src/aio/usock_win.inc
gdamore: Actually, this can fail for a variety of reasons, I think.  Looking at the MSDN site, I see that we should call GetLastError().  We do something like this elsewhere, and indeed a few lines below with CreateFileA we do this.  So we should fix this, or at least leave the comment with TODO in place until we address it.

src/aio/usock_win.inc
```diff
@@ -535,15 +536,14 @@ DWORD nn_usock_open_pipe (struct nn_usock *self, const char *name)
...
         0,
-        NULL,
+        &self->sec_attr,
         OPEN_ALWAYS,
         FILE_FLAG_OVERLAPPED,
-        NULL);
+        0);
 
     if (self->p == INVALID_HANDLE_VALUE)
         return GetLastError ();
```
gdamore: This doesn't clean up the pipe we created above.  Should it?

src/core/sock.c
gdamore: Thinking about this, a transport option that can be used for UNIX IPC calls might be nice too.  Sort of a chmod or something.   But then we should think to make sure we can express this option portably.  I'm not sure how to do that.

gdamore:
There are conflicts that have to be resolved, and feedback items.  I'd also like more thorough review by someone more familiar with the Windows API. 

In the meantime, I'm closing this PR.  Feel free to resubmit with updated code and comments addressed.

src/aio/usock_win.inc
wirebirdlabs: > We do something like this elsewhere

for reference: https://github.com/nanomsg/nanomsg/commit/10a567c89e564def7f5ae01f8c3bb1dab2acfed5#diff-f40426617d7b490658ad071cc19f5707R79

src/aio/usock_win.inc
wirebirdlabs: When `if (self->p == INVALID_HANDLE_VALUE)` is true, there's no pipe to clean up before return; seems fine as is.

src/core/sock.c
wirebirdlabs: That's an ultra-low barrier first step. Wrapped in `#IFNDEF` is a quick way to make it "tunable" externally if you wanted. Anything further, we could keep as a `nice to have` until someone contributes.

src/core/sock.c
wirebirdlabs: For reference, here's the nanomsg style guide for consistency: http://nanomsg.org/development-style.html

Style guide requires space before the open paren.

Issue #334, opened 2014-11-11
Feature Reqest: Windows XP Compability


zerotacg:
As mentioned in #330 and just to keep it separated, current implementation is depending on Windows version 6.00 (Vista) and it seems as if it is only CancelIoEx that prevents compatibility. So this might be within reach.
It will need some research though, the comment in the source mentions some issues and a quick search brings up something as well [2]. They link to how boost handles it, so that might give some leads, but maybe it's not satisfactory for nanomsg's needs.

[1] https://github.com/nanomsg/nanomsg/issues/330#issuecomment-62470467
[2] https://stackoverflow.com/questions/3921111/how-to-find-out-when-cancelio-is-done

gdamore:
I'm closing this.  I can't verify this works, and CancelIO is known to be insufficient to the task.  We might look at changes which might support making WinXP work better in the future, but this is so far a complete anti-goal.

Worse, XP support needs to be determined at run time, because I really, really don't want to try to support a binary built for XP on newer Windows platforms.

Meanwhile the XP platform is aging.  Formally, there have been no more security or technical patches for the past two years (Microsoft ceased all support activity in April 2014.)   Even Vista is in extended support at this point.

Issue #330, opened 2014-11-06
Doesn't build on Windows with CMake + mingw32


stepelu:
Here are the steps performed to reproduce the issue (version 0.4 beta):

1. Lunch CMake-gui, Configure, select 'MinGW Makefiles', specify native compiler for C by pointing to the path of gcc, Generate. OK
2. Go the build directory (I created a sub-dir named build), lunch mingw32-make, error:

```
Z:\sp\Dropbox\Prj\repo\nanomsg\__build\src\nanomsg-0.4-beta\build>mingw32-make
Scanning dependencies of target nanomsg
[  1%] Building C object src/CMakeFiles/nanomsg.dir/core/ep.c.obj
In file included from Z:\sp\Dropbox\Prj\repo\nanomsg\__build\src\nanomsg-0.4-beta\src\core\../aio/../utils/mutex.h:27:0,
                 from Z:\sp\Dropbox\Prj\repo\nanomsg\__build\src\nanomsg-0.4-beta\src\core\../aio/ctx.h:26,
                 from Z:\sp\Dropbox\Prj\repo\nanomsg\__build\src\nanomsg-0.4-beta\src\core\sock.h:29,
                 from Z:\sp\Dropbox\Prj\repo\nanomsg\__build\src\nanomsg-0.4-beta\src\core\ep.c:27:
Z:\sp\Dropbox\Prj\repo\nanomsg\__build\src\nanomsg-0.4-beta\src\core\../aio/../utils/win.h:38:5: error: unknown type name 'ADDRESS_FAMILY'
     ADDRESS_FAMILY sun_family;
     ^
Z:\sp\Dropbox\Prj\repo\nanomsg\__build\src\nanomsg-0.4-beta\src\core\../aio/../utils/win.h:40:17: error: 'ADDRESS_FAMILY' undeclared here (not in a function)
         sizeof (ADDRESS_FAMILY)];
                 ^
src\CMakeFiles\nanomsg.dir\build.make:53: recipe for target 'src/CMakeFiles/nanomsg.dir/core/ep.c.obj' failed
mingw32-make[2]: *** [src/CMakeFiles/nanomsg.dir/core/ep.c.obj] Error 1
CMakeFiles\Makefile2:1334: recipe for target 'src/CMakeFiles/nanomsg.dir/all' failed
mingw32-make[1]: *** [src/CMakeFiles/nanomsg.dir/all] Error 2
Makefile:146: recipe for target 'all' failed
[...]

PR #349, opened 2014-12-08
Experimental Windows XP support

TTimo:
This is the changes we have made for Windows XP support.

This code is working for us in production, but our use case doesn't heavily stress the nanomsg API.

To resolve the symbols problems, we look for the Vista+ only functions at runtime and use them if available. Otherwise the code runs in XP mode.

We also provide a test that highlights some of the problems we are aware of with the XP path (ipc_xp_test.c).

This should be of interest to issues #330 and #334


gdamore:
To be honest, my gut feeling here is that windows XP ought to be delivered as a different platform support.  That said, I don't regularly deliver software for windows, and certainly not for Windows XP.  Microsoft stopped supporting XP a long time ago -- the last service pack was 6 years ago.   Do we really want to expend effort supporting new software developments for an unsupported legacy platform?

djc:
I don't think the answer should necessarily be no, but it should depend on the amount of complexity, and how well it can be squared off in its own little corner without cluttering the modern stuff.

gdamore:
So, I'd like folks working with Windows and with more knowledge of that platform, to contribute/comment here.  Otherwise this is going to sit in a holding pattern -- I'm neither equipped to determine the correctness of the code, nor to test it.  And, to be completely honest, supporting Windows XP feels like asking a lot since I cannot promise that other changes won't break it.  (I'm already unhappy with having to support Windows at all -- I'd really like to have a "port" lead step up ... I'll post about that on the mailing list.)

gdamore:
So this really should be kept off into a corner, or treated as a fork.

Notably I have zero ability to test this for correct function, and I'm not prepared to support Windows XP.

Longer term, the transport and support layers need to be more pluggable, but that's nigh impossible with the current code base; understanding the FSM sufficiently well to make even an attempt is simply far too high a bar for the usual contributors.

Time continues to progress, meanwhile Windows XP becomes less relevant.  I'm closing this PR.  Folks interested in having this function can find it here, and apply the patch themselves manually.  Presumably they are also equipped to test it for proper functionality; something I simply don't have the resources for.

PR #350, opened 2014-12-16
tweaks for build snags in HP-UX (aCC)

tatsujin1:
It's mostly (hopefully) harmless casts and missing (unreachable) returns.

But, in `protocols/pubsub/trie.c`, `nn_node_child()` and bottom of `nn_node_dump()`, there's an alignment problem that I guess could be very bad:
```
"src/protocols/pubsub/trie.c", line 115: warning #4232-D: conversion from
          "struct nn_trie_node *" to a more strictly aligned type
          "struct nn_trie_node **" may cause misaligned access
          nn_node_dump (((struct nn_trie_node**) (self + 1)) [i], indent + 1);
                         ^
"src/protocols/pubsub/trie.c", line 176: warning #4232-D: conversion from
          "struct nn_trie_node *" to a more strictly aligned type
          "struct nn_trie_node **" may cause misaligned access
      return ((struct nn_trie_node**) (self + 1)) + index;
              ^
```
As I'm not sure how to fix this, I inserted a `(void *)` cast which shut the warning up, but the problem still exists, of course.

Also, the test cases `ipc_shutdown` and `tcp_shutdown` fail with "Broken pipe" signal.
If I understand the test code correctly this is kind of its purpose, i.e. sending on a socket that's shutting down?
Is it just a question of capturing the signal so the test doesn't fail, perchance? Any idea how I might go about that?

### DIFF
--- a/src/transports/ws/sws.c
+++ b/src/transports/ws/sws.c
@@ -154,10 +154,10 @@ void nn_sws_init (struct nn_sws *self, int src,
         NN_SWS_UTF8_MAX_CODEPOINT_LEN);
     self->utf8_code_pt_fragment_len = 0;
 
-    self->pings_sent;
+    /*self->pings_sent;
     self->pongs_sent;
     self->pings_received;
-    self->pongs_received;
+    self->pongs_received;*/
 
     nn_fsm_event_init (&self->done);
 }
@@ -317,6 +317,7 @@ static int nn_utf8_code_point (const uint8_t *buffer, size_t len)
 
     /*  Algorithm error; a case above should have been satisfied. */
     nn_assert (0);
+    return NN_SWS_UTF8_INVALID;
 }
 
 static void nn_sws_mask_payload (uint8_t *payload, size_t payload_len,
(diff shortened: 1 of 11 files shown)


src/transports/ws/sws.c
```diff
@@ -154,10 +154,10 @@ void nn_sws_init (struct nn_sws *self, int src,
         NN_SWS_UTF8_MAX_CODEPOINT_LEN);
     self->utf8_code_pt_fragment_len = 0;
 
-    self->pings_sent;
```
gdamore: I'd just nuke this altogether.

src/transports/ws/sws.c
```diff
@@ -317,6 +317,7 @@ static int nn_utf8_code_point (const uint8_t *buffer, size_t len)
 
     /*  Algorithm error; a case above should have been satisfied. */
     nn_assert (0);
+    return NN_SWS_UTF8_INVALID;
```
gdamore: can we choose a better return code?

gdamore:
Modulo one nit, and one unfortunate choice of return code, looks reasonable.  I'd like to see a better return code than repurposing something totally unrelated, to indicate an internal fatal error.  Also, I hope this will be committed as a single commit (git rebase) rather than 13 separate commits merged.

gdamore:
I fixed alot of issues, and now this conflicts.  Lacking HP-UX access myself, its unlikely I'm going to do much with this.

gdamore:
I'm closing this, because it no longer merges clean, and I don't have access to the HPUX test environment.  If someone else wants to reopen, with just relevant fixes, please open a new bug report and PR.

PR #353, opened 2014-12-23
Fixed a SEGV error when using nanomsg on CentOS5/RHEL5

yoink00:
Older versions of glibc have a problem with getaddrinfo_a and the size of stack it allocates for the thread that performs the lookup. This causes a process built with nanomsg to crash when it needs to do a DNS lookup.

This fixes this issue by allowing the use of getaddrinfo_a to be disabled at configuration time by specifying '--disable-getaddrinfo_a'.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -40,7 +40,8 @@ nninclude_HEADERS = \
     src/reqrep.h \
     src/pipeline.h \
     src/survey.h \
-    src/bus.h
+    src/bus.h\
+    src/tcpmux.h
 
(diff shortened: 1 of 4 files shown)


djc:
Can you state that you are submitting the patch under MIT license?

yoink00:
Yes, I am.

djc:
Thanks! I've pushed your changes as 6db4e7d and d907795 after cleaning up the commit messages and fixing a formatting nit.

PR #356, opened 2015-01-02
Fix a leak caused by not freeing the result of getaddrinfo_a().

zboszor:
A small block is leaked for every call of nn_connect() which can be massive on the client side.


marchon:
That was a very good catch.  We need this in the main line.

djc:
Thanks, merged!

PR #359, opened 2015-01-09
Deferred start of the global FSM to avoid race condition with line 1267 ...

wirebirdlabs:
...in FSM handler. The problem was that line 1267, `global->statistics_socket` would not have been initialized properly by the lines 297-305 block, causing the 10sec timer to start in the global FSM. 

@djc can you reproduce, and do you agree with this change? Delaying the FSM start was the obvious/simple fix, but you may have a better idea (feel free to refactor in this branch prior to merging) -- thanks!

### DIFF
--- a/src/core/global.c
+++ b/src/core/global.c
@@ -291,7 +291,6 @@ static void nn_global_init (void)
 
     nn_ctx_init (&self.ctx, nn_global_getpool (), NULL);
     nn_timer_init (&self.stat_timer, NN_GLOBAL_SRC_STAT_TIMER, &self.fsm);
-    nn_fsm_start (&self.fsm);
 
     /*   Initializing special sockets.  */
     addr = getenv ("NN_STATISTICS_SOCKET");
@@ -333,6 +332,8 @@ static void nn_global_init (void)
         errno_assert (rc == 0);
         self.hostname[63] = '\0';
     }
+
+    nn_fsm_start(&self.fsm);
 }
 
 static void nn_global_term (void)


djc:
Err, I think we'd need @sustrik in on this.

wirebirdlabs:
> Err ...

D'oh! My intention was to alert @tailhook instead of @djc since this had to do with statistics. Sorry for the false alarm @djc!

djc:
@tailhook ping?

gdamore:
My gut is that this approach is here is correct.  @tailhook ?

tailhook:
Well, it seems weird that it worked before :)

It seems that delaying fsm start is ok, as long as it's ok to create sockets/endpoints before start. Which seems quite counter-intuitive (even if it's perfectly works now). Probably  the whole sockets machinery should be initialized at start action. What do you think?

wirebirdlabs:
> It seems that delaying fsm start is ok, as long as it's ok to create sockets/endpoints before start. Which seems quite counter-intuitive (even if it's perfectly works now)

_caveat emptor: I'm commenting on a 4mo-old pull-request with 4mo-old knowledge, not having re-wrapped my brain around all the decisions behind the original change._

I see the argument of being counterintuitive, but in another way it remains in the spirit of how libnanomsg reserves and then later activates sockets and connections -- asynchronously.

That said -- I have not fully studied the effect of enqueueing messages to the libnanomsg FSM prior to it starting -- e.g., is it always OK to preload its incoming message queue prior to its handler becoming active? In a formal sense, there should be no prob with this design, and it's appeared (anecdotally) to hold true in the libnanomsg FSM implementation.

All things considered, I'm 100% cool with a better or more elegant solution, but in the meantime would suggest merging this patch since fixes a characterized race that manifested in production without "appearing" to cause other bad things to happen, as long as others can confirm bad things don't "appear" for them either (or ideally, could even contribute some formal tests)!! :smiley:

gdamore:
I'm going to have to stare at this a bit.  I'm a little uncomfortable with approach but the arguments here are sound.  I would love to have more anecdotal evidence that this PR doesn't cause problems before I merge it.

wirebirdlabs:
> I'm going to have to stare at this a bit.  I'm a little uncomfortable with approach but the arguments here are sound.  I would love to have more anecdotal evidence

legit.

wirebirdlabs:
This has been working well in production for a while; merge?

gdamore:
I need to catch up on PRs.  Will do so this week.

wirebirdlabs:
> I need to catch up on PRs.  Will do so this week.

Sounds good.

gdamore:
I redid this as #494 using a separate bug id and better commit comments, and have integrated here: 

commit 991bad6b0199af58e964f54296a2e89094749146
Date:   Tue Jan 6 16:37:21 2015 -0600

    fixes #493 statistics race with FSM

PR #360, opened 2015-01-09
Expanding BASIC_CHECKS to prevent segfaults on invalid socket IDs (e.g., -1)

wirebirdlabs:
When creating a binding in another language, using -1 as a sentinel value for uninitialized sockets would segfault if passed into nanomsg.


wirebirdlabs:
@sustrik What are your thoughts on this change?

djc:
Looks good to me, thanks!

PR #363, opened 2015-01-17
Change from malloc to calloc

marchon:
changes to make sure that the alloced memory block is zero filled,
takes a little more time but ALWAYS worth it.. to keep garbage out of
the memory block.

These changes are submitted under the MIT License

### DIFF
--- a/src/utils/alloc.c
+++ b/src/utils/alloc.c
@@ -56,7 +56,7 @@ void *nn_alloc_ (size_t size, const char *name)
 {
     uint8_t *chunk;
 
-    chunk = calloc (sizeof (struct nn_alloc_hdr) + size, sizeof(size_t));
+    chunk = calloc (sizeof (struct nn_alloc_hdr) + size, sizeof(uint8_t));
     if (!chunk)
         return NULL;
 
@@ -130,7 +130,7 @@ void nn_alloc_term (void)
 
 void *nn_alloc_ (size_t size)
 {
-    return calloc (size, sizeof(size_t));
+    return calloc (size, sizeof(uint8_t));
 }
 
 void *nn_realloc (void *ptr, size_t size)
--- a/src/utils/alloc.c
+++ b/src/utils/alloc.c
@@ -56,7 +56,7 @@ void *nn_alloc_ (size_t size, const char *name)
 {
     uint8_t *chunk;
 
-    chunk = malloc (sizeof (struct nn_alloc_hdr) + size);
+    chunk = calloc (sizeof (struct nn_alloc_hdr) + size, sizeof(size_t));
     if (!chunk)
         return NULL;
 
@@ -101,7 +101,6 @@ void *nn_realloc (void *ptr, size_t size)
 void nn_free (void *ptr)
 {
     struct nn_alloc_hdr *chunk;
-    
     if (!ptr)
         return;
     chunk = ((struct nn_alloc_hdr*) ptr) - 1;
@@ -131,7 +130,7 @@ void nn_alloc_term (void)
 
 void *nn_alloc_ (size_t size)
 {
-    return malloc (size);
+    return calloc (size, sizeof(size_t));
 }
 
 void *nn_realloc (void *ptr, size_t size)
@@ -145,4 +144,3 @@ void nn_free (void *ptr)
 }
 
 #endif
-


reqshark:
@djc @sustrik we need to do this anyway downstream in iOS. it cleans up the last few pieces of msgs i pull off `nn_recv()`

jnicholls:
LGTM

gdamore:
Looking at this code.  I see two problems.

1. realloc() doesn't get the same zero treatment -- so if alloc_ () could have a problem, so could realloc, right?  (Realloc is trickier, because you have to zero only the end of the after the allocation.)

2. I'd have concerns with zeroing full message contents.  This allocation function is used globally in lieu of malloc.  While the zeroing of memory for headers or other accounting structures is probably trivial, if messages are rather large, this could have a rather detrimental effect on performance. 

It would be better, IMO, to try to understand where the 'uninitialized' use occurs, and ensure that callers that ned to zero memory do so explicitly.  (Perhaps only touching the specific structure members that they aren't otherwise overwriting.)  Failing that, perhaps expose an nn_calloc() for use by those callers where the clearing semantics are easier/required, and leave nn_alloc() for use by callers that don't require this semantic (specifically allocating the buffer for message payloads.)

marchon:
You make some good points, but this article makes some interesting points
as well.

https://balexios.wordpress.com/2014/08/02/calloc-vs-malloc-memset/

On Wed, Feb 25, 2015 at 11:24 PM, gdamore <notifications@github.com> wrote:

> Looking at this code. I see two problems.
>
>    1.
>
>    realloc() doesn't get the same zero treatment -- so if alloc_ () could
>    have a problem, so could realloc, right? (Realloc is trickier, because you
>    have to zero only the end of the after the allocation.)
>    2.
>
>    I'd have concerns with zeroing full message contents. This allocation
>    function is used globally in lieu of malloc. While the zeroing of memory
>    for headers or other accounting structures is probably trivial, if messages
>    are rather large, this could have a rather detrimental effect on
>    performance.
>
> It would be better, IMO, to try to understand where the 'uninitialized'
> use occurs, and ensure that callers that ned to zero memory do so
> explicitly. (Perhaps only touching the specific structure members that they
> aren't otherwise overwriting.) Failing that, perhaps expose an nn_calloc()
> for use by those callers where the clearing semantics are easier/required,
> and leave nn_alloc() for use by callers that don't require this semantic
> (specifically allocating the buffer for message payloads.)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/nanomsg/nanomsg/pull/363#issuecomment-76120267>.
>



-- 
P THINK BEFORE PRINTING: is it really necessary?

This e-mail and its attachments are confidential and solely for the
intended addressee(s). Do not share or use them without approval. If
received in error, contact the sender
and delete them.

reqshark:
here's a good quote to that article's point:

> 23:51 \n (Nick Desaulniers): correctness trumps perf.
> 23:53 \n (Nick Desaulniers): since a compiler makes code transforms on your behalf...

gdamore:
The article you list was clearly written without a sufficient understanding of the workings of the system or of real world benchmarking. 

The calloc trick used is indeed faster - if you never overwrite any of the data. In that case the page can be left alone.  But if you do overwrite the data ever then you must pay a penalty as the original shared page must be copied (copy on write). 

The microbenchmark seems like it greatly favors calloc because this work step is skipped. But in any sane use case this would not be true.  The cost will just be paid for in the fault handler rather than in user space. 

If you have a large buffer that you're going to put nonzero data into then you don't want to pay for this clearing either before or after the allocation.  You should just malloc and save all the waste. 

Where the calloc trick wins is when you allow large empty arrays that are mostly sparse. This has to be enough to have whole pages (4 or 8k) left sparse to have any real benefit. This happens but I think it is more the exception than the rule. 

I believe illumos and Solaris just explicitly call the bzero or memset as part of calloc and don't do any MMU trickery. Frankly for most cases this is probably faster since the MMU page table updates aren't free.  It also lets us use smaller chunks of heap than a full page.

gdamore:
Yes.  But the problem is that calloc here is just a lazy way to paste over incorrectness in the code elsewhere.  Better to fix that code so we can have our cake and eat it too.

reqshark:
null terminating a `C string` is way better, the larger the message envelope, the more so. 

but for stable scalability purposes, calloc is a safeguard against human error.

gdamore:
I'm curious to understand, do we understand the errors you are encountering?  Is it in the handling of nanomsg itself, or is it in the payload?  If you're passing around strings in the message payload, these are explicitly *not* null terminated but simple byte arrays -- if you want them treated as null terminated C strings then you need to do that yourself in your application layer protocol.

If however, nanomsg is mishandling things internally -- for example assuming that some nanomsg specific structure is zero filled -- then *that* represents a more serious error, and it needs to be fixed, and fixed properly.  It *may* be the case that explicitly zeroing the structure (perhaps at allocation via calloc()) is preferable from the standpoint of simplicity.  I'm fine with the idea of doing that on small structures used inside libnanomsg, though I still feel that in some cases this is going to be suboptimal compared to explicitly initializing structure members that need it.

The whole performance diatribe above is a red herring.  calloc is not (cannot be) faster than explicitly clearing structure members; and for implementations that use paging (which I think only happens if you allocate full pages at a time -- e.g. message payloads!) hacks, is going to be *slower* due to the cost of copy-on-write.   (And anyway, none of it should happen for tiny structures -- I think even Linux doesn't use this paging hack if you're allocating a small structure.)

So, this PR gets a very explicit -1 from me, along with a request for more information about the nature of the problem being attempted to fix here.  If it is due to nanomsg structures being uninitialized, please file a bug, and lets work on a fix for that.  (With more information I am willing to help fix that problem.)  If it is an application payload problem, well, that's an application bug and out of scope for fixing here.

reqshark:
the bug rears up when using iOS to receive msgs over TCP sourced from other operating systems.

the severity seems to depend partly on which remote nanomsg binding i'm using to send messages.

nature of the bug: every so often msgs have garbage at the end, i.e. bad allocation type garbage.

why would using `calloc` completely fix it? I highly doubt it's an iOS-specific issue.

gdamore:
Are you saying the *payload* is garbled at the end?  That feels like an application error to me.  Can you provide some sample code showing the error?

reqshark:
sure do you have iOS device?

gdamore:
Note that some languages will treat "strings" as null terminated, and include the null termination as part of the payload, and others may not.  Understanding whether your application needs to have the payload null terminated or not is an application problem.  (In fact, in the protocols I normally use, I don't normally null terminate but expect that the application will terminate on receive, using the length of the message.  If you don't want to use the nanomsg supplied message length you can also send a message length in your application protocol itself -- sort of like sending a Pascal style string instead of a C style string.)

gdamore:
I do have an iOS device, but I'm more interested in seeing the source code, rather than running it.

reqshark:
ok just a sec.. i'll paste it shortly

reqshark:
[use this before the calloc commit](https://github.com/reqshark/nanomsg.ios.git)

```bash
# inside your ios project add this
$ git clone https://github.com/reqshark/nanomsg.ios.git && rm nanomsg.ios/README.md
$ cd nanomsg.ios && git checkout 9f5f0d837e
```

the other dependency, would be something like msgevents.h:

```c
#ifndef msg_events
#define msg_events

#include "nn.h"
#include "poller.h"
#include <sys/select.h>

#define NN_IN 1
#define NN_OUT 2

int grc;

int getevents (int s, int events, int timeout){
  fd_set pollset;
  int rcvfd;
  int sndfd;
  int maxfd;

  size_t fdsz;
  struct timeval tv;
  int revents;

  maxfd = 0;
  FD_ZERO (&pollset);

  if (events & NN_IN) {
    fdsz = sizeof (rcvfd);
    grc = nn_getsockopt (s, NN_SOL_SOCKET, NN_RCVFD, (char*) &rcvfd, &fdsz);
    FD_SET (rcvfd, &pollset);
    if (rcvfd + 1 > maxfd)
      maxfd = rcvfd + 1;
  }

  //strictly for sending the next
  if (events & NN_OUT) {
    fdsz = sizeof (sndfd);
    grc = nn_getsockopt (s, NN_SOL_SOCKET, NN_SNDFD, (char*) &sndfd, &fdsz);
    FD_SET (sndfd, &pollset);
    if (sndfd + 1 > maxfd)
      maxfd = sndfd + 1;
  }

  if (timeout >= 0) {
    tv.tv_sec = timeout / 1000;
    tv.tv_usec = (timeout % 1000) * 1000;
  }

  select (maxfd, &pollset, NULL, NULL, timeout < 0 ? NULL : &tv);
  revents = 0;
  if ((events & NN_IN) && FD_ISSET (rcvfd, &pollset))
    revents |= NN_IN;
  if ((events & NN_OUT) && FD_ISSET (sndfd, &pollset))
    revents |= NN_OUT;
  return revents;
}

#endif
```

and then in your iOS app somewhere run this, (obviously it would be better to do it non-blocking and somwhere after `UIApplicationMain`). 

for simplicity and example sake let's just do this:
```c


#include <assert.h>
#import <UIKit/UIKit.h>
#import "AppDelegate.h"
#include "nn.h"
#include "tcp.h"
//#include "pipeline.h"
#include "sleep.h"
#include "pubsub.h"
#include "msg_events.h"

int s, r, eid;

#define ADDR "tcp://202.202.202.202:5555"

int count = 0;

void loop(void);

int main(int argc, char * argv[]) {
  printf("hey\n");

  reconn();
  loop();

  @autoreleasepool {

    s = nn_socket(AF_SP,NN_SUB);

    nn_setsockopt (s, NN_SUB, NN_SUB_SUBSCRIBE, "", 0);

    eid = nn_connect (s, ADDR);

    loop();

    return UIApplicationMain(argc, argv, nil, NSStringFromClass([AppDelegate class]));
  }
}

void loop(void){
  //if descriptor is ready for i/o, getevents returns 1
  //this informs our app when to read from the socket and allocate a new msg buffer
  r = getevents (s, NN_IN, 10);
  if(r == 1){
    char *b = NULL;
    nn_recv (s, &b, NN_MSG, 0);
    printf("buffer: %s %d \n", b, ++count);
    nn_freemsg (b);
    loop();
  } else {
    //nn_sleep(10);
    loop();
  }
}
```

reqshark:
so you see there's not a lot of app code involved here.. it's very spare, basically the nanomsg src directory at beta 5 release

reqshark:
also one more thing, use this to set the remote connection string, in this case it would be where your pub socket lives:

```c
#define ADDR "tcp://202.202.202.202:5555"
```

gdamore:
Ok, so we need to see the sender code, but at a quick guess, this is indeed an application here.  The problem is here:

    nn_recv (s, &b, NN_MSG, 0);
    printf("buffer: %s %d \n", b, ++count);

The buffer you got is *not* guaranteed to be null terminated.  Messages are *byte arrays*, which are not the same thing as null terminated C strings.  You can "insist" that your publisher null terminate the content when it sends (I'd have to see the code for that).  For example, one way to do this when sending from a C program is to use "strlen(str)+1" to ensure that the null byte is included in the message. This "insistence" becomes part of your application layer protocol, and is not part of SP/nanomsg itself.

Demanding that nanomsg apply some null termination logic to the message buffer is actually 100% dead wrong, since you can reasonably have binary protocols where zero bytes are part of the protocol contents.  (Indeed, many many protocols do just this, including, btw, the wire protocol for SP itself!)

gdamore:
And btw, if you're concerned about hardening your protocol against random attackers, you should actually *verify* that any application layer protocol constraints (such as null termination, or if you have encoded lengths instead, that the lengths are in-bounds of the message size), before you rely upon them.  This part of correct protocol handling if you are going to be expose your protocol to the hostile Internet.

gdamore:
Finally, note that the approach used with C language strings, may not work with strings from other languages.  For example. Go does not terminate strings internally, so if you want to ensure that a null is appended you have to include it explicitly, like this "A C string\0"  (note the \0.)

reqshark:
> the approach used with C language strings, may not work with strings from other languages. 

the point of this PR is that we don't have a guarantee about the binding or who does the sending code.

in other words, this makes it compatible with all nanomsg bindings and all send code

reqshark:
at a minimum if you're against this PR would be in favor of doing it as a configurable `socketopt`?

reqshark:
perhaps something that is not the default..

gdamore:
I'm not really in favor of it, no.  The problem you're encountering is *directly* related to the fact that you think you're messages are null-terminated strings, and that is a *wrong* assumption.  Your application layer protocol handling is busted; you need to fix that instead of blaming nanomsg. 

Making nanomsg calloc() payloads to fix your small protocol error is sort of akin to requiring your car tow a trailer full of fuel in case you run out of gas because you were too lazy to keep an eye on your fuel gauge.  It may seem like it helps prevent certain kinds of tragedy (getting stuck in the middle of nowhere without fuel), but the fix actually hurts pretty badly (you're burning a *lot* more CPU for this laziness, and not really fixing the problem) and really addressing the symptom rather than the root cause.  (Because you're not checking that the actual trailer has enough gas in it, it might be empty.)

What I mean is that calloc doesn't actually guarantee a fix here.  For example, if you receive a message that is a full page size (or a multiple thereof), then calloc will not clear the next byte -- the byte after the message is actually still going to be random (in fact, in that case you may very well segfault from accessing beyond the page in that case.)  And on most platforms calloc will not clear memory beyond the explicitly requested message size.  If the message you receive is the same size as what you asked for, that terminating null is going to be overwritten by the message contents which may not be null.  The only actual case where the calloc helps you is if you allocate a region of memory that is *larger* than the received message.  You really need to have a very friendly cooperating peer to do that.  (Or, you *can* use this trick in the application code by pre-allocating the received message with at least one byte larger than the incoming message, and null terminating after receive.  That's a perfectly correct solution, but it requires the app do the right thing.)

The *only* correct approach here is to fix your application.

marchon:
At this point - Without some type of solution - we can PROVE that there is
random corruption of received messages coming out of the nanomsg library.

I is important the we understand there are 2 underlying issues.

1) It is presently a problem that is impacting users - and is not
necessarily going to be caught by our testing methodology, unless we adapt
that strategy to check for buffer overruns - which is exactly what the
result of this can be.

2) Any solution - even one that is technically tactical rather than
strategic is a better condition than we have now.

not fixing this problem can lead to exploits like explained in this
article.

http://www.mathyvanhoef.com/2013/02/understanding-heap-exploiting-heap.html




On Mon, Mar 2, 2015 at 3:29 PM, Bent Cardan <notifications@github.com>
wrote:

> perhaps something that is not the default..
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/nanomsg/nanomsg/pull/363#issuecomment-76810728>.
>



-- 
P THINK BEFORE PRINTING: is it really necessary?

This e-mail and its attachments are confidential and solely for the
intended addressee(s). Do not share or use them without approval. If
received in error, contact the sender
and delete them.

gdamore:
marchon: no.  The messages are not being corrupted as I understand it.  The problem is that nanomsg transfers *exactly* the bytes you request.  Unless you *insist* that that transfer include a terminating zero byte (by including it in your message payload), it will not be included.

gdamore:
Just to be clear, if someone proves that the problem is uninitialized nanomsg control structure members, then that would represent a real libnanomsg bug deserving a fix.  But so far what I've seen is symptomatic of incorrect application expectations, and the proposed "fix" actually doesn't really fix anything at all.  It pastes over the problem some (even most!) of the time, but doesn't prevent the overrun cases -- because if you receive a message that is full size, and then try to blithely do string operations on it, you have no guarantee that the message is null terminated.

This is precisely analogous to using fgets().  While fgets() will not overrun a buffer, if the input takes as many bytes as requested (or more), then the buffer isn't null terminated.  The usual pattern here is to do something like this:

  buf = malloc(n);
  fgets(buf, n-1, stdin);
  buf[n] = 0;

Now replace "fgets()" in the above with "nn_recv" and you have your solution.

marchon:
Actually - Nanomsg is passing a pointed to memory that "IT ALLOCATED"
but It did not clear the entire memory block that it allocated.

When Nanomsg returns a pointed to the memory block that it has allocated -
and passes those functions to other languages that "EXPECT" a null
terminated memory block it is the library's responsibility to make sure
that it does not contribute to potential security holes.

Here is an interesting technical explaination of the issues around allocs /
reallocs and the entire are of problems.



http://www.informit.com/articles/article.aspx?p=1249298

marchon: no. The messages are not being corrupted as I understand it. The
> problem is that nanomsg transfers *exactly* the bytes you request. Unless
> you *insist* that that transfer include a terminating zero byte (by
> including it in your message payload), it will not be included.
>




-- 
P THINK BEFORE PRINTING: is it really necessary?

This e-mail and its attachments are confidential and solely for the
intended addressee(s). Do not share or use them without approval. If
received in error, contact the sender
and delete them.

marchon:
Bent - do we still have the code that we used to debug this problem?

The one where we allocated the buffers - filled them with a pattern of
bytes - and then pulled nano message off of the queue with the overwritten
bit pattern appearing to be part of the bytes returned by the request for a
nanomessage result?

George.

On Mon, Mar 2, 2015 at 3:49 PM, gdamore <notifications@github.com> wrote:

> Just to be clear, if someone proves that the problem is uninitialized
> nanomsg control structure members, then that would represent a real
> libnanomsg bug deserving a fix. But so far what I've seen is symptomatic of
> incorrect application expectations, and the proposed "fix" actually doesn't
> really fix anything at all. It pastes over the problem some (even most!) of
> the time, but doesn't prevent the overrun cases -- because if you receive a
> message that is full size, and then try to blithely do string operations on
> it, you have no guarantee that the message is null terminated.
>
> This is precisely analogous to using fgets(). While fgets() will not
> overrun a buffer, if the input takes as many bytes as requested (or more),
> then the buffer isn't null terminated. The usual pattern here is to do
> something like this:
>
> buf = malloc(n);
> fgets(buf, n-1, stdin);
> buf[n] = 0;
>
> Now replace "fgets()" in the above with "nn_recv" and you have your
> solution.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/nanomsg/nanomsg/pull/363#issuecomment-76818202>.
>



-- 
P THINK BEFORE PRINTING: is it really necessary?

This e-mail and its attachments are confidential and solely for the
intended addressee(s). Do not share or use them without approval. If
received in error, contact the sender
and delete them.

gdamore:
nanomsg doesn't pass the allocated memory block to any functions that expect null termination.  Again, it just things the thing is an array of bytes that might perfectly well contain any number of bytes that are zero.

There are many many application protocols (in fact, arguably this is true of the *vast* majority of them), that do not pass C style strings.  (Much more common when passing variable length string data is to pass a leading size ala Pascal - which is often encoded in *binary*, and that encoding can include null bytes -- for example the value "256" in a typical 16-bit value contains two bytes, 0x01, and 0x00.)

The *application* that takes a nanomsg packet and expects to be able to use C standard library routines that operate on null terminated strings has the responsibility to ensure that the memory contents actually are a well-formed C string. 

Demanding nanomsg do this for you is trying to use a hammer to drill a hole.  You can make a hole with a hammer, but you probably won't like the result, and you're more likely to break things than to have a correct solution.

gdamore:
Actually my fgets() example above is incorrect, because it turns out that fgets() actually does ensure null termination.  A better example would be strncat().  But again, the ideas are similar.

When you deal with bytes coming off a wire, you should not expect that they will contain any particular values.   You (the application author) are responsible for validating the received data.  If you have a higher level protocol, you can do this in convenience functions in library code -- but that would only be appropriate for a protocol that insists on exchanging ASCII strings, and in a library that works in the C runtime environment.  It is not appropriate for nanomsg, which can work with arbitrary application payloads.

gdamore:
@marchon  I do expect that your observed behavior is that if you prealloc a buffer, and fill with bytes, like this: 

buf = malloc(2048);
memset(buf, 'A', 5);
buf[10] = 0; /* ensure we don't stop printing at *some point* */

n = nn_recv(sock, buf, 2048, 0);

and then send it a message consisting solely of the single byte "B", the receiver will get back a value that looks like "BAAAAAAAA".  If you printf() that without null terminating, its going to be ugly.

But in this case the fix is relatively easy:

n = nn_recv(s, buf, 2048-1, 0);
buf[n] = 0;

You can alternatively change your malloc to be one byte larger (2049) than your argument to nn_recv().  If you use NN_MSG asking the library to allocate a message for you, then you're going to have to copy the message somewhere else like this:

n = nn_recv(s, &buf, NN_MSG, 0);
newbuf = malloc(n + 1);
memcpy(newbuf, buf, n);
newbuf[n] = 0;

marchon:
gdamore you are exactly right.

n = nn_recv(s, &buf, NN_MSG, 0);
newbuf = malloc(n + 1);
memcpy(newbuf, buf, n);
newbuf[n] = 0

Would work - but shouldn't have to if we always terminate the buffer with
\0 appended to the nanomsg allocated buffer it solves both problems at
once, and I am happy with that. But expecting the library consumer to be
educated to doing that is inconsistent with what their default assumprtion
of behavior would be. Especially when they are NOT explicitly doing the
allocation from their application code.




On Mon, Mar 2, 2015 at 4:15 PM, gdamore <notifications@github.com> wrote:

> @marchon <https://github.com/marchon> I do expect that your observed
> behavior is that if you prealloc a buffer, and fill with bytes, like this:
>
> buf = malloc(2048);
> memset(buf, 'A', 5);
> buf[10] = 0; /* ensure we don't stop printing at *some point* */
>
> n = nn_recv(sock, buf, 2048, 0);
>
> and then send it a message consisting solely of the single byte "B", the
> receiver will get back a value that looks like "BAAAAAAAA". If you printf()
> that without null terminating, its going to be ugly.
>
> But in this case the fix is relatively easy:
>
> n = nn_recv(s, buf, 2048-1, 0);
> buf[n] = 0;
>
> You can alternatively change your malloc to be one byte larger (2049) than
> your argument to nn_recv(). If you use NN_MSG asking the library to
> allocate a message for you, then you're going to have to copy the message
> somewhere else like this:
>
> n = nn_recv(s, &buf, NN_MSG, 0);
> newbuf = malloc(n + 1);
> memcpy(newbuf, buf, n);
> newbuf[n] = 0;
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/nanomsg/nanomsg/pull/363#issuecomment-76825152>.
>



-- 
P THINK BEFORE PRINTING: is it really necessary?

This e-mail and its attachments are confidential and solely for the
intended addressee(s). Do not share or use them without approval. If
received in error, contact the sender
and delete them.

gdamore:
Having nanomsg zero the end of the buffer only works if nanomsg *overallocates* the buffer.  Also, having it zero the buffer like this is *also* inconsistent with behavior seen when using other system calls.  For example, none of these null terminate:

read()
recv()
recvmsg().

Furthermore, its a gross error (and inconsistent) to apply any meaning of any bytes *beyond* the value returned by nn_recv().  That is, if nn_recv() gives you a return of "4", then by definition the results of any access to the fifth byte is "undefined", particularly if NN_MSG is specified.

Furthermore, if you imagine that you called nn_recv() with a concrete message size into a fixed buffer, it would be a gross error for nanomsg to write to any space *beyond* that size (even just to add a null terminator.)  For example, a perfectly valid protocol usage is to receive a network encoded integer:

struct {
    int x;
   int y;
} mystruct;

count = nn_recv(sock, &mystruct.x, sizeof (mystruct.x), 0);

If nn_recv started writing null termination bytes, it would probably (depending on the machine organization) actually overwrite the value of mystruct.y.  That would clearly be undesirable.

Treat nn_recv() analogously to recv() (which is what Martin claims to have set forth in his design goal of adhering to BSD socket semantics), and treat NN_MSG as a simple convenience wrapper that does malloc(), and you have the right semantic.   Applying further semantics such as null termination is 100% in the responsibility of the application.

gdamore:
Just to be clear, I remain a very strong -1 against this change.

gdamore:
Upon further reflection, I suppose a reasonable way to handle this might be to have a different form of NN_MSG -- call it NN_MSG_STR for example, which allocates an extra byte and null terminates the payload buffer.  I can see for some protocols -- e.g. JSON based ones, this would be a form of convenience for the calling program, making sure that the message buffer has an extra terminating byte at the end of it.

To do this, you create an alternate form of NN_MSG as indicated, and then please do not use calloc (which sucks from a performance standpoint), but merely append a single null byte to the end of the message.  E.g. if nn_recv() would return 4, then do buf[4]=0.

That PR would look quite different from this one but should still be relatively small.

gdamore:
Btw, that NN_MSG_STR hack I suggested is *not* something I'd expect FFI callers to use.  Its really a convenience for C programmers working directly against libnanomsg, IMO.  I believe most non-C runtimes (e.g. Java, Go, Python, etc.) actually use length encoding rather than null termination. I'm pretty sure C++ follows the C conventions here, but I can't think of any language other than C or C++ that uses or typically needs null termination for correctness.

marchon:
gdamore - in response to your post about this change "Just to be clear, I
remain a very strong +1 in favor of this change and have gone out of my way
to provide detail as to the necessity."

I appreciate your expertise in this area, but I do believe that you are
wrong, and that your position while well intentioned don't represent good
library design, and this a a library after all.

To be more specific I believe that you are in error about the non-C
runtimes.  "I believe most non-C runtimes (e.g. Java, Go, Python, etc.)
actually use length encoding rather than null termination."

According to the Python Software Foundation (the subject matter experts)
have an explanation of interacting with C / C++ and it is quite clear from
the long explanation that strings and stream object are expected to be null
filled and null terminated.
https://docs.python.org/3/library/ctypes.html#module-ctypes

If you read the link attached (summarized below); (which is talking about
interfacing with c via ctypes) it appears clear that the strings are
expected to be null terminated, and in the example the python
create_string_buffer contains 3 NULL bytes in a row.

While I expect that you need more than my word, do please review the other
links for additional validation.  If necessary, I will continue down the
road of providing more examples, but it should be at this point unnecessary
as my positions are fact based. This is not a question of what you think
the best practice should be - this in fact a library design issue.

With that in mind - our proposed change improves the situation for non-C
programmers without changing any external user expectations. Anyone
depending on this bad behavior is barking up the wrong tree. So the
negative impact at best will be in the price of clearing the memory. It
fixes more problems than it creates.





c_wchar_p <https://docs.python.org/3/library/ctypes.html#ctypes.c_wchar_p>
wchar_t * (NUL terminated)string or Nonec_char_p
<https://docs.python.org/3/library/ctypes.html#ctypes.c_char_p>char * (NUL
terminated)bytes object or None

>>> from ctypes import *>>> p = create_string_buffer(3)            # create a 3 byte buffer, initialized to NUL bytes>>> print(sizeof(p), repr(p.raw))3 b'\x00\x00\x00'>>> p = create_string_buffer(b"Hello")     # create a buffer containing a NUL terminated string>>> print(sizeof(p), repr(p.raw))6 b'Hello\x00'>>> print(repr(p.value))b'Hello'>>> p = create_string_buffer(b"Hello", 10) # create a 10 byte buffer>>> print(sizeof(p), repr(p.raw))10 b'Hello\x00\x00\x00\x00\x00'>>> p.value = b"Hi">>> print(sizeof(p), repr(p.raw))10 b'Hi\x00lo\x00\x00\x00\x00\x00'>>>


On Mon, Mar 2, 2015 at 7:44 PM, gdamore <notifications@github.com> wrote:

> Btw, that NN_MSG_STR hack I suggested is *not* something I'd expect FFI
> callers to use. Its really a convenience for C programmers working directly
> against libnanomsg, IMO. I believe most non-C runtimes (e.g. Java, Go,
> Python, etc.) actually use length encoding rather than null termination.
> I'm pretty sure C++ follows the C conventions here, but I can't think of
> any language other than C or C++ that uses or typically needs null
> termination for correctness.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/nanomsg/nanomsg/pull/363#issuecomment-76863794>.
>



-- 
P THINK BEFORE PRINTING: is it really necessary?

This e-mail and its attachments are confidential and solely for the
intended addressee(s). Do not share or use them without approval. If
received in error, contact the sender
and delete them.

marchon:
https://www.securecoding.cert.org/confluence/display/seccode/EXP33-C.+Do+not+read+uninitialized+memory


EXP33-C. Do not read uninitialized memory
<https://www.securecoding.cert.org/confluence/display/seccode/EXP33-C.+Do+not+read+uninitialized+memory>

Uninitialized automatic variables or dynamically allocated memory has
indeterminate
values
<https://www.securecoding.cert.org/confluence/display/seccode/BB.+Definitions#BB.Definitions-indeterminatevalue>,
which for objects of some types can be a trap representation
<https://www.securecoding.cert.org/confluence/display/seccode/BB.+Definitions#BB.Definitions-traprepresentation>.
Reading such trap representations is undefined behavior
<https://www.securecoding.cert.org/confluence/display/seccode/BB.+Definitions#BB.Definitions-undefinedbehavior>
 (seeundefined behavior 10
<https://www.securecoding.cert.org/confluence/display/seccode/CC.+Undefined+Behavior#CC.UndefinedBehavior-ub_10>
 and undefined behavior 12
<https://www.securecoding.cert.org/confluence/display/seccode/CC.+Undefined+Behavior#CCUndefinedBehavior-ub12>);
it can cause a program to behave in an unexpected
<https://www.securecoding.cert.org/confluence/display/seccode/BB.+Definitions#BB.Definitions-unexpectedbehavior>
manner
and provide an avenue for attack. In many cases, compilers issue a warning
diagnostic message when reading uninitialized variables (see MSC00-C.
Compile cleanly at high warning levels
<https://www.securecoding.cert.org/confluence/display/seccode/MSC00-C.+Compile+cleanly+at+high+warning+levels>
for
more information).

....
Risk Assessment

Reading uninitialized variables is undefined behavior
<https://www.securecoding.cert.org/confluence/display/seccode/BB.+Definitions#BB.Definitions-undefinedbehavior>
and
can result in unexpected program behavior
<https://www.securecoding.cert.org/confluence/display/seccode/BB.+Definitions#BB.Definitions-unexpectedbehavior>.
In
some cases, these security flaws
<https://www.securecoding.cert.org/confluence/display/seccode/BB.+Definitions#BB.Definitions-securityflaw>
may
allow the execution of arbitrary code.

Reading uninitialized variables for creating entropy is problematic,
because these memory accesses can be removed by compiler optimization.
VU#925211 <http://www.kb.cert.org/vuls/id/925211> is an example of a
vulnerability
<https://www.securecoding.cert.org/confluence/display/seccode/BB.+Definitions#BB.Definitions-vulnerability>
caused
by this coding error.

Rule

Severity

Likelihood

Remediation Cost

Priority

Level

EXP33-C

High

Probable

Medium

*P12*

*L1*




On Mon, Mar 2, 2015 at 8:55 PM, George Lambert <marchon@gmail.com> wrote:

> gdamore - in response to your post about this change "Just to be clear, I
> remain a very strong +1 in favor of this change and have gone out of my
> way to provide detail as to the necessity."
>
> I appreciate your expertise in this area, but I do believe that you are
> wrong, and that your position while well intentioned don't represent good
> library design, and this a a library after all.
>
> To be more specific I believe that you are in error about the non-C
> runtimes.  "I believe most non-C runtimes (e.g. Java, Go, Python, etc.)
> actually use length encoding rather than null termination."
>
> According to the Python Software Foundation (the subject matter experts)
> have an explanation of interacting with C / C++ and it is quite clear from
> the long explanation that strings and stream object are expected to be null
> filled and null terminated.
> https://docs.python.org/3/library/ctypes.html#module-ctypes
>
> If you read the link attached (summarized below); (which is talking about
> interfacing with c via ctypes) it appears clear that the strings are
> expected to be null terminated, and in the example the python
> create_string_buffer contains 3 NULL bytes in a row.
>
> While I expect that you need more than my word, do please review the other
> links for additional validation.  If necessary, I will continue down the
> road of providing more examples, but it should be at this point unnecessary
> as my positions are fact based. This is not a question of what you think
> the best practice should be - this in fact a library design issue.
>
> With that in mind - our proposed change improves the situation for non-C
> programmers without changing any external user expectations. Anyone
> depending on this bad behavior is barking up the wrong tree. So the
> negative impact at best will be in the price of clearing the memory. It
> fixes more problems than it creates.
>
>
>
>
>
> c_wchar_p <https://docs.python.org/3/library/ctypes.html#ctypes.c_wchar_p>
> wchar_t * (NUL terminated)string or Nonec_char_p
> <https://docs.python.org/3/library/ctypes.html#ctypes.c_char_p>char * (NUL
> terminated)bytes object or None
>
> >>> from ctypes import *>>> p = create_string_buffer(3)            # create a 3 byte buffer, initialized to NUL bytes>>> print(sizeof(p), repr(p.raw))3 b'\x00\x00\x00'>>> p = create_string_buffer(b"Hello")     # create a buffer containing a NUL terminated string>>> print(sizeof(p), repr(p.raw))6 b'Hello\x00'>>> print(repr(p.value))b'Hello'>>> p = create_string_buffer(b"Hello", 10) # create a 10 byte buffer>>> print(sizeof(p), repr(p.raw))10 b'Hello\x00\x00\x00\x00\x00'>>> p.value = b"Hi">>> print(sizeof(p), repr(p.raw))10 b'Hi\x00lo\x00\x00\x00\x00\x00'>>>
>
>
> On Mon, Mar 2, 2015 at 7:44 PM, gdamore <notifications@github.com> wrote:
>
>> Btw, that NN_MSG_STR hack I suggested is *not* something I'd expect FFI
>> callers to use. Its really a convenience for C programmers working directly
>> against libnanomsg, IMO. I believe most non-C runtimes (e.g. Java, Go,
>> Python, etc.) actually use length encoding rather than null termination.
>> I'm pretty sure C++ follows the C conventions here, but I can't think of
>> any language other than C or C++ that uses or typically needs null
>> termination for correctness.
>>
>> —
>> Reply to this email directly or view it on GitHub
>> <https://github.com/nanomsg/nanomsg/pull/363#issuecomment-76863794>.
>>
>
>
>
> --
> P THINK BEFORE PRINTING: is it really necessary?
>
> This e-mail and its attachments are confidential and solely for the
> intended addressee(s). Do not share or use them without approval. If
> received in error, contact the sender
> and delete them.
>



-- 
P THINK BEFORE PRINTING: is it really necessary?

This e-mail and its attachments are confidential and solely for the
intended addressee(s). Do not share or use them without approval. If
received in error, contact the sender
and delete them.

reqshark:
so cert says dont do it

reqshark:
@marchon when are you going to learn how to use github instead of email

marchon:
What - this conversation is on GitHub? I thought it was the mailing list -
I should learn how to do that.

G.


On Mon, Mar 2, 2015 at 9:33 PM, Bent Cardan <notifications@github.com>
wrote:

> @marchon <https://github.com/marchon> when are you going to learn how to
> use github instead of email
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/nanomsg/nanomsg/pull/363#issuecomment-76875743>.
>



-- 
P THINK BEFORE PRINTING: is it really necessary?

This e-mail and its attachments are confidential and solely for the
intended addressee(s). Do not share or use them without approval. If
received in error, contact the sender
and delete them.

gdamore:
Reading python references, it appears that it distinguishes the differences between null terminated C strings, and arbitrary byte arrays (pointers to arrays of bytes -- it uses C_POINTER(c_char) for the latter, and has a "first" class item for null terminated strings.)  That's not terribly surprising since C programmers often expect special handling of the null byte.  (Recall I'm primarily a C programmer by trade -- I work in the kernel mostly, although I've done tons of work the C library on illumos systems, as well as for various BSD systems.)

Your argument, @marchon, doesn't really hold water.  The values passed are *not* generally C strings.  The fact that your application wants to treat packet data as null terminated string data is actually a programming error in your application, if it doesn't validate that it is safe to do so before hand.  (That would also be true if using various other C routines, like write(), memcpy(), etc.  Only the str* and printf* family of routines really expect special handling of null bytes.  All other C library routines that operate on byte regions are "8-bit clean".)

As I said, I'd be willing to support a special "helper" for your application that is a custom NN_MSG variant (I proposed NN_MSG_STR) as a convenience.  I'd be vehemently opposed to making the behavior the default because it is more often *wrong* than it is right.  The special exception here for protocols that exchange strictly ASCII or UTF-8 data, such as JSON strings, as well as certain kinds of toy protocols.  Most serious protocols these days use binary encodings and those that don't often have to do crazy contortions -- like MIME encoding email to pass over legacy SMTP -- but even modern SMTP supports 8-bit clean transfer these days.)

In order to implement this *correctly* you *MUST* allocate an extra byte.  That may have surprising effects to the end user.  For example, if a program sends and receives messages of precisely 4k, the allocations will now be 4097 bytes, which is a really really poor fit for a variety of heap configurations.  (In many cases this will result in 2 pages being allocated instead of 1, doubling the memory cost!)  If you're exchanging pure binary data, you neither need nor want that extra useless byte on the end.

The advice given by CERT is correct, and relevant, and supports my statements here.  If you use a function that expects a null terminated string, and you *the application developer* have not ensured that the values being passed are actually null terminated strings, then you can well be violating the secure programming rules.  You can quite reasonably trigger segmentation violation (SIGSEGV) errors, Or you could wind up accessing memory used for other things. 

TO BE CLEAR.  nanomsg message buffers *are not null terminated C strings*.  They are *byte arrays*.  To treat them as null terminated strings is *wrong*, unless you have first ensured that they are indeed suitable for such use (such as by terminating them yourself, or checking for a null termination.  One way to check is with a utility like strnlen() - note the use of the "n" there.)  The issue here is making sure that callers understand the semantic here.

The sockets convention in use here is that recv() and related functions receive byte arrays, not strings.  (In fact, on the wire, even most protocols -- such as JSON-RPC over HTTP -- do not send zero bytes.  Instead they send a length and rely on the application to terminate appropriately.)  Zero terminating packets would be quite surprising to anyone used to working with more common networking stacks -- that includes at least the BSD sockets library, System V STREAMs, WinSock, OpenSSL, ØMQ, RabbitMQ.  In fact, I cannot think of any typical networking library that *does* do null termination like this. A brief survey I did of HTTP client libraries didn't turn up any that did this kind of C string termination either, although my search was hardly exhaustive.

gdamore:
Btw, nn_recv makes this especially clear, because the buffer is actually not a char pointer, but a void pointer.  That is, it is saying this is a pointer to memory to store data, without making any claims about the data type of the contents.

Many C APIs sadly conflate char pointers and void pointers.  (I think this comes about from older compilers without support for void pointers.)  But, nanomsg gets it "right" in this regard.  It also does the same thing for nn_send().  The point is a void pointer, not a character pointer.

gdamore:
Btw, its really poor protocol design to depend on null termination.  Doing so means your application is not going to be 8-bit clean, and therefore cannot be used to exchange raw binary data.  So, while this lazy approach can seem useful, in my experience this kind of approach later on comes back to bite you in the arse when you want to expand your protocol to exchange other things.

jnicholls:
I think this thread has gone on long enough.  The bottom line is – coming from an expert's point of view – @gdamore is completely spot on with regards to everything said here, and you should close this issue and move on to more important things :)

gdamore:
I can't close the issue.  I'm neither the maintainer (nor a committer), nor the submitter.

djc:
@gdamore, thanks for digging into this. I agree with your analysis that this is not a bug.

nickdesaulniers:
Some confusing here seems to be around the idea that the way "strings" are represented in C are the canonical form of strings in all other languages, when whether that is true or not is wholly orthogonal and irrelevant to nanomsg's protocol, which does not call for C style strings.

If you want a C style string from a message received from nanomsg, you __must__ copy the message into a new buffer with length + 1 and explicitly null terminate it.  Nanomsg is more of a binary protocol than a text based protocol, so sending additional bytes for null termination is unnecessary.  If the raw bytes you sent should be treated by the receiver as a C style string, it's on the receiver or sender to handle null termination at the application level, not libnanomsg binding level.

Sorry for not letting sleeping dogs lie. :cold_sweat: :fearful:

PR #364, opened 2015-01-23
replace asserts with return -1,

glycerine:
so nn_tcpmuxd() can be sanely called in a library. For example, in my R bindings, I would like to use nn_tcpmuxd(), and I can't have a failed call crashing the entire R application [https://github.com/glycerine/nanomsgardvark/commit/2b6af3a74b83b3f4be3a017b86027359e59a047d].

### DIFF
--- a/src/devices/tcpmuxd.c
+++ b/src/devices/tcpmuxd.c
@@ -91,19 +91,20 @@ int nn_tcpmuxd (int port)
 
     /*  Start listening on the specified TCP port. */
     tcp_listener = socket (AF_INET, SOCK_STREAM, IPPROTO_TCP);
-    errno_assert (tcp_listener >= 0);
+    if (tcp_listener < 0) { return -1; }
     opt = 1;
     rc = setsockopt (tcp_listener, SOL_SOCKET, SO_REUSEADDR, &opt,
         sizeof (opt));
-    errno_assert (rc == 0);
+    if (rc != 0) { return -1; }
     memset (&tcp_addr, 0, sizeof (tcp_addr));
     tcp_addr.sin_family = AF_INET;
     tcp_addr.sin_port = htons (port);
     tcp_addr.sin_addr.s_addr = INADDR_ANY;
     rc = bind (tcp_listener, (struct sockaddr*) &tcp_addr, sizeof (tcp_addr));
-    errno_assert (rc == 0);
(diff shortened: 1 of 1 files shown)


marchon:
If we are going to do this here, we need to do that everywhere.  But it does make sense to me at first blush..  We can take a look at the other files to see where else it needs to be done.

djc:
@sustrik uses asserts very deliberately to flush out bugs. I don't think we should wholesale remove them unless we can show for each of them that it's not actually a bug, but reason out why it happens, and then we should enter a proper error path.

Can we do so for each of the asserts in `nn_tcpmux()`?

glycerine:
In cases where the error represents *external error conditions* (from the os: disk out of space, no file handles left) or user error (requested port already occupied, malformed/unparsable ip address given) then the library should really not crash. This is because if it crashes, then it is impossible to use it as a library in an application that cannot crash on the user. In other cases, asserts are perfectly fine when they are insisting upon correct internal program logic. This allows the library logic to be corrected and improved or generalized when needed.

All the errno_asserts here that I've changed to returning -1 represent external conditions or user-input errors, not internal logic errors. Therefore returning -1 is the appropriate action here.

I agree that I wish there was a more general approach to handling errors available, especially when there is more information needed than can be conveyed by errno. Howevere this particular function nn_tcpmuxd() is an initializer of a service. In this case, since it is a standalone function, I think we can safefly proceed without waiting for an entire review of the library (which is a separate issue).

djc:
That sounds fair. Should we make sure an errno is set before we return from `nn_tcpmux()`, then? It doesn't look like your patch does that.

glycerine:
errno_assert usage means that the error information is in errno. Each of these system calls sets errno: socket, setsockopt, bind, listen.

I guess we could put in an assert(errno != 0) just before each return. Seems somewhat unusual style, but I'm not opposed to it if it clarifies the concern for the reader.

glycerine:
I do see that errno is not cleared before the first call to socket, so that should probably be done to avoid false positive from leftover other errors. I'll add to the pull.

djc:
Can you state that you are submitting these patches under the MIT license?

glycerine:
Yes, naturally, this code in licensed under the MIT license.

djc:
Squashed and edited the commit message little bit, committed as 12bbd3c. Thanks!

PR #370, opened 2015-02-11
Enhancements to the request-reply SP RFC

rzezeski:
* Fixed some typos.
* Reworked one of the paragraphs to be a little tighter.
* Generated a fresh .txt file (it seems there have been several changes since this was done, and also someone may have modified this file manually because I noticed that xml2rfc wants 2 spaces after a period, a holdover from typewriter days).

Feel free to use this patch as-is or make your own changes based on it. I just thought I'd share.


djc:
Thanks!

PR #384, opened 2015-03-10
Run autogen.sh to generate configure script - README does not mention th...

anshumang:
Added a step in README for non-Windows build...


djc:
Thanks!

reqshark:
nice!

PR #388, opened 2015-03-11
surveyor protocol broken - no backtrace (fixes #387)

### LINKED ISSUE #387: surveyor protocol broken - no backtrace
Due to the lack of a backtrace, it is impossible to connect multiple surveyors to a single respondent, because we would not know where to route responses.   This makes it impossible to set up devices using these sockets, and it also prevents perfectly reasonable topologies where a single responder may want to support multiple surveyors.   (Example use case for that: a printer that supports multiple print servers -- which may be acting in an HA capacity.)  In fact, I think it would be reasonable and possible to build up various voting / consensus algorithms using multiple respondents and multiple surveyors, if the protocol is fixed.

I've proposed using a backtrace just like REQ/REP, but since we're breaking the wire format, I'm bumping the protocol numbers up slightly.  I'm not going to bother doing this for the websocket transport since I think that transport is so brand-spanking new that I don't think there are conflicting cases in deployment yet.

gdamore:
This adds backtraces as has been extensively discussed on the mailing lists, to the Survey protocol.  The logic is heavily borrowed from the REQ/REP processing.  Initial test results look good, and I have successfully connected multiple surveyors to a single responder socket and seen it work.  (This can simply be tested with nanocat, by running each of the following commands in a separate window):

% ./nanocat --respondent --bind=tcp://127.0.0.1:4001 -A -D "ack"
% ./nanocat --surveyor --connect=tcp://127.0.0.1:4001 -A -D one -i 1
% ./nanocat --surveyor --connect=tcp://127.0.0.1:4001 -A -D two -i 1

### DIFF
--- a/rfc/sp-surveyor-01.txt
+++ b/rfc/sp-surveyor-01.txt
@@ -1,0 +1,728 @@
+
+
+
+
+Internet Engineering Task Force                          G. D'Amore, Ed.
+Internet-Draft
+Intended status: Informational                         February 19, 2015
+Expires: August 23, 2015
+
+
+                Surveyor/Respondent Scalability Protocol
+                          sp-request-reply-01
+
+Abstract
+
+   This document defines a scalability protocol used for performing
+   surveys and collecting responses amongst a number of stateless
+   processing nodes, and returning the results of those surveyors.  This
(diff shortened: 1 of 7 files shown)


gdamore:
Integrated.

PR #405, opened 2015-04-27
Support for nn_geterror

vseryakov:
This new function returns an error for an endpoint and works for now with TCP bind/connect endpoints returning errors when any of such endpoint cannot bind/connect for any reason.


djc:
This is a pretty crazy pull request, seemingly containing a few unrelated changes and a zillion merges. Can you squash it down to something more reasonable and single-topic, please?

reqshark:
+1 please rebase/squash

vseryakov:
oops, sorry, i reused the whole repo which i used to keep track of the upstream project, will do

PR #406, opened 2015-04-28
Added nn_geterror function that returns the last errno for an endpoint

vseryakov:
This will allow to check status of an endpoint especially for TCP connect/bind situations

### DIFF
--- a/doc/nn_geterror.txt
+++ b/doc/nn_geterror.txt
@@ -1,0 +1,51 @@
+nn_geterror(3)
+===========
+
+NAME
+----
+nn_geterror - retrieve the current error for an endpoint
+
+
+SYNOPSIS
+--------
+*#include <nanomsg/nn.h>*
+
+*int nn_geterror (int s, int ep);*
+
+
+DESCRIPTION
+-----------
+It allows at any time request for an endpoint the last error set due to a condition 
+that prevented the endpoint operating normally.
+
+RETURN VALUE
+------------
+Returns value of 'last_errno' for the given socket and endpoint.
+
+
+ERRORS
+------
+No errors are defined.
+
+
+EXAMPLE
+-------
+
+----
+int ep = nn_connect(s, "tcp://127.0.0.1:5555"))
+rc = nn_send (s, "ABC", 3, 0);
+if (rc < 0)
+    printf ("nn_send failed with error code %d\n", nn_geterror (s, ep));
+----
+
+
+SEE ALSO
+--------
+linknanomsg:nn_strerror[3]
+linknanomsg:nanomsg[7]
+
+AUTHORS
+-------
+Vlad Seryakov <vseryakov@gmail.com>
+Martin Sustrik <sustrik@250bpm.com>
+
--- a/src/core/sock.c
+++ b/src/core/sock.c
@@ -535,6 +535,29 @@ int nn_sock_rm_ep (struct nn_sock *self, int eid)
     return 0;
 }
 
+int nn_sock_geterror_ep (struct nn_sock *self, int eid)
+{
+    struct nn_list_item *it;
+    struct nn_ep *ep;
+    int err = 0;
+
+    nn_ctx_enter (&self->ctx);
+
+    /*  Find the specified enpoint. */
+    ep = NULL;
+    for (it = nn_list_begin (&self->eps);
+          it != nn_list_end (&self->eps);
+          it = nn_list_next (&self->eps, it)) {
+        ep = nn_cont (it, struct nn_ep, item);
+        if (ep->eid == eid) {
+            err = ep->last_errno;
+            break;
+        }
+    }
+    nn_ctx_leave (&self->ctx);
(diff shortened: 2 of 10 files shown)


src/core/sock.c
```diff
@@ -535,6 +535,29 @@ int nn_sock_rm_ep (struct nn_sock *self, int eid)
...
 
+int nn_sock_geterror_ep (struct nn_sock *self, int eid)
+{
+    struct nn_list_item *it;
+    struct nn_ep *ep;
+    int err = 0;
+
+    nn_ctx_enter (&self->ctx);
+
+    /*  Find the specified enpoint. */
```
gdamore: "endpoint". (spelling)

src/core/sock.h
```diff
@@ -165,6 +165,9 @@ int nn_sock_add_ep (struct nn_sock *self, struct nn_transport *transport,
 /*  Remove the endpoint with the specified ID from the socket. */
 int nn_sock_rm_ep (struct nn_sock *self, int eid);
 
+/* Return last_errno for the endpoint with the specified ID */
```
gdamore: nit. Add a period at the end of the sentence.  (To match other lines in this file.)

gdamore:
This looks like a useful change.  I want to spend a little more time looking at it, if you don't mind.

vseryakov:
No problem, i have several more enhancements in my queue, i am building a backend system in node.js and use nanomsg for messaging, cache....

doc/nn_geterror.txt
```diff
@@ -0,0 +1,51 @@
...
+NAME
+----
+nn_geterror - retrieve the current error for an endpoint
+
+
+SYNOPSIS
+--------
+*#include <nanomsg/nn.h>*
+
+*int nn_geterror (int s, int ep);*
```
gdamore: The other examples single quote (with ') the named arguments.  While I'm not a big fan of whatever markup this style is, we should be consistent for now.

doc/nn_geterror.txt
```diff
@@ -0,0 +1,51 @@
...
+
+SEE ALSO
+--------
+linknanomsg:nn_strerror[3]
+linknanomsg:nanomsg[7]
+
+AUTHORS
+-------
+Vlad Seryakov <vseryakov@gmail.com>
+Martin Sustrik <sustrik@250bpm.com>
```
gdamore: I think you should remove Martin's name -- you're the sole author of this document.

doc/nn_geterror.txt
```diff
@@ -0,0 +1,51 @@
...
+--------
+*#include <nanomsg/nn.h>*
+
+*int nn_geterror (int s, int ep);*
+
+
+DESCRIPTION
+-----------
+It allows at any time request for an endpoint the last error set due to a condition 
+that prevented the endpoint operating normally.
```
gdamore: English wordsmithing (I guess English isn't your first language.)  Try this:

Retrieves the last error condition for a given endpoint on the socket.  The returned value is the same as for 'nn_errno(3)' except that the error scope is limited to a specific endpoint.

src/core/sock.c
```diff
@@ -535,6 +535,29 @@ int nn_sock_rm_ep (struct nn_sock *self, int eid)
...
+    ep = NULL;
+    for (it = nn_list_begin (&self->eps);
+          it != nn_list_end (&self->eps);
+          it = nn_list_next (&self->eps, it)) {
+        ep = nn_cont (it, struct nn_ep, item);
+        if (ep->eid == eid) {
+            err = ep->last_errno;
+            break;
+        }
```
gdamore: QUESTION: What is the expected behavior if the endpoint does not exist.  I'd argue that we ought to return EINVAL.

src/nn.h
```diff
@@ -356,6 +356,7 @@ NN_EXPORT int nn_send (int s, const void *buf, size_t len, int flags);
 NN_EXPORT int nn_recv (int s, void *buf, size_t len, int flags);
 NN_EXPORT int nn_sendmsg (int s, const struct nn_msghdr *msghdr, int flags);
 NN_EXPORT int nn_recvmsg (int s, struct nn_msghdr *msghdr, int flags);
+NN_EXPORT int nn_geterror (int s, int how);
```
gdamore: seems like "how" is not really the right name here -- maybe "endpoint" ?

gdamore:
Ok, I'm pretty happy with this.  If you can do the following I'll be even happier:

a) Fix up the nits & wordsmithing stuff I identified.

b) Lets figure out what the answer to a missing end point should be. I notice that the other cases where we can use an ep ID (nn_shutdown) returns -EINVAL for "no such endpoint", and I really think we ought to do that here too.

c) Once the above two are resolved, rebase this down to a single commit.  The comments should be along the lines of "fixes #<id> <synopsis>"  (yes, this is a new style).

vseryakov:
during rebase/squash i messed up my local git, so recreating new pull request

PR #409, opened 2015-05-02
Multiple buffer message receiving with nanomsg allocated buffers

baxtersa:
Sent multiple buffer messages should be received in the same number of buffers it was sent with when allowing nanomsg to allocate the buffers for them.

### DIFF
--- a/src/core/global.c
+++ b/src/core/global.c
@@ -758,6 +758,35 @@ int nn_sendmsg (int s, const struct nn_msghdr *msghdr, int flags)
         nn_msg_init_chunk (&msg, chunk);
         nnmsg = 1;
     }
+    else if (msghdr->msg_iovlen > 1 && msghdr->msg_iov [0].iov_len == NN_MSG) {
+        /*  Compute the total size of the message. */
+        sz = 0;
+        for (i = 0; i != msghdr->msg_iovlen; ++i) {
+            iov = &msghdr->msg_iov [i];
+            if (nn_slow (iov->iov_len != NN_MSG)) {
+                errno = EINVAL;
+                return -1;
+            }
+            if (nn_slow (!iov->iov_base)) {
+                errno = EFAULT;
+                return -1;
+            }
+            iov->iov_len = strlen (*(const char**) iov->iov_base) + sizeof (uint8_t);
+            sz += iov->iov_len;
+        }
+        
+        /*  Create a message object from the supplied scatter array. */
+        nn_msg_init (&msg, sz);
+        sz = 0;
+        for (i = 0; i != msghdr->msg_iovlen; ++i) {
+            iov = &msghdr->msg_iov [i];
+            memcpy (((uint8_t*) nn_chunkref_data (&msg.body)) + sz,
+                    *(uint8_t**) iov->iov_base, iov->iov_len);
+            sz += iov->iov_len;
+        }
+        
+        nnmsg = 1;
+    }
     else {
 
         /*  Compute the total size of the message. */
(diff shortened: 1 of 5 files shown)


gdamore:
While I appreciate the desire for a scatter & gather, I think this design has several flaws.  Most importantly it changes the wire protocol in a surprising way - messages exchanged using this new semantic cannot be exchanged with other peers. 

Furthermore I'm opposed to adding additional boundaries into the protocol or the API.  If know a priori you can use explicit receive vectors that are reallocated. 

I'd support a design that used fixed sizes and then used an iov with NN_MSG for the last chunk (dynamically allocating "the rest"). 

Also remember that nanomsg is a binary protocol and therefore use of ASCII strings in the protocol is ill advised. It's inefficient and the message sizes are variable if you do that.  Fixed binary numbers are far better. 

That said I repeat that I'm unlikely to be supportive of a change to the wire protocol for a feature like this.

baxtersa:
Thanks for the feedback.

Discarding the reliance on ASCII strings, how would you feel about messages prepended with the size of the following binary stream?  Would you contend that it's the user's responsibility to construct the data in that way and not nanomsg's?

I obviously wouldn't want to change the wire protocol for the existing message types, but maybe a more self describing nn_iovec struct that kept an array of buffer lengths for each iov would allow additional information to be sent with the message so that there's enough information on receive to decompose the single squished buffer into its components.

Such a message could still be exchanged with peers using fixed size iovs, but would allow the flexibility I'm trying to accomplish with this pull request.

baxtersa:
@gdamore , if you could, let me know if this is a more promising path to go down.  I'd like to be able to work something out, but if this is a hopeless endeavor I'll stop pursuing it as a pull request and bring it into my project's code.

If you agree with the potential benefits and think it's worth spending more time on, I'll add some more tests and protection on the nn_recvmsg side against reading bogus buffer sizes.

gdamore:
> On May 6, 2015, at 2:12 PM, Sam Baxter <notifications@github.com> wrote:
> 
> @gdamore <https://github.com/gdamore> , if you could, let me know if this is a more promising path to go down. I'd like to be able to work something out, but if this is a hopeless endeavor I'll stop pursuing it as a pull request and bring it into my project's code.
> 
> 

My feeling is that this is the right approach (handling it in your application) — I don’t think the benefits of this are sufficiently generally useful to warrant changing or extending the wire protocol.

Btw, what is really your motivation here?  (In other words, why and how is splitting the data into fixed sized “portions” useful to you?)  Normally the cases I’ve seen for this are for network protocols where you need to split off headers, and the splitting is usually done to facilitate data alignment on things such as NICs.  (Most common case is where ethernet is misaligned with a 14 byte header, but we’d like to have the following IP header and frame properly double-word aligned.)  I’m having a little difficulty understanding the use case in application code though.

	- Garrett

baxtersa:
> Btw, what is really your motivation here?

My motivation is my intuition about the ease of use of this API.  It seems natural to me that if you can send a message via multiple vectors, you should be able to receive the message in the same format as which it was sent (and without having to know the length of the incoming vectors or alternatively having to reallocate/copy).

> (In other words, why and how is splitting the data into fixed sized “portions” useful to you?)

We're trying to use nanomsg to send commands across instances of our application.  We think it would be convenient to be able to send discretized messages containing a variety of metadata and potentially multiple data buffers as separate vectors.  This would eliminate the need in client code to parse an unknown bitstream for metadata tags and could ease debugging of receiving such messages.

src/core/global.c
```diff
@@ -886,6 +922,22 @@ int nn_recvmsg (int s, struct nn_msghdr *msghdr, int flags)
         *(void**) (msghdr->msg_iov [0].iov_base) = chunk;
         sz = nn_chunk_size (chunk);
     }
+    else if (msghdr->msg_iovlen > 1 && msghdr->msg_iov [0].iov_len == NN_MSG) {
+        chunk = nn_chunkref_getchunk (&msg.body);
+        
+        sz = 0;
+        for (i = 0; i != msghdr->msg_iovlen; ++i) {
+            iov = &msghdr->msg_iov [i];
```
baxtersa: We can add some more checks in here to assert things such as iov->iov_len == NN_MSG before we read the length prefix, or ensuring that our accumulated iov_len's aren't > our total chunk size.

Issue #404, opened 2015-04-26
nn_bind does not fail when port is already in use


lekv:
I noticed that nn_bind seems to succeed even when the tcp port is already in use by another program/instance of my program. For my own code it happens with REQ/REP sockets, however it also seems to work with the local_thr/remote_thr test programs. You can start local_thr "tcp://*:5000" 256 256 twice. If you run remote_thr, it will first connect to the first instance, on the second run it will connect to the second instance.

For my own code I added explicit checks to nn_bind()'s return value to ensure it's not caused by disabled assertions in release builds.

Is this behavior expected?

This is the relevant code from my own program:
```cpp
    int network_socket = nn_socket( AF_SP_RAW, NN_REP );
    int ret = nn_bind( network_socket, "tcp://*:5000" );
    if ( ret < 0 )
    {
        std::cout << "Could not bind: " << nn_strerror( errno ) << std::endl;
    }
    else
    {
        std::cout << "Bind successful" << std::endl;
    }
    assert( ret >= 0 );
    (void)ret;
```

vseryakov:
yes, this is expected, bind is done in the background and it will keep trying forever, no way right now to know if it failed or not, there was a proposal to export errno function for an endpoint but it never got through

gdamore:
The errno proposal is outstanding, and is likely to be integrated with a few minor changes.

gdamore:
I'm closing this for now; there are no code changes needed for this.

PR #410, opened 2015-05-02
Adds support for nn_geterror function to return the last errno from an endpoint

vseryakov:


baxtersa:
Should this be implemented for ipc transport as well?  It's a similar connection-base transport where a failed bind attempt still tries to reconnect in the background, correct?

Also, I've cherry-picked this pull-request into some local code, and I get crashes if the first bind fails and then I call nn_close on the socket because even with this nn_geterror function, there's no way to shut down the reconnect timer in the background.  Is that perhaps worth implementing alongside this function?  Was this pull-request intended to address issue #404?

vseryakov:
Good point about the nn_close, this needs to be handled but i would have this merged first, then would work on ipc transport and deal with reconnect/shutdown, so we do not have multiple feqatures/fixes in one PR.

baxtersa:
Sounds good to me.  I'm going to look into the nn_close crash locally and see how far I can get on a fix.  I'd welcome any and all help though, as this is my first dive into nanomsg's fsm logic :).

baxtersa:
Thanks @vseryakov !  This seems to be working.

I've got ipc transport working, so once this is merged I'll open up a pull-request for that.

baxtersa:
As I commented in #413 , this doesn't seem to be working on Windows.  Are you already looking into this?  I'm getting started on trying to figure this out for IPC now.

gdamore:
I've not had time yet.  Perhaps another contributor can help?

vseryakov:
This PR only about TCP case, let's not mix even if they are similar in logic. if TCP transport works fine on Windows i think this Pr can be merged.

baxtersa:
I have a separate pull request open for IPC transport, so I agree with only dealing with TCP logic here.  But unfortunately it looks like TCP isn't working on windows.  Once it is, I say we merge some things.

vseryakov:
can you paste the errors here please, i do have win vm but it will take some time top make it work and compile nn, maybe i can figure out the error remotely, tnx

baxtersa:
The error isn't too enlightening as to what the problem is.  We're failing 

nn_assert(nn_geterror (sc, rc) != 0);

on line 130 of the tcp.c test file.  Disabling that assert, we seem to still fail the other nn_assert (nn_geterror (_,_) _) tests you've added.  It seems like Windows Socket errno gets set differently than Unix errno.  If I had to guess, I'd start looking at usock_win.inc, because that's where the low level system calls live, but that's just a hunch.

vseryakov:
just one more thing befor ei will jump into spawning win vm, can you print nn_geterror before the asserts to see the actual values. also, can you increase the sleep to a second or more to see if this is not a test race condition, i am pretty sure nn_errno and etc works the same way and i am not dealing with the errno directly. Thank you

baxtersa:
Well, good news is changing the sleep time to a 1500 milliseconds seems to fix this issue for tcp.  Bad news is 1500 milliseconds is a long wait, and this doesn't seem to fix the issues with my separate ipc pull request.  If you could change those sleep times for this pull request this should be ready to merge.

vseryakov:
the timeouts i chose completely out of the blue, because bind is happening in the background this behavior is normal. i will update the test.

gdamore:
This PR has three separate commits, one of which was cherry picked with #468 

Please rebase this to master, and make it just a single commit fixing just the single requested issue (the geterror call), and then I will look again.

src/core/sock.c
```diff
@@ -535,6 +535,34 @@ int nn_sock_rm_ep (struct nn_sock *self, int eid)
...
+          it != nn_list_end (&self->eps);
+          it = nn_list_next (&self->eps, it)) {
+        ep = nn_cont (it, struct nn_ep, item);
+        if (ep->eid == eid) {
+            err = ep->last_errno;
+            break;
+        }
+        ep = NULL;
+    }
+    /*  The endpoint doesn't exist. */
```
gdamore: Its interesting that this error is indistinguishable from an error on the ep itself.  Is it valuable to be able to separate this case from some other -EINVAL case on the socket?

src/transports/tcp/ctcp.c
```diff
@@ -365,8 +366,12 @@ static void nn_ctcp_handler (struct nn_fsm *self, int src, int type,
                 nn_epbase_clear_error (&ctcp->epbase);
                 return;
             case NN_USOCK_ERROR:
-                nn_epbase_set_error (&ctcp->epbase,
```
gdamore: TCP changes already integrated,, modulo the _set_error() call above.

gdamore:
I'm going to close this PR for now.  Please feel free to reopen when just the errorno functionality is ready, or open a new PR.

Issue #411, opened 2015-05-05
Rebinding an inproc socket crashes


baxtersa:
This seems to be a false assumption in the state machine handling, potentially because either we don't reset our state when the shutdown occurs, or because we are incorrect in our assumption about the state we are in.

gdamore:
Rebinding an inproc socket after the originally bound socket shuts down crashes unless all other connected sockets shutdown before rebind.

PR #416, opened 2015-05-22
Fixes crash binding an inproc socket after originally bound socket closes.

baxtersa:
Address issue #411 .

### DIFF
--- a/src/transports/inproc/binproc.c
+++ b/src/transports/inproc/binproc.c
@@ -232,6 +232,9 @@ static void nn_binproc_handler (struct nn_fsm *self, int src, int type,
                 nn_fsm_bad_action (binproc->state, src, type);
             }
 
+        case NN_BINPROC_SRC_SINPROC:
+            return;
+                
         default:
             nn_fsm_bad_source (binproc->state, src, type);
         }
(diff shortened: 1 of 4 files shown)


gdamore:
Running behind on analysis of PRs these days.  But I promise to look this week.

baxtersa:
Not a problem.  Even if this doesn't get merged this week, if I could get some feedback that would be great.

gdamore:
At first blush this looks OK.  To be honest, I'm not that familiar, immediately, with Martin's inproc FSM implementation.  I'd feel better if we could get another reviewer here, but as I said, it *looks* ok to me.

gdamore:
Admittedly what would be enormously helpful here, and is missing from all of this code, are state diagrams, showing each of the various states and transitions.  I'll probably have to draw those out to properly support this code base long term.

gdamore:
I've merged this, but I've rebased your code, fixed the commit comments, and addressed some whitespace problems in your code that showed up when I looked at this in the command line.

The tests passed, so its upstream now.  I'm still not 100% comfortable with the inproc code -- it clearly suffers from some race conditions -- in fact much of libnanomsg feels thread-unsafe to me.  But this is clearly an improvement.

PR #417, opened 2015-05-26
fixes #194 ETIMEDOUT not returned when deadline passes in surveyor socket

### LINKED ISSUE #194: ETIMEDOUT not returned when deadline passes in surveyor socket
According to my understanding of the documentation, when a NN_SURVEYOR socket that is blocked on nn_recv reaches NN_SURVEYOR_DEADLINE, the socket returns with -1 and nn_errono is set to ETIMEDOUT.

(Actually, the documentation says that `receive` returns with ETIMEDOUT. But that doesn't make sense, so let's go with what I'm understanding.)

Instead, `EFSM` is returned. I traced it as much as I could, I can't offer any insight as to why.

The problem is that EFSM is an error that should crash the program, whereas ETIMEDOUT is an expected condition that I'm looking/hoping for.

Let me know if this isn't helpful, but here is a complete example (modified from Tim's tutorial):

EDIT: This was tested on the latest commit from master branch on both OS X and Windows 8 64bit.

```C
#include <assert.h>
#include <libc.h>
#include <stdio.h>
#include <nanomsg/nn.h>
#include <nanomsg/survey.h>

#define SERVER "server"
#define CLIENT "client"
#define DATE   "DATE"

int server (const char *url)
{
  int sock = nn_socket (AF_SP, NN_SURVEYOR);
  assert (sock >= 0);
  assert (nn_bind (sock, url) >= 0);
  sleep(1); // wait for connections
  int sz_d = strlen(DATE) + 1; // '\0' too
  printf ("SERVER: SENDING DATE SURVEY REQUEST\n");
  int bytes = nn_send (sock, DATE, sz_d, 0);
  assert (bytes == sz_d);
  while (1)
    {
      char *buf = NULL;
      int bytes = nn_recv (sock, &buf, NN_MSG, 0);

// This does not work either. It's how the original tutorial from Tim tried it.
//      if (bytes == ETIMEDOUT) {
//        printf("ETIMEDOUT.\n\n");
//        break;
//      }
      if (bytes >= 0)
      {
        printf ("SERVER: RECEIVED \"%s\" SURVEY RESPONSE\n", buf);
        nn_freemsg (buf);
      } else if (nn_errno() == ETIMEDOUT) {
        printf("ETIMEDOUT.\n\n");
        break;
      } else {
        printf("ERROR %s\n", nn_strerror(nn_errno()) );
        break;
      }
    }
  return nn_shutdown (sock, 0);
}

int client (const char *url, const char *name)
{
  int sock = nn_socket (AF_SP, NN_RESPONDENT);
  assert (sock >= 0);
  assert (nn_connect (sock, url) >= 0);
  while (1)
    {
      char *buf = NULL;
      int bytes = nn_recv (sock, &buf, NN_MSG, 0);
      if (bytes >= 0)
        {
          nn_freemsg (buf);
          char *d = "some date";
          int sz_d = strlen(d) + 1; // '\0' too
          printf ("CLIENT (%s): SENDING DATE SURVEY RESPONSE\n", name);
          int bytes = nn_send (sock, d, sz_d, 0);
          assert (bytes == sz_d);
        }
    }
  return nn_shutdown (sock, 0);
}

int main (const int argc, const char **argv)
{
  if (strncmp (SERVER, argv[1], strlen (SERVER)) == 0 && argc >= 2)
    return server (argv[2]);
  else if (strncmp (CLIENT, argv[1], strlen (CLIENT)) == 0 && argc >= 3)
    return client (argv[2], argv[3]);
  else
    {
      fprintf (stderr, "Usage: survey %s|%s <URL> <ARG> ...\n",
               SERVER, CLIENT);
      return 1;
    }
}
```

And for convenience:

```sh
gcc survey.c /usr/local/lib/libnanomsg.a -o survey
./survey server ipc:///tmp/survey.ipc & server=$!
./survey client ipc:///tmp/survey.ipc client0 & client0=$!
./survey client ipc:///tmp/survey.ipc client1 & client1=$!
./survey client ipc:///tmp/survey.ipc client2 & client2=$!
sleep 3
kill $server $client0 $client1 $client2
```

rodbzro:
Couldn't find a way to determine that surveyor socket has timed out, so here goes new definitions.

I've read a plethora of fsm/timer state combinations to uniquely identify if timeout has occured but one changes states with the other and the original timeout call isn't stored anywhere, so I see no way to backtrace if their current "out of survey" state has come from a deadline timeout.

Anyway, would be glad to know if there is another way without new definitions.

### DIFF
--- a/src/protocols/survey/surveyor.c
+++ b/src/protocols/survey/surveyor.c
@@ -56,6 +56,8 @@
 
 #define NN_SURVEYOR_SRC_DEADLINE_TIMER 1
 
+#define NN_SURVEYOR_TIMEDOUT 1
+
 struct nn_surveyor {
 
     /*  The underlying raw SP socket. */
@@ -76,6 +78,9 @@ struct nn_surveyor {
 
     /*  Protocol-specific socket options. */
     int deadline;
+
+    /*  Flag if surveyor has timed out */
+    int timedout;
 };
 
 /*  Private functions. */
@@ -128,6 +133,7 @@ static void nn_surveyor_init (struct nn_surveyor *self,
     nn_timer_init (&self->timer, NN_SURVEYOR_SRC_DEADLINE_TIMER, &self->fsm);
     nn_msg_init (&self->tosend, 0);
     self->deadline = NN_SURVEYOR_DEFAULT_DEADLINE;
+    self->timedout = 0;
 
     /*  Start the state machine. */
     nn_fsm_start (&self->fsm);
@@ -236,8 +242,13 @@ static int nn_surveyor_recv (struct nn_sockbase *self, struct nn_msg *msg)
     surveyor = nn_cont (self, struct nn_surveyor, xsurveyor.sockbase);
 
     /*  If no survey is going on return EFSM error. */
-    if (nn_slow (!nn_surveyor_inprogress (surveyor)))
-       return -EFSM;
+    if (nn_slow (!nn_surveyor_inprogress (surveyor))) {
+        if (surveyor->timedout == NN_SURVEYOR_TIMEDOUT) {
+            surveyor->timedout = 0;
+            return -ETIMEDOUT;
+        } else
+            return -EFSM;
+    }
 
     while (1) {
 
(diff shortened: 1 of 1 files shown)


gdamore:
This looks very good.  Can you please edit the commit comments to match the bug synopsis though?  Like this:

fixes #194 ETIMEDOUT not returned when deadline passes in surveyor socket

This can be done simply via commit --amend, and resubmitting or updating the PR.  Then I will merge.  (Or if you prefer, I can do the above myself, but you'll have to merge your changes in that case as the git history will differ.)

rodbzro:
Is this ok?

gdamore:
Yes.  Merging shortly.

gdamore:
Ah, but this fails the survey test.  I'm going to go ahead and fix that too.

PR #430, opened 2015-06-24
Allowed nn_sleep to be exported on Windows.

vsajip:
At present, the `nanomsg.dll` built on Windows does not export the `nn_sleep` API, even though there is an implementation on Windows. This appears to be just because the symbol isn't marked for export using `NN_EXPORT`.

### DIFF
--- a/src/utils/sleep.c
+++ b/src/utils/sleep.c
@@ -20,6 +20,7 @@
     IN THE SOFTWARE.
 */
 
+#include "../nn.h"
 #include "sleep.h"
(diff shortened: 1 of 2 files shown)


gdamore:
fixed in 
 commit e3588bc987f60fa841c153df89a70c7a3f1eb02c
Author: Vinay Sajip <vinay_sajip@yahoo.co.uk>
Date:   Wed Jun 24 00:57:39 2015 +0100

    Allowed nn_sleep to be exported on Windows.

Issue #446, opened 2015-07-26
efd_win.inc global critical section should be mutex


pijyoi:
Although the code creating the socketpair takes care to release the Event upon failure, it is still possible for the process to crash outside of the control of the library. 

This could cause the Event to be permanently held until all software using nanomsg blocking on the Event have been killed. A more robust way would be to use a Mutex which gets abandoned if the process crashes.

zhaozg:
releate with https://github.com/nanomsg/nanomsg/issues/450 ?

PR #449, opened 2015-08-12
Address issue #446

qbradley:
This addresses issue #446.

Use a global mutex instead of a global event to protect the file descriptor creation on Windows.
An acquired mutex will be released when the thread that owns it exits without releasing the mutex explicitly, for example when a process is terminated.
The caveat is that any other process waiting for the mutex will get WAIT_ABANDONED instead of WAIT_OBJECT_0 returned from WaitForSingleObject. However, the mutex will be acquired by the previously waiting process. This is documented here near the remarks on WAIT_ABANDONED return value: https://msdn.microsoft.com/en-us/library/windows/desktop/ms687032

I tested this change by running the unit tests.
I further tested it by running "nanocat --req --bind tcp://127.0.0.1:8000 --data abcd -Q" instrumented to delay while the mutex was held. I terminated the one that had acquired the mutex and observed the other process continue seeing the WAIT_ABANDONED return code.

### DIFF
--- a/src/core/sock.c
+++ b/src/core/sock.c
@@ -824,11 +824,13 @@ static void nn_sock_shutdown (struct nn_fsm *self, int src, int type,
     }
     if (nn_slow (sock->state == NN_SOCK_STATE_STOPPING_EPS)) {
 
+        /* We are no longer interested in pipe events */
+        if (src == 0)
+            return;
+
         /*  Endpoint is stopped. Now we can safely deallocate it. */
-        if (!(src == NN_SOCK_SRC_EP && type == NN_EP_STOPPED)) {
-            fprintf (stderr, "src=%d type=%d\n", (int) src, (int) type);
-            nn_assert (src == NN_SOCK_SRC_EP && type == NN_EP_STOPPED);
-        }
+        nn_assert (src == NN_SOCK_SRC_EP && type == NN_EP_STOPPED);
+
         ep = (struct nn_ep*) srcptr;
         nn_list_erase (&sock->sdeps, &ep->item);
         nn_ep_term (ep);
@@ -935,6 +937,8 @@ static void nn_sock_handler (struct nn_fsm *self, int src, int type,
         default:
 
             /*  The assumption is that all the other events come from pipes. */
+            nn_assert (src == 0);
+
             switch (type) {
             case NN_PIPE_IN:
                 sock->sockbase->vfptr->in (sock->sockbase,
--- a/src/utils/efd_win.inc
+++ b/src/utils/efd_win.inc
@@ -56,15 +56,16 @@ int nn_efd_init (struct nn_efd *self)
 
     /*  This function has to be enclosed in a system-wide critical section
         so that two instances of the library don't accidentally create an efd
-        crossing the process boundary. 
-        CAUTION: This critical section has machine-wide scope. Thus, it must
-        be properly exited even before crashing the process by an assertion. */
-    sync = CreateEvent (&sa, FALSE, TRUE, "Global\\nanomsg-port-sync");
+        crossing the process boundary. */
+    sync = CreateMutex (&sa, FALSE, "Global\\nanomsg-port-sync");
     win_assert (sync != NULL);
 
     /*  Enter the critical section. */
     dwrc = WaitForSingleObject (sync, INFINITE);
-    nn_assert (dwrc == WAIT_OBJECT_0);
+    /* WAIT_OBJECT_0 indicates we have acquired the mutex
+       WAIT_ABANDONED indicates we have acquired the mutex, but that the
+       thread that owned the mutex last exited without releasing it first. */
+    nn_assert (dwrc == WAIT_OBJECT_0 || dwrc == WAIT_ABANDONED);
 
     /*  Unfortunately, on Windows the only way to send signal to a file
         descriptor (SOCKET) is to create a full-blown TCP connecting on top of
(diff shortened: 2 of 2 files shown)


wirebirdlabs:
It may be unrelated to this pull request, but this commit triggered the `tcp_shutdown` test to fail on all three Windows builds: https://ci.appveyor.com/project/gdamore/nanomsg/build/1.0.40

Here is a search link that shows all Issues that mention the keyword `tcp_shutdown`: https://github.com/nanomsg/nanomsg/search?q=tcp_shutdown&type=Issues

gdamore:
I’m uncomfortable proceeding with this pull request given that the tests all failed, and failed in the same place.

I’d like someone more familiar with Windows to have a look at it.

	- Garrett

wirebirdlabs:
@qbradley can you run `tcp_shutdown` directly and paste the failure mode (probably a stack dump)?

qbradley:
I will try to repro and will report back. When I built the RUN_TESTS project locally all tests passed. If it is an intermittent failure I'll also see if it fails without my changes.

qbradley:
I ran tcp_shutdown.exe 10 times. It passed 5 times and failed 5 times as follows:

```
src=0 type=33988
Assertion failed: src == NN_SOCK_SRC_EP && type == NN_EP_STOPPED (C:\Users\qbrad_000\dev\nanomsg\src\core\sock.c:830)
```

Then I reverted my changes and ran tcp_shutdown.exe 10 times. It passed 7 times and failed 3 with the same error.

More than likely this is an existing intermittent failure. Just in case my change increases the likelihood of its occurrence I'll try and dig a little deeper in the tcp_shutdown.exe failure.

wirebirdlabs:
> an existing intermittent failure

Solid analysis. See link above for issues mentioning `tcp_shutdown`.

gdamore:
Thanks for working on this.

qbradley:
I made some progress reducing the complexity of the repro. Modifying main of tcp_shutdown.c as follows still repros:

```c
int main ()
{
    int sb;
    struct nn_thread thread;

    /*  Test race condition of sending message while socket shutting down  */

    sb = test_socket (AF_SP, NN_PUSH);
    test_bind (sb, SOCKET_ADDRESS);

    nn_thread_init (&thread, routine2, NULL);
    nn_atomic_init(&active, 1);

    while (active.n) {
        (void) nn_send (sb, "hello", 5, NN_DONTWAIT);
    }

    nn_thread_term (&thread);
    nn_atomic_term(&active);

    test_close (sb);

    return 0;
}
```

Although sometimes it takes many tries to fail.

qbradley:
The assert is because a NN_PIPE_OUT event is received after the socket has started stopping.
I think this is a bad assert that assumes only endpoint events will be received after shutdown.
This failure is not windows specific. I get the same failure runningn the ipc_shutdown on ubuntu.

A pipe out or in event during shutdown is not a reason for concern, I think. We won't read or write to the pipe anyway, so we can and should ignore them.

wirebirdlabs:
> A pipe out or in event during shutdown is not a reason for concern, I think. We won't read or write to the pipe anyway, so we can and should ignore them.

Seems that way... but still not quite. Ignore this assertion, and the crash moves to a more critical place -- accessing freed memory after the socket has been destroyed.

The underlying cause is within `nn_sock_recv()` and `nn_sock_send()` since these calls [release the mutex on the socket] (https://github.com/nanomsg/nanomsg/blob/master/src/core/sock.c#L670-L677), allowing a contentious `nn_close()` from a separate thread to gain the mutex and destroy the socket. By the time send/recv tries to regain the mutex, it dereferences a freed ptr :-(

After spending time trying to patch this issue with a few different fixes, we're relegated to the only solid workaround -- ensure `nn_close()` is never called during a blocking send or recv based on usage at the framework/application layer, either through 1) guaranteeing any one socket access remains single-threaded, or 2) introducing additional mutexing at the higher layer.

That said, this advice is based on memory from months ago, and I'd be glad to help brainstorm new solutions.

qbradley:
I have updated my branch with a proposed fix.

What I meant by ignoring was to not continue with execution but return from the handler. There is no work to do for a pipe event when in the stopping state.

The subsequent code should only be executed for an event originating from the endpoint when in the stopping state.

I added an assert and a comment to make what I believe to be the correct intention of the code more clear.

Let me know what you think.

gdamore:
tcp_shutdown tests still failing, so I don't think your fix addresses it.

If I find some time in the next several days/week, I'll see if  I can noodle this out.  Typically in other situations like this I've used a flag, protected by a lock, to indicate when a connection or resource was being torn down.

The internal architecture of libnanomsg makes this a little harder right now.  Sadly.

src/core/sock.c
```diff
@@ -824,11 +824,13 @@ static void nn_sock_shutdown (struct nn_fsm *self, int src, int type,
     }
     if (nn_slow (sock->state == NN_SOCK_STATE_STOPPING_EPS)) {
 
+        /* We are no longer interested in pipe events */
+        if (src == 0)
+            return;
```
gdamore: I'm looking at this, and I think maybe we miss the cleanup that would occur in finish2.  Probably instead of return here, we should just go to finish2.

gdamore:
So, this PR, only the change for efd_win.inc is appropriate.  I'm uncomfortable with the other changes, and I think we may have addressed those problems another way.

gdamore:
I've taken some of your code, and made it better.  I used a non-INFINITY timeout for the mutex, because frankly its better to eventually give up and report an error to the caller than to wait forever.

This was integrated with several other windows specific fixes in ab699df74978645ebe5bb53f0b602ba4bf32ed4b

PR #454, opened 2015-08-14
fix nn_hash items

yan97ao:
nn_hash.items will increase by nn_hash_insert but won't decrease


gdamore:
On paper looks good.  I need to check to see if there are locking considerations here.  I don't think there are, but give me some time for analysis please.

gdamore:
fixed in 968e90b685bcebd42bd05cf52c3f4e43cf83c8ba

PR #459, opened 2015-09-02
New NN_RCVMAXSIZE socket option

bnewbold:
See previous discussion: http://www.freelists.org/post/nanomsg/issue-with-received-message-size,1

"-1 for no limit" matches zeromq's MAXMSGSIZE.

Without this feature, a trivial DOS attack on an open tcp socket is to send send a bogus message header (aka, uint64_t payload size), which will result in libnanomsg trying to allocate an impossibly large amount of memory, failing, and crashing.

Here NN_RCVMAXSIZE is of "int" type, while, eg, TCP's payload length can be up to uint64_t size. I think this is sort of justifiable because the amount of memory that can be allocated is presumably limited by the local bitwidth, but really i'm just too lazy to create a new socket option type. Note that if uint64_t is chosen as the option type then "-1 for no limit" breaks, while int64_t doesn't match the full possible TCP size (uint64_t). Also, uint64_t is just what happens to have been chosen as the payload size limit for TCP; in theory future transports could define even larger limits.

These patches only contain an implementation for the tcp transport. Implementations for tcpmux and ipc look pretty trivial and i'm happy to supply them if the tcp transport looks good. Other transports might be more involved.

I considered copying/caching the rcvmaxsize option in the pipebase or stcp structs (aka, at the per-connection level) so `nn_pipebase_getopt()` wouldn't need to get called on every message received, but that seemed like premature optimization.


wirebirdlabs:
This feature is most excellent; I will certainly use it.

src/core/sock.c
```diff
@@ -313,6 +314,11 @@ static int nn_sock_setopt_inner (struct nn_sock *self, int level,
                 return -EINVAL;
             dst = &self->rcvbuf;
             break;
+        case NN_RCVMAXSIZE:
+            if (nn_slow (val <= -1))
```
gdamore: Doesn't this mean you can't actually set it to the documented -1 for unlimited?

wirebirdlabs:
> I think this is sort of justifiable because the amount of memory that can be allocated is presumably limited by the local bitwidth, but really i'm just too lazy to create a new socket option type

No chance this is an issue of being too lazy. I'll posit, rather, than anyone who gripes about this limit is the one too lazy, and is making some gravely suboptimal decisions about the size of messages being sent.

In the future, your implementation could be upgraded if pedantic consistency with `uint64_t` is deemed worth it, but for now, `int` is an entirely reasonable choice.

tests/tcp.c
```diff
@@ -178,6 +178,28 @@ int main ()
...
+    opt = 4;
+    nn_setsockopt(sb, NN_SOL_SOCKET, NN_RCVMAXSIZE, &opt, sizeof (opt));
+    nn_sleep (100);
+    test_send (s1, "ABC");
+    test_recv (sb, "ABC");
+    test_send (s1, "0123456789012345678901234567890123456789");
+    void * dummy_buf;
+    int s;
+    s = nn_recv(sb, dummy_buf, NN_MSG, NN_DONTWAIT);
+    if (s != -1 || errno != EAGAIN) {
```
gdamore: This should probably be expressed as an nn_assert() to match the rest of the style in this test program.

src/core/sock.c
```diff
@@ -122,6 +122,7 @@ int nn_sock_init (struct nn_sock *self, struct nn_socktype *socktype, int fd)
     self->linger = 1000;
     self->sndbuf = 128 * 1024;
     self->rcvbuf = 128 * 1024;
+    self->rcvmaxsize = 512 * 1024 * 1024;
```
gdamore: To be honest, this seems rather excessive.  I'd set by default to either -1 (unlimited), or something much smaller (1 MB).  Mangos sets this to 1MB by default.

512MB is so huge that it offers virtually no protection from abuse.

src/core/sock.c
wirebirdlabs: Seems that way. What would expected behavior be? Any integer <=-1 disables? Or only -1 disables, and <=-2 returns EINVAL? I don't have strong opinions, except to choose the lesser surprising option.

gdamore:
Mostly looks good.  Admittedly I'd like to see the commits collapsed into a single commit, with a comment of style matching the other most recent comments ("fixes #459 New NN_RCVMAXSIZE socket option")

src/core/sock.c
wirebirdlabs: I might tend to accept anything as <=-1 as acceptable, then changing documentation to say "any negative integer disables the limit"

src/core/sock.c
wirebirdlabs: Agree. When seeing this feature, this is the first priority I considered -- setting a decently robust/secure default threshold. 1MB feels acceptable, but it could be argued up to ~10MB or so. 512MB is too high a default today in 2015 :-)

src/core/sock.c
```diff
@@ -397,6 +403,9 @@ int nn_sock_getopt_inner (struct nn_sock *self, int level,
         case NN_RCVBUF:
             intval = self->rcvbuf;
             break;
+        case NN_RCVMAXSIZE:
```
wirebirdlabs: Wait a minute... see comments above. It seems lines 317-321 should be deleted.

tests/tcp.c
wirebirdlabs: Rather, here is an example of a conventional assert for nanomsg tests: https://github.com/nanomsg/nanomsg/blob/master/tests/tcp.c#L78-L80

doc/nn_serial.txt
```diff
@@ -0,0 +1,55 @@
+nn_serial(7)
```
wirebirdlabs: This looks very cool! But seems to have snuck in on the wrong commit :-)

src/core/sock.c
gdamore: Well, I prefer to use specific -1 as the documented case, in case we ever want to have anything else in the future (unlikely, I know).  It will also catch incorrect attempts to set some value that is too large (via rollover).

src/core/sock.c
gdamore: I'd stick with 1MB.   Even 10MB is too big.   In some circumstances 1MB is too big.  (Imagine opening up a 1000 sockets and sending one of these to each one.  Congratulations -- you've allocated 1GB on the server.)

For internet facing frames, I'd choose a much smaller value < 100K I think, and also limit the number of connections overall.  We lack the ability to limit.  (The limit could be imposed per peer, but that means quite a bit more work.)

doc/nn_serial.txt
gdamore: Yep, glad you caught that @wirebirdlabs .  I was speaking with the author about his plans for nn_serial. :-)

src/core/sock.c
bnewbold: This was a typo, good catch. Should be `(val < -1)`. I think "negative but not -1" should be reserved and thus an EINVAL. I'll fix and also add a test.

bnewbold:
Pushed changes:

- removed nn_serial.txt (oops!)
- refactored tests/tcp.c to follow style of other tests. Added test that "-2" as a value is invalid.
- switched to 1MB (1024kB) as the new default max. I think this is a good balance, given that the socket will just mysteriously close for the sender end.
- reworked stcp.c code a bit to use "opt" and "opt_sz" temporary variables; seemed cleaner this way.
- fix the -1/-2 value check

I'll send an email to the list, as the new max size could be counter-intuitive to folks ("why isn't my 3MB message going through if TCP payload header is uint64_t long?). Should I also update the, eg, nn_recv and nn_recvmsg man pages?

bnewbold:
I just pushed one more small change:

```
-                    if (opt >= 0 && size > opt) {
+                    if (opt != -1 && size > opt) {
```

Now, when checking, the option must explicitly be -1 for TCP to allow any message size (instead of just any negative number). This complements the earlier change that negative numbers below -1 could be reserved for some other purpose.

wirebirdlabs:
> switched to 1MB (1024kB) as the new default max. I think this is a good balance, given that the socket will just mysteriously close for the sender end.

This is the reason I've desired to add a Close Handshake to the nanomsg protocol. (you can search the term "Close Handshake" in this repo's issues and also on the mailing list archive to see more thoughts) This is a formalized message in RFC 6455 for WebSocket, and I feel something similar would help applications using nanomsg sockets to react more sanely to the condition where a peer declares it is gone (in this case, for breaking protocol)

gdamore:
I still think close handshake is not necessary and doesn't really solve anything.  It helps one remote end understand that a connection was rejected rather than silently dropped, but serves little purpose other than that -- since indeed there is no guarantee that the close itself will be delivered.

(This then gets into a need to replicate the TCP FIN close mechanism, which is ... ugly.)

gdamore:
So can you please rebase these into a single commit with the commit message requested?  Once done, I can merge it into master.

bnewbold:
Pushed

wirebirdlabs:
> Pushed

:-)

PR #464, opened 2015-09-09
fix #462: documentation build instructions outdated

### LINKED ISSUE #462: documentation build instructions outdated
`doc/README` says all I need to do is "make doc", but that's a no-op on my Debian (linux) development machine, and manpages don't end up getting installed when I build/install.

Maybe I just don't know something about autotools?

There's also a `make diagrams` target, which does have some documentation in the header.

bnewbold:


gdamore:
I think the --enable-doc should only be required when *formatting* man pages after they have been changed.  Most users needn't worry about this, which spares them the effort of having xmlto or other tooling.  So I'd like to have this documented separately from the primary user-facing installation document.

README
```diff
@@ -22,7 +22,7 @@ Build it
 2.  Unpack it.
 3.  Go to the root directory of the package.
 4.  sh autogen.sh
-5.  ./configure
+5.  ./configure (with `--enable-doc` for html and man pages)
```
gdamore: Specifically, in my earlier comments, I'd leave this annotation out.

bnewbold:
I agree it would be nice to not need dependencies installed when building, but I started with the user-facing install docs (README) and found it confusing and counter-intuitive that man pages were not installed when installing from source following the instructions. If they were somehow installed without reformatting then perhaps we could leave the mention out of the README.

Should we commit the formatted manpages in the repo (like with txt/xml under ./rfc/) so they can still be installed? That doesn't seem like a great option.

gdamore:
I actually do think we ought to commit the formatted pages.

gdamore:
The more I think about this, the more I think the problem is a disconnect between developer/maintainer mode, and end-user mode.

The documentation is provided as .txt formatted pages, and they are indeed committed to the git history (just checked).

Now the reality is that it should never be necessary for an end-user to run autogen.sh.  This is necessary when checking out source from github, but really most folks should be building from the released tarballs rather than out of git.  (Except that these days, building from git is probably more common.)

This is just another case where automake/autoconf fall down badly.  I need to revive my work towards a cmake based build for all platforms.

bnewbold:
I've updated this patch to only add directions to `doc/README` (it doesn't touch the top-level README).

gdamore:
fixed in aff10b4d11f8aa8edcef4cb1d87a13eb8ed2e18c

PR #471, opened 2015-10-06
 Allow the last iov size to be set to NN_MSG.

evadot:
This allow dynamic sized payload after some fixed-sized protocol data.

### DIFF
--- a/src/core/global.c
+++ b/src/core/global.c
@@ -894,6 +894,15 @@ int nn_recvmsg (int s, struct nn_msghdr *msghdr, int flags)
         for (i = 0; i != msghdr->msg_iovlen; ++i) {
             iov = &msghdr->msg_iov [i];
             if (nn_slow (iov->iov_len == NN_MSG)) {
+                if (i == msghdr->msg_iovlen - 1) {
+                    void *buf;
+                    buf = nn_allocmsg(sz, 0);
+                    if (nn_slow (buf != NULL)) {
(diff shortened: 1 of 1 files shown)


gdamore:
I'm less confident of this change.  I'm still going through the code trying to make sure we don't leak the data, and that we don't crash on a cleanup.

I'm also of the opinion that this change isn't really adding much value.  Since the new block is dynamically allocated, you could just "peek" into the payload by advancing past your fixed size headers.  That simply adding a value to a pointer.

Am I missing something here?  I'd like to understand the use case a little better.  What is the purpose of separating the headers?

(Oh, and if the headers are small, then one shouldn't worry about the cost of copying them.  That's ~nothing.  Larger payloads do matter though.)

gdamore:
Its actually worse than that.  NN_MSG is supposed to be zero copy, and for single iov case it *is*.  But for this case, we wind up having to do an allocation, and a data copy, in order to keep everything straight.  I think the benefits of this proposal fall far short of outweighing the risks, and if you were planning to do this you'd actually get *better* performance and more efficient code by just using a single element iov array.  You'll be far better off that way.

I'll wait a little while before closing this PR to give you a chance to respond, but at the moment I'd say its very unlikely that I'll integrate this change.

evadot:
I'm closing this PR.
As you said peeking in the payload is easy.
It was a rookie mistake for me to wanted to add this functionnality :)

PR #477, opened 2015-10-21
fixes #476 TCP shutdown problem

### LINKED ISSUE #476: TCP shutdown problem
The errno assertion in nanomsg is unfortunately overly optimistic.  It presumes that sendmsg only returns one of a defined set of documented errnos.

Sadly, this isn't the case always.  I noticed on MacOS X that darwin would occasionally return EPROTOTYPE to this... in contradiction to the man page.  It took a little while to figure this out, but this information is probably helpful:

http://erickt.github.io/blog/2014/11/19/adventures-in-debugging-a-potential-osx-kernel-bug/

As for our case, we'll just relax the assertion, and treat all these cases as connection failure errors.

gdamore:


gdamore:
Integrated in 2163280c83be0ff6977726e23bb2171113b58f44

PR #492, opened 2015-10-23
Wsfail

gdamore:


src/transports/ws/sws.c
```diff
@@ -70,9 +71,9 @@
 #define NN_WS_OPCODE_UNUSED5 0x05
 #define NN_WS_OPCODE_UNUSED6 0x06
 #define NN_WS_OPCODE_UNUSED7 0x07
-#define NN_WS_OPCODE_CLOSE NN_WS_MSG_TYPE_CLOSE
-#define NN_WS_OPCODE_PING NN_WS_MSG_TYPE_PING
-#define NN_WS_OPCODE_PONG NN_WS_MSG_TYPE_PONG
+#define NN_WS_OPCODE_CLOSE 0x08
+#define NN_WS_OPCODE_PING 0x09
+#define NN_WS_OPCODE_PONG 0x0A
```
reqshark: what are opcode's `PING` and `PONG` for? Are they used for the handshake?

wirebirdlabs: Part of the WebSocket spec as per RFC 6455: https://tools.ietf.org/html/rfc6455#section-5.5.2

wirebirdlabs: Expanding; even though they are part of the official RFC 6455 spec, and even though they currently exist within libnanomsg, they should probably be encapsulated within the transport layer, rather than exposed via the application layer.

Basically, their existence, like the Close Handshake, and also message types of either Binary or UTF-8, really throw a wrench into things for this library :-\

gdamore:
As you can see, it took some effort to refine this.  I've collapsed the code (rebased) into a single commit, and I'm now happy with it. 

(Sadly, I've discovered that the tcp_shutdown case causes problems not just for itself, but for any subsequent tests using TCP on AppVeyor.  I've disabled that test on AppVeyor for now.  It still runs on Travis CI though.)

src/transports/ws/sws.c
gdamore: Right... these are no longer expressed to the application at all.

src/transports/ws/sws.c
```diff
@@ -773,14 +744,21 @@ static int nn_sws_fail_conn (struct nn_sws *self, int code, char *reason)
...
     self->fail_msg_len += payload_len;
 
-    self->instate = NN_SWS_INSTATE_FAILING;
+    self->instate = NN_SWS_INSTATE_CLOSED;
 
-    /*  On the next recv, the connection will be failed. Why defer
-        until the next recv? Semantically, until then, this incoming
-        message has not been interpreted, so it's not until then that
-        it could be failed. This type of pre-processing is necessary
-        to early fail chunked transfers. */
```
wirebirdlabs: @gdamore -- I notice your change here "more sanely" immediately fails the peer rather than queueing the action (which would have been fulfilled in line 494 above), but this code was originally developed in order to satisfy behavior of chunked transfers.

This behavior was evidenced when using the full Autobahn Test Suite.

libnanomsg of course never uses chunked transfers, but I was defensively coding against whether a browser (or user agent) might executively decide to chunk transfer from a javascript call that otherwise appeared like a single bulk transfer. At the time, there was no clear evidence whether browsers or user agents were granted the prerogative to make such a transformation based on app-level javascript.

(In the same vein, that is why PING/PONG existed here -- it was not evident whether browsers would have the authority to PING their peer, without having been requested to do so from client-side javascript.)

I don't have strong opinions right now, or even hard evidence to support one way or the other whether "chunked" could be chosen for us outside the will of either libnanomsg or a corresponding javascript implementation, but just wanted to share some design decisions.

src/transports/ws/sws.h
```diff
@@ -82,6 +83,12 @@ struct nn_sws {
     struct nn_fsm fsm;
     int state;
 
+    /*  Endpoint base. */
+    struct nn_epbase *epbase;
+
+    /*  Default message type set on outobund frames. */
```
wirebirdlabs: typo -- "outbound"

src/transports/ws/ws.c
```diff
@@ -148,81 +158,3 @@ static int nn_ws_optset_getopt (struct nn_optset *self, int option,
...
-    /*  Return only the message type (opcode). */
-    if (*msg_type == (NN_WS_MSG_TYPE_GONE | NN_SWS_FRAME_BITMASK_FIN))
-        *msg_type = NN_WS_MSG_TYPE_GONE;
-    else
-        *msg_type &= NN_SWS_FRAME_BITMASK_OPCODE;
-
-    nn_freemsg (cmsg_buf);
-
-    return rc;
-}
```
wirebirdlabs: I agree removing these websocket-specific exported helper methods. It's much better to set message type as socket-level option :-)

tests/ws.c
```diff
@@ -29,15 +30,58 @@
...
+    test_send (sc, "GOOD");
+    test_recv (sb, "GOOD");
+
+    /*  and the bad ... */
+    strcpy((char *)bad, "BAD.");
+    bad[2] = (char)0xDD;
+    test_send (sc, (char *)bad);
+
+    /*  Make sure we dropped the frame.  Apparently it comes back as EAGAIN
+        rather than ETIMEDOUT.  I think this possibly a doc bug. */
```
wirebirdlabs: If the UTF-8 is invalid, should this be EINVAL or even EPROTO or similar? EAGAIN and ETIMEDOUT don't make sense at all.

src/transports/ws/sws.c
gdamore: @wirebirdlabs I don't think this is quite right.  It looked like it was meant to fail chunked transfers *in the face of a deferred* read notification to the user.  That is, we issue a recv response to the user, but we we don't wait for that to occur before issuing the CLOSE to the peer.  (Put another way, it looks like you were attempting to do the fast fail that autobahn wanted, while still waiting until you got all the chunks to issue the response to the app.)

Since we never wait to issue a response, the behavior I've coded here should represent a "fast-failure" like Autobahn wants to see -- that is we close the websocket as soon as we see a failure, without waiting for the full frame to be received.

Note that it appears that many browsers actually don't do the fast fail correctly here, particularly for mangled UTF-8, at least according tot the test results I've seen.

That said, yes, chunked transfers could reasonably happen.  Yes, fast failure is desirable.  No I don't think it is absolutely necessary.  But as I said, I don't think I've broken this aspect of your original behavior.

I hope that all makes sense.

tests/ws.c
gdamore: No.  The issue here is that transport layer has discarded the frame (and closed the connection).  We don't report these details back to the application, ever.

Think of this as a transport level failure, rather than an application failure.

The other thing is that other messages could reasonably arrive, and we'd never notice any error in a real world application, except that our mangled frame got dropped.  Someone would need to look at their javascript logs to see the Close message reason, I think.  That's not unique to us either -- I think all websocket implementations are probably similar here.

src/transports/ws/sws.h
gdamore: Will fix.

src/transports/ws/ws.c
gdamore: Yep, and now I have verification that the socket option actually works. :-)

tests/ws.c
gdamore: The right place to report this, if ever we could, would be at the *sender side*.  Arguably in *that* case we could actually capture this and report it back to the sender.  Even that might not be very satisfactory in the face of e.g. device frameworks and proxies, that are just relaying data.

This is all part of the design for nanomsg -- it always just works (except when it doesnt'), and never reports errors back to the application.  (Unfortunately it *does* crash/core on assertion failures far more often than I'd like, but that's a separate rant.)

src/transports/ws/sws.c
wirebirdlabs: What you're saying does not sound wrong ...

... but what I can say is that at the time, when developing this part that feels/looks wonky compared to your edit, it was done for a reason.

Now, I cannot say if that reason was a good reason, or even if it was then, is now *still* a good reason -- but it was definitely to address some clever corner-case that the Autobahn fuzzer had concocted.

Or maybe, it was to address a deficiency in the sync/async send/recv capabilities of libnanomsg vs requirements here ... I can't quite recall ... this was almost a year ago, and I'll do everything I can to mentally spool up again to match your renewed vigor here. :+1:

(For context about the talk about the "Autobahn" tester, and also apparent "lack" of websocket testing in libnanomsg -- there exists a PR I created that provides a comprehensive RFC 6455 fuzzer and stress tester, except we debated whether this PR would be necessary for libnanomsg or not. The arguments against where 1) it introduces new dependencies, such as Python and Python packages, and 2) libnanomsg does not attempt to be, and specifically should not attempt to be, fully RFC 6455 compliant. I agree with these two arguments.)

tests/ws.c
wirebirdlabs: > would be at the sender side

This is sender-side; line 69 above.

That said ... I now see that sws.c does not validate that an outbound message "claimed" to be UTF-8 is indeed UTF-8 -- it only validates incoming (untrusted) messages. This seems reasonable.

Meaning, of course, it could not return EINVAL or EPROTO without validating the claim of the application. That said ... it's still not evident why EAGAIN or ETIMEDOUT would be the expected error code ...

src/transports/ws/sws.c
wirebirdlabs: A more direct way of talking about this modification -- did you change it in order to simplify what seemed "wonky" or "wrong", or was it specifically to address a characterized failure? Don't take this question as accusatory -- rather, it's highlighting my own immediate lack of ability to give positive, helpful review of this change, without yet having mentally spooled up again!

src/transports/ws/sws.c
wirebirdlabs: At the time this was developed, it was not explicitly stated whether a browser or user agent could unsolicitedly (from the perspective of app code) could PING. For this reason, I chose to defensively go ahead and aim for full RFC 6455 compliance, such that your-browser-du-jour would not tend to crash `libnanomsg`. Now, in 2015, it's still not any more clear. With hindsight, I would tend to criticize the existence of PING/PONG in RFC 6455, since it just kinda throws a monkey wrench into things.

gdamore: Agreed with that ping and pong are kind of silly.  I think the idea was to do a midlayer keepalive replacement. 

Much about websockets feels wonky.  The whole UTF-8 validation thing is stupid in my opinion. There is no good reason for a transport layer to be doing payload validation. 

That said I think you did the right thing in implementing it.

src/transports/ws/sws.c
gdamore: It came about as I started looking at cleaning up changes to remove those functions and add the type option.  When I started testing I found that I got back a zero length frame which was NOT what should have happened.  And that led me here. 

So yes I encountered specific behavior that I believed was incorrect.  I realize you did it that way on purpose but I think you've since come to agree with me that websockets should not be a special snowflake but instead behave just like any other nanomsg transport.

tests/ws.c
gdamore: If its not evident, that’s a failure in my ability to express what is
happening.

The recv is timing out due to no input.  Because we drop the frame (and the
connection) when the invalid data is first received at the bottom transport
layer, without ever sending anything up to the application.   (No zero
length frame, nor any error.  The application is entirely shielded from
transport errors — this is part of the design principles of libnanomsg.)

Without a RCVTIMEO timer, the recv doesn’t timeout at all, but just blocks
forever.  I put the timeout in so that I could detect that the frame was
dropped without delivery.

Now it turns out that the code is buggy here, and returns EAGAIN instead of
ETIMEDOUT.  That’s a bug in the core socket handling code, and I plan to
fix it (its a trivial change) to make it conform to the documented
behavior.  EAGAIN is only for reporting an empty recv queue when using
non-blocking I/O.

Does that clear it up at all?

On Sat, Oct 24, 2015 at 7:57 PM, Wirebird Labs LLC <notifications@github.com

wirebirdlabs: > The recv is timing out due to no input

~~the recv on which line? I just see a `send` on line 69.~~

I now found it within the new `test_drop`. This is a nit/opinion, but perhaps that function could be renamed `test_recv_fail_with_error`? or similar? such function/macro could be useful in making many tests more concise.

(my misunderstanding is due to code reviews on github, where it's easier to overlook detail, rather than being in an IDE with richer context)

src/transports/ws/sws.c
wirebirdlabs: > got back a zero length frame which was NOT what should have happened

Aha. I think I hear where you're coming from now, and probably agree with what you did here.

As for the zero-length message -- the *payload* is zero length, but the message carried information in the CMSG header about the disconnected peer. Which was the application's way of detecting the peer's "close handshake", yet i think we've come to the agreement that the spirit of nanomsg should not reveal this to the application layer.

tests/ws.c
gdamore: Right.  The intention is that new test “test_drop()” tests that a message
was dropped.  I do intend to use this in future tests. :-)
 recv_fail_with_error suggests that it gets back an error, but in this case
its really “recv_got_nothing”. :-)  Which is mostly why I like
“test_drop()”.  To me at least, this is a concise description of what its
testing for.

On Sun, Oct 25, 2015 at 9:01 AM, Wirebird Labs LLC <notifications@github.com

wirebirdlabs: I like the semantic of `drop` in the context of `ETIMEDOUT`. Meaning, once the EAGAIN is replaced with ETIMEDOUT in this case, the expected error would not need to be passed as a parameter -- e.g., `test_drop (sb);`. A function along the lines of `test_recv_fail_with_error()` would also pass the expected error type

PR #494, opened 2015-10-25
fixes #493 statistics race with FSM

### LINKED ISSUE #493: statistics race with FSM
According to PR #359 there is a race caused by starting the FSM too soon.  Deferring the start solves the problem, apparently.

gdamore:


gdamore:
This was fixed a while ago.

PR #502, opened 2015-10-27
Assertion or Access Violation when calling nn_close while nn_recv blocks in separate thread

wirebirdlabs:
This pull request demonstrates and partially addresses another type of multithreading problem during TCP Shutdown on Windows.

It is split into two commits on purpose:

1) the first commit demonstrates the current issue in the form of a test -- the assertion failure in `efd.c`
2) the second commit fixes this issue by giving Windows platforms the same semantics as *NIX platforms, as shown above in `efd.c`

**However** -- this patch turns a guaranteed failure (old line 70 of `efd.c`) into a race condition within `sock.c`.

To best demonstrate this race condition, increase `TEST3_LOOPS ` in line 41 of `tcp_shutdown.c` test to be 10 or 100 iterations.

The current value of 1 iteration makes a lower likelihood of the race condition showing up, but it certainly still can manifest.

### DIFF
--- a/src/utils/efd.c
+++ b/src/utils/efd.c
@@ -67,7 +67,20 @@ int nn_efd_wait (struct nn_efd *self, int timeout)
         tv.tv_usec = timeout % 1000 * 1000;
     }
     rc = select (0, &self->fds, NULL, NULL, timeout >= 0 ? &tv : NULL);
-    wsa_assert (rc != SOCKET_ERROR);
+
+    if (nn_slow (rc == SOCKET_ERROR)) {
+        rc = nn_err_wsa_to_posix (WSAGetLastError ());
+        errno = rc;
+        
+        /*  Treat these as a non-fatal errors, typically occuring when the
+            socket is being closed from a separate thread during a blocking
+            I/O operation. */
+        if (nn_fast (rc == EINTR || rc == ENOTSOCK))
+            return -EINTR;
+    }
+
+    wsa_assert (rc >= 0);
+
     if (nn_slow (rc == 0))
         return -ETIMEDOUT;
     return 0;
--- a/src/utils/err.c
+++ b/src/utils/err.c
@@ -160,6 +160,8 @@ int nn_err_wsa_to_posix (int wsaerr)
         return ECONNABORTED;
     case WSAECONNRESET:
         return ECONNRESET;
+    case WSAENOTSOCK:
+        return ENOTSOCK;
     case ERROR_BROKEN_PIPE:
         return ECONNRESET;
     case WSAESOCKTNOSUPPORT:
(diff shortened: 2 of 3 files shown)


gdamore:
PR seems to cause hangs in the test.  See CI details.

wirebirdlabs:
> hangs

Indeed -- I'm guessing it's the new test with `tcp_shutdown` -- are you able to see any obvious problems with that test? I don't have an environment readily available to test, but could fire up something on OS X later to troubleshoot ...

wirebirdlabs:
Here's a shot in the dark (not tested) -- is it due to https://github.com/wirebirdlabs/featherweight-nanomsg/blob/5ed04c5464ca5b3c808b8ba5589070f29a4c799b/tests/tcp_shutdown.c#L107-L109 ??

gdamore:
Yep.  The nn_recv is blocking forever, I think.

I think nn_close() doesn’t alert the blocked threads that the file is no
longer open.  Feels like a nasty bug.   And probably difficult to fix.

I’m pretty sure EINTR is also the wrong errno in this case — we should see
ETERM instead.

On Tue, Oct 27, 2015 at 9:36 AM, Wirebird Labs LLC <notifications@github.com

gdamore:
Could be -- if you are expecting to receive that signal to wake up the thread.  However, most operating systems actually disable receipt of the SIGPIPE signal elsewhere, and return EPIPE inline.  (So for example, this shouldn't be an issue on Linux.  Unless disabling SIGPIPE somehow also disables the kernel from returning EPIPE or EBADFD on a closed pipe?

wirebirdlabs:
I tried a quick build on OS X and could reproduce the hang; commenting out referenced lines above (suppression of SIGPIPE) did not help.

Should we open an issue and discuss further there, or continue here?

Briefly, I feel that this is a test that should pass, and even though the proposed fix in the PR seems to work on Windows, I'm not sure yet exactly the fix for *NIX platforms.

gdamore:
I'm reasonably comfortable with the changes to the windows code in efd.c.  I'm far less comfortable with tcp_shutdown.c changes.  That test is incredibly fragile, for a variety of reasons, I think.  I'm really not keen to further complicate it.  Frankly, I'd rather create a new test file, and add tests to that, than to add to tcp_shutdown.c  -- a big part of my reason behind that is that it is already close to impossible to diagnose the failures in tcp_shutdown.

Lets also separate this into separate PRs please.  One PR for the windows fix, and another entirely for enhancing the tests.

wirebirdlabs:
I will refactor the test into a new test.

That said, the goal of one PR was to demonstrate 1) the problem in a test, then 2) the fix. If you prefer two separate PRs, that's cool, but just wanted to share rationale why it was originally submitted as an atomic PR.

wirebirdlabs:
Closing; this PR is supplanted by #520 based on discussion in this thread

Issue #505, opened 2015-11-04
ws.c test periodically hangs


wirebirdlabs:
This issue is to continue a conversation first started from a pull request: https://github.com/nanomsg/nanomsg/pull/503#issuecomment-153775151

> @gdamore writes: "I think the windows code is particularly fragile... I have some theories here, but it seems like some tests trash the stack. The tcp_shutdown test was particularly hard here; once it failed it left the stack in a state that even other tests would fail subsequently."

Can you elaborate on this?

reqshark:
On the node binding we started seeing TCP timeouts in our tests back in September, which I attribute then to slow-joiner syndrome inherit in PUBSUB topology, especially where millisecond-level deliveries hinge on connection ordering. That's a good recipe for hanging indefinitely, since publishers could send off a message before the TCP CONNECTION is established by the subscriber. 

Moreover, when using `nn_connect()`, especially over TCP, it's my understanding that the return value (whether or not it blocks) represents the endpoint, *not the fact that the handshake between server and client has been completed*

Just think about it, PUB SUB was designed so that a publisher could begin sending messages long before any subscribers come along to recv, so missed messages with that pattern are just part of it.

I've also noticed a lot of round-about dily-dallying in the `nn_sock_recv()` loop, [the `usock.c` file](https://github.com/nanomsg/nanomsg/blob/master/src/core/sock.c#L650-L694). Recalculation of deliberate timeouts and other potential back-offs might be overkill for the purpose of a recv operation. gentlemen, correct me if I'm wrong there please.

gdamore:
You might be on to something @reqshark.

In mangos I had some of these problems, so I used a repeated try protocol to determine when the connection was established (with an eventual timeout) before actually starting the real test.

Probably the test code should be rewritten to be a little more robust -- ensuring that the connection is actually established before running the real test.  (And waiting up to ~5 seconds for the connection to establish, with repeated broadcast attempts on the socket at frequent -- say 50 msec or so -- intervals.)

wirebirdlabs:
> slow-joiner syndrome inherit in PUBSUB topology

It seems you're describing pretty well the expected behavior, which is also the observed behavior.

Just to clarify this issue -- the problem here is not missed messages -- those are expected. The problem is the **hang** caused sometimes.

Said another way -- it should be expected that the following psuedocode would fail:

```
//assume REMOTE_ADDRESS is bound and publishing
nn_socket (NN_SUB);
[...]

Issue #371, opened 2015-02-14
Assertion failed: !ipv6 (src/transports/utils/iface.c:88)


rzezeski:
The interface lookup code assumes that each name-family pair has one unique interface, but that is not always the case. E.g. on my macbook the `lo0` interface listens on two ipv6 addresses: host local `::1`, and link local `fe80::1`. These are represented as two separate `struct ifaddrs` from `getifaddrs(3)`.

Here is a minimal reproducer:

```C
#include <assert.h>
#include <nanomsg/nn.h>
#include <nanomsg/reqrep.h>

void
main()
{
	int eid, optval = 0, rv, sock;

	sock = nn_socket(AF_SP, NN_REP);
	assert(sock != -1);
	rv = nn_setsockopt(sock, NN_SOL_SOCKET, NN_IPV4ONLY,
	    &optval, sizeof (optval));
	assert(rv != -1);
	eid = nn_bind(sock, "tcp://lo0:7777");
	assert(eid != -1);
}
```

```
$ ./multi-ipv6-fail
Assertion failed: !ipv6 (src/transports/utils/iface.c:88)
Abort trap: 6
```

ricki:
Same issue here. In my case, `eth0` has three `AF_INET6` entries. A link local `fec0::/10`, a Teredo tunneling `2001::/32`, and the default unicast address, which should be preferred over the others acording to RFC 6724.

gdamore:
So other systems (illumos) create a unique logical interface for each IP address.  Apparently mac does something different, allowing multiple IPs to be present.  I even see unicast 127.94.0.1 and 127.94.0.2 on my lo0 port on OSX 10.10.1!

PR #507, opened 2015-11-05
fixes #506 ipaddr regression affecting iOS/Windows

### LINKED ISSUE #506: interface/address issue prevents bind/connect
confirmed connect and bind failures on iOS platform, also reported tonight on Windows

msgs over TCP/WS have failed silently/inconsistently after updating to beta 7, but we've pinpointed the issue thanks to @jl777 reporting with @gdamore discussion in [nanomsg gitter chatroom](https://gitter.im/nanomsg/nanomsg).

The problem for me on iOS resolves immediately by reverting
```git
commit a896c3051f1b8a0da4fd92819cca9f102dd610e2
Author: Garrett D'Amore garrett@damore.org
Date: Sun May 31 14:17:08 2015 -0700
fixes #371 Assertion failed: !ipv6 (src/transports/utils/iface.c:88)
```

 accordingly here's #507 submitted under the project's MIT license

reqshark:
Signed-off-by: Bent Cardan <bent@nothingsatisfies.com>

### DIFF
--- a/src/transports/utils/iface.c
+++ b/src/transports/utils/iface.c
@@ -41,27 +41,6 @@ static void nn_iface_any (int ipv4only, struct sockaddr_storage *result,
 
 #include <ifaddrs.h>
 
-static int nn_is_v4_unicast (const struct sockaddr *sa)
-{
-    const struct sockaddr_in *sin = (const void *)sa;
-    if (sa->sa_family != AF_INET)
-        return 0;
-    if ((ntohl(sin->sin_addr.s_addr) & 0xF0000000U) == 0xE0000000U) {
-        return 1;
-    }
-    return 0;
-}
-
-static int nn_is_v6_unicast (const struct sockaddr *sa)
-{
-    const struct sockaddr_in6 *sin6 = (const void *)sa;
-    if (sa->sa_family != AF_INET6)
-        return 0;
-    if (sin6->sin6_addr.s6_addr[0] == 0xff)
-        return 0;
-    return 1;
-}
-
 int nn_iface_resolve (const char *addr, size_t addrlen, int ipv4only,
     struct sockaddr_storage *result, size_t *resultlen)
 {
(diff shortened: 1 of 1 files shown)


reqshark:
i watched appveyor pass, updated the commit and got a WS timeout. 

unfortunately this doesn't fix #505

gdamore:
Backing out is probably not the right choice here -- it may be something we do, but ultimately we still need to fix the original bug -- #371

wirebirdlabs:
> Backing out is probably not the right choice here -- it may be something we do, but ultimately we still need to fix the original bug

Agree.

It's possible (likely?) a nuanced bug has arisen based on how https://github.com/nanomsg/nanomsg/commit/a896c3051f1b8a0da4fd92819cca9f102dd610e2 sorts unicast vs multicast addresses.

Meaning, even though that step may have introduced what feels like a regression, it's definitely headed in the right direction, and should not be reverted.

reqshark:
So I'll need to take a closer look at what's actually happening here and report back those findings

gdamore:
So the problem here is that you're attempting to bind to an interface name, which is problematic at best.  Better to bind to an explicit IP address than an interface name.

I need more information about the specific code that is timing out or failing; I'd also like to have some samples that illustrate the interfaces that are plumbed on your system. (Such as the output from ifconfig -a)

Again, if you specify a numeric IP address instead of an interface name, you bypass all this code, and probably save all the problems.

In fact, I feel that using interface names so fundamentally flawed (specific to operating systems, cannot ever work on Windows, and just generally fragile both from an *implementation* choice *and* from an application design view) that I'd like to just rip this functionality out altogether.

To be clear: use of named outbound interfaces (e.g. "lo0") in nanomsg URIs is IMO fundamentally broken, and cannot be solved in any portable fashion.  I'm therefore *strongly* in favor of removing it.

reqshark:
well I'm using IP address numbers, not named interfaces, so apprently I'm not bypassing this code.

I'm taking a look at the call stack right now and I'll update with some more information shortly

gdamore:
Please show me an example of the URL you are binding to.

gdamore:
And ideally also, output from ifconfig -a or whatever analog you have.

gdamore:
Reading the code, the only way you get there is if nn_literal_resolve() fails.  That only fails if you give it an IP address string that it can't parse.

It can parse strings of the form "[::1]" and "127.0.0.1".  Using IPv6 addresses without enclosing brackets likely fails however.

gdamore:
Please also move this to #506.  I'm going to actually close this PR, because no matter what other action we may take on this, accepting this PR isn't one of them.

PR #508, opened 2015-11-05
fixes #505 hang in websocket test

### LINKED ISSUE #505: ws.c test periodically hangs

wirebirdlabs:


gdamore:
Well, that's weird. This change broke Linux somehow.  So either we've regressed, or it doesn't address the problem completely.  Interesting that one of the Windows checks failed too, but that seems unrelated to this failure.

reqshark:
update the macros so this change only applies to windows

gdamore:
@reqshark No.  Not unless there is no other way.  That way leads to spaghetti and a more difficult to support code base.

reqshark:
well, come on let's face it, the systems are different, they've been designed to behave differently, whether intentionally or not is another issue.

> That way leads to spaghetti and a more difficult to support code base.

clearly articulated code comments go a long way to combat that type of thing, but of course you're right.

gdamore:
I've confirmed that the problem is that attempting to get the transport specific option set fails due to a block on the global lock.  We're recursively trying to lock.  More soon.

wirebirdlabs:
> fails due to a block on the global lock

Interesting. I wonder if you won't come back with a resolution along the lines of dynamically allocating each socket and using a pointer/handle to such allocations as the socket identifier rather than [0, 1023] on a global array ;-)

Regardless the resolution, would you consider accepting this PR as-is, then building upon it, since this PR represents another type of correctness? (specifically, caching the socket type once on connect)

Staying tuned.

gdamore:
No, this PR cannot be accepted as is.  It definitely is a 100% hang condition.  I am working on a more robust fix.  Stay tuned.

wirebirdlabs:
> No, this PR cannot be accepted as is

In case it wasn't clear, i meant build your "hang" fix upon this "correctness" fix in the `pr-for-505` branch -- not necessarily yet merging this branch verbatim. Since, this branch does not fix the hang, but it does fix another correctness bug.

gdamore:
Oh, I'm working starting from your work, whether git reflects that or not in the end.

gdamore:
I'm thinking it would be painful to fix this in a way that allows endpoint transport options to be retrieved or accessed (or changed) during data transport, but setting it up to work during connect/bind is easy.

This is fallout from the design of the existing locks.  I think this is a reasonable limitation.

wirebirdlabs:
no worries; cherry-picking from `pr-for-505` will also work, if you're pretty confident in your fix. the benefit of working directly on the `pr-for-505` is that if we needed to iterate, i could easily pull your changes again (wihtout having to re-cherry-pick), but this particular issue is straightforward enough i don't anticipate further back-and-forth. the joys of git, right?

gdamore:
Probably the way this will work is I will fix the lock recursion independently, and apply that change *under* yours.  So yours will be cherry picked on top of my fix.

wirebirdlabs:
> cherry picked on top of my fix

That's cool. As long as it's a bona fide cherry pick or a rebase or similar -- and not a copy and paste -- it'll be easy for me to then get back to mainline on my fork. :-)

gdamore:
Merged along with #515 as part of PR #516.

commit d4238b8f97b83963523cecf18cd5a031bf8f06d3
Date:   Thu Nov 5 10:58:01 2015 -0600

    fixes #505 hang in websocket test

PR #513, opened 2015-11-10
fixes #506 interface/address issue prevents bind/connect

### LINKED ISSUE #506: interface/address issue prevents bind/connect

gdamore:
This eliminates the support for named interfaces in nanomsg URLs.  It only ever worked on a subset of platforms, and that was inconsistent.  We require a literal IP address to be supplied now.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,5 +1,6 @@
 # Copyright (c) 2013 Luca Barbato
 # Copyright (c) 2013 GoPivotal, Inc.  All rights reserved.
+# Copyright 2015 Garrett D'Amore <garrett@damore.org>
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"),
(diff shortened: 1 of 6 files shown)


gdamore:
Note that use of "names" like "localhost" was never supported either.  Frankly *that* would probably have been more useful.  Still, IP addresses are more reliable.  In the future someone could add support for resolving local DNS names here.

PR #520, opened 2015-11-11
fixes #502 Assertion or Access Violation when calling nn_close while nn_recv blocks in separate thread

wirebirdlabs:
This PR replaces #502 by splitting out the test from `tcp_shutdown` into a new test called `async_shutdown`

Again, for clarity, this PR fixes one type of bug on Windows, but then exposes a new failure mode downstream. The test -- but not the fix -- might expose further issues on *NIX systems.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -83,6 +83,7 @@ add_libnanomsg_test (survey)
 add_libnanomsg_test (bus)
 
 #  Feature tests.
+add_libnanomsg_test (async_shutdown)
 add_libnanomsg_test (block)
 add_libnanomsg_test (term)
 add_libnanomsg_test (timeo)
--- a/Makefile.am
+++ b/Makefile.am
@@ -485,6 +485,7 @@ PROTOCOL_TESTS = \
     tests/bus
 
 FEATURE_TESTS = \
+    tests/async_shutdown \
     tests/block \
     tests/term \
     tests/timeo \
--- a/src/utils/efd.c
+++ b/src/utils/efd.c
@@ -67,7 +67,20 @@ int nn_efd_wait (struct nn_efd *self, int timeout)
         tv.tv_usec = timeout % 1000 * 1000;
     }
     rc = select (0, &self->fds, NULL, NULL, timeout >= 0 ? &tv : NULL);
-    wsa_assert (rc != SOCKET_ERROR);
+
+    if (nn_slow (rc == SOCKET_ERROR)) {
+        rc = nn_err_wsa_to_posix (WSAGetLastError ());
+        errno = rc;
+        
+        /*  Treat these as a non-fatal errors, typically occuring when the
+            socket is being closed from a separate thread during a blocking
+            I/O operation. */
+        if (nn_fast (rc == EINTR || rc == ENOTSOCK))
+            return -EINTR;
+    }
+
+    wsa_assert (rc >= 0);
+
     if (nn_slow (rc == 0))
         return -ETIMEDOUT;
     return 0;
--- a/src/utils/err.c
+++ b/src/utils/err.c
@@ -160,6 +160,8 @@ int nn_err_wsa_to_posix (int wsaerr)
         return ECONNABORTED;
     case WSAECONNRESET:
         return ECONNRESET;
+    case WSAENOTSOCK:
+        return ENOTSOCK;
     case ERROR_BROKEN_PIPE:
         return ECONNRESET;
     case WSAESOCKTNOSUPPORT:
--- a/tests/async_shutdown.c
+++ b/tests/async_shutdown.c
@@ -1,0 +1,76 @@
+/*
+    Copyright (c) 2012 Martin Sustrik  All rights reserved.
+    Copyright (c) 2015 Jack R. Dunaway.  All rights reserved.
+
+    Permission is hereby granted, free of charge, to any person obtaining a copy
+    of this software and associated documentation files (the "Software"),
+    to deal in the Software without restriction, including without limitation
+    the rights to use, copy, modify, merge, publish, distribute, sublicense,
+    and/or sell copies of the Software, and to permit persons to whom
+    the Software is furnished to do so, subject to the following conditions:
+
+    The above copyright notice and this permission notice shall be included
+    in all copies or substantial portions of the Software.
+
+    THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+    IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+    THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+    LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+    FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+    IN THE SOFTWARE.
+*/
+
+#include "../src/nn.h"
+#include "../src/pair.h"
+#include "../src/pubsub.h"
+#include "../src/pipeline.h"
+#include "../src/tcp.h"
+
+#include "testutil.h"
+#include "../src/utils/attr.h"
+#include "../src/utils/thread.c"
+#include "../src/utils/atomic.c"
+
+/*  Test condition of closing sockets that are blocking in another thread. */
+
+#define TEST_LOOPS 10
+#define SOCKET_ADDRESS "tcp://127.0.0.1:5557"
+
+struct nn_atomic active;
+
+static void routine (NN_UNUSED void *arg)
+{
+    int s;
+    int rc;
+    int msg;
+
+    nn_assert (arg);
+
+    s = *((int *) arg);
+
+    /*  We don't expect to actually receive a message here;
+        therefore, the datatype of 'msg' is irrelevant. */
+    rc = nn_recv (s, &msg, sizeof(msg), 0);
+
+    errno_assert (nn_errno () == EINTR);
+}
+
+int main ()
+{
+    int sb;
+    int i;
+    struct nn_thread thread;
+
+    for (i = 0; i != TEST_LOOPS; ++i) {
+        sb = test_socket (AF_SP, NN_PULL);
+        test_bind (sb, SOCKET_ADDRESS);
+        nn_sleep (100);
+        nn_thread_init (&thread, routine, &sb);
+        nn_sleep (100);
+        test_close (sb);
+        nn_thread_term (&thread);
+    }
+
+    return 0;
+}


gdamore:
Good news.  I think I've figured out this particular hang.  Basically, the nn_recv() keeps looping, even though it gets interrupted, because it doesn't see the STOPPING_EPS state.  This is an *easy* thing to fix properly.  Stay tuned.

wirebirdlabs:
> Good news. I think I've figured out this particular hang. Basically, the nn_recv() keeps looping, even though it gets interrupted, because it doesn't see the STOPPING_EPS state. This is an easy thing to fix properly

Most excellent.

gdamore:
Thanks for your test code.  It made narrowing this down quickly much much easier.

gdamore:
Agreed.  I'm going to make some minor changes to these, along with my fix for the upper layer problem, and probably commit once I verify I have fixes that work.

gdamore:
Hmm... now I'm not sure.  I'm getting different results on Darwin.  Stay tuned.

wirebirdlabs:
> Hmm... now I'm not sure

This summarizes my >1yr narrative on this issue.

Currently, it seems the most robust resolution is to do away with the statically-allocated socket table (https://github.com/nanomsg/nanomsg/blob/master/src/core/global.c#L234-L235) that requires the global mutex, instead allocating each socket ad-hoc as it is created.

The effective socket identifier, from the perspective of the library user -- rather than being an integer on the range of [0, NN_MAX_SOCKETS-1] -- would be a pointer to a pointer.

Then, sockets individually can mutex themselves when appropriate, without denying service to the rest of the sockets in the loaded instance.

The ultimate problem with the send and recv loops in `sock.c` is that they release a mutex that could be destroyed by the time they try to regain the mutex: https://github.com/nanomsg/nanomsg/blob/master/src/core/sock.c#L601-L608

wirebirdlabs:
> Currently, it seems the most robust resolution

Another potential solution is to truly fix `NN-LINGER`, which currently is not even implemented.

One current vector that's working charmingly well is to zombify the individual socket (https://github.com/wirebirdlabs/featherweight-nanomsg/commit/cb1408979e4951e0dd0b5fe4ae4779df9b0641e1), then allowing events rolling in during this zombified period to be handled as if the socket were still active (https://github.com/wirebirdlabs/featherweight-nanomsg/commit/79e41f6f9909849b33733873be0a4b102022d4d2).

It seems `NN_LINGER` could be defined as the period in which the socket is "closed" until the time it is "deallocated" -- which decently maps to what "zombified" means today.

gdamore:
At this point I'm not in favor of changing from an indexed table to a list of pointers.  The zombification is similar to the approach I'm looking at.  It was working for me on illumos systems, but I've seen problems with MacOS.  The hang on MacOS occurs elsewhere.

I'm starting to suspect that the different polling mechanisms (which are all used to cope with the async I/O nature and the need to support poll/select) are part of the problem.  I think the fact that closing() a FD doesn't always issue a notifciation/wakeup from the efd_wait() is the problem.

This would be *lots* easier if we could just use normal blocking I/O and threads.  Even elimination of select() compatible so that we could use condition variables instead would be a *huge* benefit.  But break probably gazillions of consumers who've come to rely on the select() / poll() compatibility feature. (So we can't do it. :-(   )

wirebirdlabs:
> break probably gazillions of consumers who've come to rely on the select() / poll() compatibility feature

Have you verified elsewhere this feature remains a requirement? There weren't any that spoke up on: http://www.freelists.org/post/nanomsg/Use-nanomsg-sockets-with-select-on-Windows

wirebirdlabs:
> I think the fact that closing() a FD doesn't always issue a notifciation/wakeup from the efd_wait() is the problem

This rings a bell, when I was attempting to do some refactoring (that did not pan out) early in the year (last 5 commits on this abandoned branch: https://github.com/wirebirdlabs/featherweight-nanomsg/commits/abandoned-sock-refactor)

gdamore:
So this is interesting... on darwin, close() of a fd does not cause POLLIN wakeups in poll().  Which is why it hangs differently on MacOS.

wirebirdlabs:
> So this is interesting... on darwin, close() of a fd does not cause POLLIN wakeups in poll(). Which is why it hangs differently on MacOS

You're probably way ahead, but here's a trick to overcome this, by creating a private pipe always in index 0 of the poll set: http://stackoverflow.com/questions/19133138/cancel-a-socket-poll-operation

gdamore:
That's a helpful link.  I think I can just write to the polled fd since that is already what this is supposed to be doing.  Stay tuned.

wirebirdlabs:
> just write to the polled fd since that is already what this is supposed to be doing

The trick here would be ensuring you could detect the value written as a sentinel value, and not something valid from a peer. Seems risky/brittle.

gdamore:
To be honest, I'm tempted to just enforce that we have a timeout, say 1 sec, and wake up on close.  Its crap, but it would work.  I can't believe operating systems think its OK to block in poll with invalid FDs.  This is friggin' insane.

The hard part here is not sending the message, its actually figuring out when you can actually close the file descriptor.  Because you can't close it until all reads are done, and you don't know when *that* is complete.

wirebirdlabs:
> The hard part here is not sending the message, its actually figuring out when you can actually close the file descriptor. Because you can't close it until all reads are done, and you don't know when that is complete.

This plays into why it feels `NN_LINGER` should be fully-fleshed out as a viable resolution for this issue.

Right now, `nn_close()` is a synchronous operation, and this call is where we're having all the cross-platform hanging problems. There also exists a comment in that function saying that's it's current implementation is suboptimal: https://github.com/nanomsg/nanomsg/blob/master/src/core/global.c#L576-L577

Instead, I feel `nn_close()` should effectively be *asynchronous*. It would synchronously set the socket as being zombified, but would not yet necessarily `close()` the underlying FD's or `free()` associated resources.

Once the `NN_LINGER` period has elapsed, cleanup occurs, including throwing away all outstanding read/write buffers, then `free()`ing and then returning the nanomsg socket descriptor [0, NN_MAX_SOCKETS-1] to the global pool.

That said, at the app layer, the developer can assume that `nn_close()`, even though no longer a synchronous operation, the library makes the guarantee that the resources will be properly cleaned up once the linger period has elapsed. (If not, this is a library bug that needs to be fixed)

Thoughts?

gdamore:
Fixing this "properly" is going to take some more time.  I have a clear idea, and yeah, I'm gonna have to create a new file descriptor to add to the poll group; this is also going to lead ultimately to fixing the LINGER problem, since once I have that I can add the linger support properly.  This will also require adding a bit more state to the socket, since I'm going to need to have reference counts on threads that are in nn_sendmsg() or nn_recvmsg() so that I can determine when they are done.  Probably also a new NN_CLOSING state or somesuch.  (I don't think I can use NN_STOPPING.)

It's not something I can do right this very second, as I think its several hours of work to get this completely correct and have any assurance of it actually being so.  Sadly, its also going to require testing on several different platforms (Mac, Solaris, Linux, and Windows at least) to have any confidence that I've got it right everywhere.

This whole efd nonsense is the result of trying to fake real UNIX file descriptors.  I cannot say this strongly enough -- this was a substantial architectural mistake.  Especially since we're *not* going to emulate the *broken* behavior on broken systems.  (E.g. perhaps on Darwin the fact that close() doesn't terminate poll() or recv() ought to be behavior we emulate *on* Darwin.  Frankly that's what happens today -- the system calls are busted and we behave accordingly.)

wirebirdlabs:
Legit.

> Probably also a new NN_CLOSING state

NN_LINGERING might be a good name.

wirebirdlabs:
> This whole efd nonsense is the result of trying to fake real UNIX file descriptors. I cannot say this strongly enough -- this was a substantial architectural mistake

To be fair, let's classify it as innovation that didn't necessarily pan out. :-)

And that said, I've not seen a solid answer yet as to why we can't get rid of http://www.freelists.org/post/nanomsg/Use-nanomsg-sockets-with-select-on-Windows to eliminate complexity. Why not?

gdamore:
Solving it for windows solves *part* of the problem while leaving the rest.  More importantly,  at this point I have no idea what language ports / runtimes, other consumers, etc. might be reliant on the whole select() thing.

gdamore:
Well this is interesting.

On MacOS: closing a local end of a pipe doesn't interrupt a poll() on that descriptor from a different thread.  If you close it synchronously from the same thread, the poll returns immediately.

But, if you close the *far* end of a pipe from a pipe, it does abort the poll.  So maybe that's a simpler fix still.

gdamore:
aha.  The next problem is that file descriptors get reused, so your poll may wind up polling against some other file descriptor.  Argh.

wirebirdlabs:
> The next problem is that file descriptors get reused, so your poll may wind up polling against some other file descriptor. Argh

Yep. This is also problematic at higher layers, such as framework and application -- where additional bookkeeping and mutexing is required to solve this issue -- which is what led my train of thought toward *"The effective socket identifier, from the perspective of the library user -- rather than being an integer on the range of [0, NN_MAX_SOCKETS-1] -- would be a pointer to a pointer"* as being able to solve multiple layers of problems. Foremost, the incorrectness tracked by this Github issue, and also usability/correctness of library usage at higher layers.

And if not a raw pointer, at least an ID that's reasonably unique per session of the library, such as an autoincrementing integer.

gdamore:
Yeah, autoincrementing integers would be better, probably.  That probably
breaks efficient lookup.

Believe me, I’d prefer to pass around *pointers*, except that with a socket
/ file descriptor semantic people expect that they can use these things
even after being closed, which basically makes it hard to guard against
use-after-free.

Again, the whole “socket” / “file descriptor” equivalence semantic instead
of using an opaque handle (pointer) is just an unfortunate design choice.
I don’t see how to fix that without breaking semantics and library API. :-(

All this calls for a new / alternate library with much more sensible
semantics and API.  Which is more than I am able to do *right now*.   For
now I just want to achieve the point of minimal brokenness in this library.

On Sun, Nov 15, 2015 at 9:59 AM, Wirebird Labs LLC <notifications@github.com

gdamore:
Please have a look at PR #522   -- I think this addresses these problems.  Its a bit more convoluted to get it all work, since i had to fix issues in multiple places (eventfd and core sock state machine), but so far it seems ok.

gdamore:
So the global socket checks are not properly protected.  There are NN_BASIC_CHECKS but those checks are not done carefully under the global lock.  There are some further issues here as well that I'm working on.

wirebirdlabs:
> autoincrementing integers ... probably breaks efficient lookup

Perhaps, but not noticeable; the most naïve approach here makes socket creation time go from O(1) to O(N) worst cast on a negligible lookup that is many orders of magnitude from the actual bottleneck during socket creation. After that, a single modulo of the exposed FD still provides O(1) lookup for all other socket operations.

Given a bit of time, I'd like to look into this, whether those statements actually hold and are easily implemented ...

wirebirdlabs:
> NN_BASIC_CHECKS but those checks are not done carefully under the global lock

To expand this; NN_BASIC_CHECKS *and the subsequent actions that expect the preconditions guaranteed by NN_BASIC_CHECKS* are not done under the global lock.

gdamore:
I'm closing this, as I believe PR #525 addresses all of these problems.  Please read, test, and review that PR.  I'm hoping to push to master soon, but I'd like other folks to to try it out; its a rather large diff, so I'm trying to be a little more cautious.

gdamore:
I don't think we need to worry about this. I'm pretty sure the work I've done earlier today resolves these issues.

gdamore:
This should be *closed closed.

PR #522, opened 2015-11-15
Fix502

gdamore:
This commit is just to get some tests started on Linux and Windows via CI. 

This basically includes all of the logic I've created to harden shutdown properly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -83,6 +83,7 @@ add_libnanomsg_test (survey)
 add_libnanomsg_test (bus)
 
 #  Feature tests.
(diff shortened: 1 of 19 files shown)


gdamore:
Huh.  This seems to pass on Windows, Darwin, and Solaris, but fails on Linux.  Too many different event handling notification mechanisms!

gdamore:
Ok, I reproduce this async-shutdown mutex error pretty consistently on Linux.  Argh.

gdamore:
Closing this, because #525  is the "real" consolidated PR with all these changes.

PR #525, opened 2015-11-16
fixes #65 nn_close should abort blocking calls on socket

### LINKED ISSUE #65: nn_close should abort blocking calls on socket
If there is no timeout defined on a socket, and the socket is closed while another thread is waiting in nn_recv(), it would be very helpful for me to get an error back directly. Similar to how nn_term() works.

This reproduces the problem for me:
https://gist.github.com/op/5641440

### LINKED ISSUE #519: How to close a device without closing everything.
What you have suggested have tried already but it didnt work.
Please find the code below what i have used.

class NNHTest : public ::testing::Test
{
public:
nn_helper_t proxy_st ;

};

struct info_my_param_t
{
int sock1;
int sock2;
};

static void * loop_device_test( void* args )
{
int rc;
struct info_my_param_t * imp = (struct info_my_param_t *) args ;
rc = nn_device ( imp->sock1, imp->sock2 );
printf("value of nn_device rc = %d\n", rc);

return NULL;

}

TEST_F(NNHTest, StopDeviceTest )
{
	int rc ,nnbd, con;
//device

    int sock1 = nn_socket ( AF_SP_RAW , NN_REP );
	printf("value of sock1 = %d\n", sock1);

	nnbd = nn_bind ( sock1 , "tcp://127.0.0.1:6699" );
	printf("value of nnbd = %d\n", nnbd);

	int sock2 = nn_socket ( AF_SP_RAW , NN_REQ );
	printf("value of sock2 = %d\n", sock2);

	con = nn_connect ( sock2 , "tcp://127.0.0.1:6799" );
	printf("value of con = %d\n", con);

//device loop start

	struct info_my_param_t imp ;
	imp.sock1 = sock1 ;
	imp.sock2 = sock2 ;
	struct info_my_param_t *ptr = (struct info_my_param_t *)memdup( (char*)&imp, sizeof(struct info_my_param_t) )  ;

	rc = pthread_create( &proxy_st.threadid, NULL, loop_device_test, ptr  );
	printf("value of pthread_create rc = %d\n", rc);
	sleep(1);

//closing sock1
	rc = nn_close(imp.sock1);
	printf("value of nn_close(imp.sock1) rc = %d\n", rc);

//closing sock2
	rc = nn_close(imp.sock2);
	printf("value of nn_close(imp.sock2) rc = %d\n", rc);

//joining main thread
	pthread_join( proxy_st.threadid, NULL);

}

OUTPUT:

keshav@keshav-Latitude-E6540:~/clone/BD/Test/Shared/Core$ ./ubuntuRelease64/test-static
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from NNHTest
[ RUN ] NNHTest.DeviceTestReqResp
value of sock1 = 0
value of nnbd = 1
value of sock2 = 1
value of con = 1
value of pthread_create rc = 0
value of nn_close(imp.sock1) rc = 0
value of nn_close(imp.sock2) rc = 0

Note : its hanging here

### LINKED ISSUE #523: Want a test to verify nn_close() aborts nn_recv()
We've seen lots of problems around this, and as I have new code coming, we'd really to have a nice test for this.  Thankfully @wirebirdlabs has provided one in PR #520 that only needs minor tweaks.

### LINKED ISSUE #524: Sockets array is not thread-safe
The sockets array is used in a very MT-unsafe manner, which is probably the root of many many issues.  In particular, nn_close() can cause pretty much random and undefined behavior for any threads that may be attempting to perform I/O against the same sockets. 

I believe the design goal here was to have one thread doing nn_send/nn_recv, and that same thread would be responsible for ultimately closing the socket with nn_close().

Of course, nn_device() flies in the face of that design.  And its darned useful to be able to close a socket to shut down things from another thread.

What is necessary here is substantial surgery in global.c and sock.c.  We need to keep a reference count on sockets, and only clean up & destroy things when we know that no other consumers are present.  And we need to use the states to prevent sockets from getting references on sockets that are in the middle of shutting down.

gdamore:
fixes #524 Sockets array is not thread-safe
fixes #519 How to close a device without closing everything.
fixes #523 Want a test to verify nn_close() aborts nn_recv()

Portions contributed by Jack R. Dunaway <jack@wirebirdlabs.com>.

### DIFF
--- a/src/core/global.c
+++ b/src/core/global.c
@@ -101,19 +101,6 @@
     the type should be changed to uint32_t or int. */
 CT_ASSERT (NN_MAX_SOCKETS <= 0x10000);
 
-/*  This check is performed at the beginning of each socket operation to make
-    sure that the library was initialised, the socket actually exists, and is
-    a valid socket index. */
-#define NN_BASIC_CHECKS \
-    if (nn_slow (s < 0 || s > NN_MAX_SOCKETS)) {\
-        errno = EBADF;\
-        return -1;\
-        }\
-    if (nn_slow (!self.socks || !self.socks [s])) {\
-        errno = EBADF;\
-        return -1;\
-    }
-
 #define NN_CTX_FLAG_ZOMBIE 1
 
 #define NN_GLOBAL_SRC_STAT_TIMER 1
@@ -139,10 +126,12 @@ struct nn_global {
     /*  Combination of the flags listed above. */
     int flags;
 
-    /*  List of all available transports. */
+    /*  List of all available transports.  Note that this list is not
+        dynamic; i.e. it is created during global initialization and
+        is never modified. */
     struct nn_list transports;
 
-    /*  List of all available socket types. */
+    /*  List of all available socket types.  Again this list is not dynamic.*/
     struct nn_list socktypes;
 
     /*  Pool of worker threads. */
@@ -178,7 +167,7 @@ static void nn_global_add_socktype (struct nn_socktype *socktype);
 
 /*  Private function that unifies nn_bind and nn_connect functionality.
     It returns the ID of the newly created endpoint. */
-static int nn_global_create_ep (int s, const char *addr, int bind);
+static int nn_global_create_ep (struct nn_sock *, const char *addr, int bind);
 
 /*  Private socket creator which doesn't initialize global state and
     does no locking by itself */
@@ -190,6 +179,10 @@ static void nn_global_handler (struct nn_fsm *self,
 static void nn_global_shutdown (struct nn_fsm *self,
     int src, int type, void *srcptr);
 
+/*  Socket holds. */
+static int nn_global_hold_socket(struct nn_sock **sockp, int s);
+static int nn_global_hold_socket_locked(struct nn_sock **sockp, int s);
+static void nn_global_rele_socket(struct nn_sock *);
 
 int nn_errno (void)
 {
@@ -304,7 +297,7 @@ static void nn_global_init (void)
         self.statistics_socket = nn_global_create_socket (AF_SP, NN_PUB);
         errno_assert (self.statistics_socket >= 0);
 
-        rc = nn_global_create_ep (self.statistics_socket, addr, 0);
+        rc = nn_global_create_ep (self.socks[self.statistics_socket], addr, 0);
         errno_assert (rc >= 0);
     } else {
         self.statistics_socket = -1;
@@ -555,7 +548,7 @@ int nn_socket (int domain, int protocol)
 
     rc = nn_global_create_socket (domain, protocol);
 
-    if(rc < 0) {
+    if (rc < 0) {
         nn_global_term ();
         nn_glock_unlock ();
         errno = -rc;
@@ -570,27 +563,46 @@ int nn_socket (int domain, int protocol)
 int nn_close (int s)
 {
     int rc;
+    struct nn_sock *sock;
 
-    NN_BASIC_CHECKS;
-
-    /*  TODO: nn_sock_term can take a long time to accomplish. It should not be
-        performed under global critical section. */
     nn_glock_lock ();
-
-    /*  Deallocate the socket object. */
-    rc = nn_sock_term (self.socks [s]);
-    if (nn_slow (rc == -EINTR)) {
+    rc = nn_global_hold_socket_locked (&sock, s);
+    if (nn_slow (rc < 0)) {
         nn_glock_unlock ();
+        errno = -rc;
+        return -1;
+    }
+
+    /*  Start the shutdown process on the socket.  This will cause
+        all other socket users, as well as endpoints, to begin cleaning up.
+        This is done with the glock held to ensure that two instances
+        of nn_close can't access the same socket. */
+
+    nn_sock_stop (sock);
+    nn_glock_unlock ();
+
+    /*  We have to drop both the hold we just acquired, as well as
+        the original hold, in order for nn_sock_term to complete. */
+    nn_sock_rele (sock);
+    nn_sock_rele (sock);
+
+    /*  Now clean up.  The termination routine below will block until
+        all other consumers of the socket have dropped their holds, and
+        all endpoints have cleanly exited. */
+    rc = nn_sock_term (sock);
+    if (nn_slow (rc == -EINTR)) {
+        nn_global_rele_socket (sock);
         errno = EINTR;
         return -1;
     }
 
     /*  Remove the socket from the socket table, add it to unused socket
         table. */
-    nn_free (self.socks [s]);
+    nn_glock_lock ();
     self.socks [s] = NULL;
     self.unused [NN_MAX_SOCKETS - self.nsocks] = s;
     --self.nsocks;
+    nn_free (sock);
 
     /*  Destroy the global context if there's no socket remaining. */
     nn_global_term ();
@@ -604,97 +616,126 @@ int nn_setsockopt (int s, int level, int option, const void *optval,
     size_t optvallen)
 {
     int rc;
+    struct nn_sock *sock;
 
-    NN_BASIC_CHECKS;
-
-    if (nn_slow (!optval && optvallen)) {
-        errno = EFAULT;
-        return -1;
-    }
-
-    nn_glock_lock ();
-    rc = nn_sock_setopt (self.socks [s], level, option, optval, optvallen);
-    nn_glock_unlock ();
+    rc = nn_global_hold_socket (&sock, s);
     if (nn_slow (rc < 0)) {
         errno = -rc;
         return -1;
     }
-    errnum_assert (rc == 0, -rc);
 
+    if (nn_slow (!optval && optvallen)) {
+        rc = -EFAULT;
+        goto fail;
+    }
+
+    rc = nn_sock_setopt (sock, level, option, optval, optvallen);
+    if (nn_slow (rc < 0))
+        goto fail;
+    errnum_assert (rc == 0, -rc);
+    nn_global_rele_socket (sock);
     return 0;
+
+fail:
+    nn_global_rele_socket (sock);
+    errno = -rc;
+    return -1;
 }
 
 int nn_getsockopt (int s, int level, int option, void *optval,
     size_t *optvallen)
 {
     int rc;
+    struct nn_sock *sock;
 
-    NN_BASIC_CHECKS;
-
-    if (nn_slow (!optval && optvallen)) {
-        errno = EFAULT;
-        return -1;
-    }
-
-    nn_glock_lock ();
-    rc = nn_sock_getopt (self.socks [s], level, option, optval, optvallen);
-    nn_glock_unlock ();
+    rc = nn_global_hold_socket (&sock, s);
     if (nn_slow (rc < 0)) {
         errno = -rc;
         return -1;
     }
-    errnum_assert (rc == 0, -rc);
 
+    if (nn_slow (!optval && optvallen)) {
+        rc = -EFAULT;
+        goto fail;
+    }
+
+    rc = nn_sock_getopt (sock, level, option, optval, optvallen);
+    if (nn_slow (rc < 0))
+        goto fail;
+    errnum_assert (rc == 0, -rc);
+    nn_global_rele_socket (sock);
     return 0;
+
+fail:
+    nn_global_rele_socket (sock);
+    errno = -rc;
+    return -1;
 }
 
 int nn_bind (int s, const char *addr)
 {
     int rc;
+    struct nn_sock *sock;
 
-    NN_BASIC_CHECKS;
-
-    nn_glock_lock();
-    rc = nn_global_create_ep (s, addr, 1);
-    nn_glock_unlock();
+    rc = nn_global_hold_socket (&sock, s);
     if (rc < 0) {
         errno = -rc;
         return -1;
     }
 
+    rc = nn_global_create_ep (sock, addr, 1);
+    if (nn_slow (rc < 0)) {
+        nn_global_rele_socket (sock);
+        errno = -rc;
+        return -1;
+    }
+
+    nn_global_rele_socket (sock);
     return rc;
 }
 
 int nn_connect (int s, const char *addr)
 {
     int rc;
+    struct nn_sock *sock;
 
-    NN_BASIC_CHECKS;
-
-    nn_glock_lock();
-    rc = nn_global_create_ep (s, addr, 0);
-    nn_glock_unlock();
-    if (rc < 0) {
+    rc = nn_global_hold_socket (&sock, s);
+    if (nn_slow (rc < 0)) {
         errno = -rc;
         return -1;
     }
 
+    rc = nn_global_create_ep (sock, addr, 0);
+    if (rc < 0) {
+        nn_global_rele_socket (sock);
+        errno = -rc;
+        return -1;
+    }
+
+    nn_global_rele_socket (sock);
     return rc;
 }
 
 int nn_shutdown (int s, int how)
 {
     int rc;
+    struct nn_sock *sock;
 
-    NN_BASIC_CHECKS;
-
-    rc = nn_sock_rm_ep (self.socks [s], how);
+    rc = nn_global_hold_socket (&sock, s);
     if (nn_slow (rc < 0)) {
+        errno = -rc;
+        return -1;
+    }
+
+    rc = nn_sock_rm_ep (sock, how);
+    if (nn_slow (rc < 0)) {
+        nn_global_rele_socket (sock);
         errno = -rc;
         return -1;
     }
     nn_assert (rc == 0);
 
+    nn_global_rele_socket (sock);
     return 0;
 }
 
@@ -741,24 +782,29 @@ int nn_sendmsg (int s, const struct nn_msghdr *msghdr, int flags)
     void *chunk;
     int nnmsg;
     struct nn_cmsghdr *cmsg;
+    struct nn_sock *sock;
 
-    NN_BASIC_CHECKS;
-
-    if (nn_slow (!msghdr)) {
-        errno = EINVAL;
+    rc = nn_global_hold_socket (&sock, s);
+    if (nn_slow (rc < 0)) {
+        errno = -rc;
         return -1;
     }
 
+    if (nn_slow (!msghdr)) {
+        rc = -EINVAL;
+        goto fail;
+    }
+
     if (nn_slow (msghdr->msg_iovlen < 0)) {
-        errno = EMSGSIZE;
-        return -1;
+        rc = -EMSGSIZE;
+        goto fail;
     }
 
     if (msghdr->msg_iovlen == 1 && msghdr->msg_iov [0].iov_len == NN_MSG) {
         chunk = *(void**) msghdr->msg_iov [0].iov_base;
         if (nn_slow (chunk == NULL)) {
-            errno = EFAULT;
-            return -1;
+            rc = -EFAULT;
+            goto fail;
         }
         sz = nn_chunk_size (chunk);
         nn_msg_init_chunk (&msg, chunk);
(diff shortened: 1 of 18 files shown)


gdamore:
@wirebirdlabs I've gone ahead and merged all these changes into a single changeset.  The rationale for doing so is that the entire set of changes are required for a complete fix; including just a portion of them leads to brokenness.  I've tried to give you credit in a sensible manner, although git only allows me to record one author.  (Given the rather large amount of refactoring I did in sock.c, core.c, and the efd framework, it didn't really feel right leaving you as the primary commit author.)

gdamore:
Btw, this fixes a slew of errors/failures that I've seen on illumos systems.  Testing is clean on Solaris/illumos, Darwin (MacOS X), Linux, and Windows.  Finally.

wirebirdlabs:
> although git only allows me to record one author. (Given the rather large amount of refactoring I did in sock.c, core.c, and the efd framework, it didn't really feel right leaving you as the primary commit author.)

This is acceptable; forward progress on this issue is infinitely more valuable than pristine commit history.

I'll try to scatter a few comments in the commit in the next few minutes while you have inertia ... then a more thorough testing next few days ...

src/core/global.c
```diff
@@ -101,19 +101,6 @@
...
-#define NN_BASIC_CHECKS \
-    if (nn_slow (s < 0 || s > NN_MAX_SOCKETS)) {\
-        errno = EBADF;\
-        return -1;\
-        }\
-    if (nn_slow (!self.socks || !self.socks [s])) {\
-        errno = EBADF;\
-        return -1;\
-    }
-
```
wirebirdlabs: In a nutshell, if this went away, where did it go? Briefly scanning this changeset it seems straightforward (to me), but can you say in your own words the design decision here?

src/core/global.c
```diff
@@ -178,7 +167,7 @@ static void nn_global_add_socktype (struct nn_socktype *socktype);
 
 /*  Private function that unifies nn_bind and nn_connect functionality.
     It returns the ID of the newly created endpoint. */
-static int nn_global_create_ep (int s, const char *addr, int bind);
+static int nn_global_create_ep (struct nn_sock *, const char *addr, int bind);
```
wirebirdlabs: This feels like a significant change; briefly, describe design decision?

src/core/global.c
```diff
@@ -190,6 +179,10 @@ static void nn_global_handler (struct nn_fsm *self,
 static void nn_global_shutdown (struct nn_fsm *self,
     int src, int type, void *srcptr);
 
+/*  Socket holds. */
+static int nn_global_hold_socket(struct nn_sock **sockp, int s);
+static int nn_global_hold_socket_locked(struct nn_sock **sockp, int s);
+static void nn_global_rele_socket(struct nn_sock *);
```
wirebirdlabs: The terminology "hold" -- it's not necessarily bad, not necessarily wrong, but a new concept at least to this codebase. Let's consider if the word "hold" best fits, localizes well, etc ...

src/core/global.c
```diff
@@ -570,27 +563,46 @@ int nn_socket (int domain, int protocol)
...
 {
     int rc;
+    struct nn_sock *sock;
 
-    NN_BASIC_CHECKS;
-
-    /*  TODO: nn_sock_term can take a long time to accomplish. It should not be
-        performed under global critical section. */
     nn_glock_lock ();
+    rc = nn_global_hold_socket_locked (&sock, s);
```
wirebirdlabs: Describing to someone who would have zero previous knowledge; why is it necessary to "hold" a socket on line 569 when a "global lock" was acquired on line 568?

src/core/global.c
```diff
@@ -570,27 +563,46 @@ int nn_socket (int domain, int protocol)
...
+        This is done with the glock held to ensure that two instances
+        of nn_close can't access the same socket. */
+
+    nn_sock_stop (sock);
+    nn_glock_unlock ();
+
+    /*  We have to drop both the hold we just acquired, as well as
+        the original hold, in order for nn_sock_term to complete. */
+    nn_sock_rele (sock);
+    nn_sock_rele (sock);
```
wirebirdlabs: Lines 586 and 587 are duplicated?

Also, what does "rele" mean? Can this be expanded to its full meaning?

src/core/global.c
gdamore: This logic was replaced with inline calls to nn_global_hold_socket(), which includes these checks, but also puts a *hold* on the socket structure, preventing it from getting destroyed out from underneath us.  Basically instead of these spot checks, I'm using a much more robust reference counting system.

src/core/global.c
```diff
@@ -604,97 +616,126 @@ int nn_setsockopt (int s, int level, int option, const void *optval,
...
 {
     int rc;
+    struct nn_sock *sock;
 
-    NN_BASIC_CHECKS;
-
-    if (nn_slow (!optval && optvallen)) {
-        errno = EFAULT;
-        return -1;
-    }
```
wirebirdlabs: This conditinal could actually remain outside the lock, since it's a gross sanity check, yes?

src/core/global.c
gdamore: I've used this logic in *numerous* other code bases.  "Hold" means to put a reference count on it, and therefore prevent another routine from destroying the object.  hold/rele are paired.

src/core/global.c
gdamore: Because the hold persists -- that is it extends beyond the time when when lock is released.

This paradigm is common in many other code bases:

grab lock.
find object.
bump refcnt (hold the object)
drop the lock

... use the object ...

grab lock
drop reference count
possibly clean up object (we don't really do this quite the same here)
drop lock

src/core/sock.c
```diff
@@ -48,6 +49,7 @@
 #define NN_SOCK_STATE_ZOMBIE 3
 #define NN_SOCK_STATE_STOPPING_EPS 4
 #define NN_SOCK_STATE_STOPPING 5
+#define NN_SOCK_STATE_FINI 6
```
wirebirdlabs: `FINI` -- can this be expanded? For clarity.

src/core/global.c
gdamore: read the comment above.  586 and 587 are *not* duplicated.  There are two reference counts to drop.

(Hmm.. I should move the comment at 584 to above 581 for clarity.)

rele is shorthand for "release"

"hold"/"release"

src/core/sock.c
```diff
@@ -188,31 +194,54 @@ void nn_sock_zombify (struct nn_sock *self)
...
+        till they are all closed.  This loop is not interruptible, because
+        making it so would leave a partially cleaned up socket, and we don't
+        have a way to defer resource deallocation. */
+    for (;;) {
+        rc = nn_sem_wait (&self->termsem);
+        if (nn_slow (rc == -EINTR))
+            continue;
+        errnum_assert (rc == 0, -rc);
+        break;
+    }
```
wirebirdlabs: Without having tested; what are your thoughts on how long this could block?

src/core/sock.c
gdamore: FINI - "finish".  I could have use "STOPPED" here equally, although STOPPED has some other uses which may be confusing in this file.

src/core/sock.c
```diff
@@ -188,31 +194,54 @@ void nn_sock_zombify (struct nn_sock *self)
...
+
+    /*  Also, wait for all holds on the socket to be released.  */
+    for (;;) {
+        rc = nn_sem_wait (&self->relesem);
+        if (nn_slow (rc == -EINTR))
+            continue;
+        errnum_assert (rc == 0, -rc);
+        break;
+    }
```
wirebirdlabs: Same question; how long could this take? Is it denying service to other sockets in the meantime? (i.e., by holding the global mutex? or is the global mutex not held here?)

src/core/global.c
gdamore: Yes, but if you look carefully, you'll see that there is no "lock" held, only the "hold" on the socket structure.  So while this sanity check could be moved, the question is really one of consistency and style.  TBH I don't have a strong opinion about it, and am happy to move it if you think it helps readability.

src/core/sock.c
```diff
@@ -869,7 +934,15 @@ static void nn_sock_shutdown (struct nn_fsm *self, int src, int type,
         /*  Protocol-specific part of the socket is stopped.
             We can safely deallocate it. */
         sock->sockbase->vfptr->destroy (sock->sockbase);
-        sock->state = NN_SOCK_STATE_INIT;
```
wirebirdlabs: Feels "right" replacing this line!

src/core/global.c
gdamore: Actually, the "hold" terminology I picked up from kernel programming, where it is incredibly common.  I don't know how else to describe what I'm doing.

src/core/sock.c
```diff
@@ -1006,7 +1079,7 @@ void nn_sock_report_error (struct nn_sock *self, struct nn_ep *ep, int errnum)
     if (errnum == 0)
         return;
 
-    if(ep) {
+    if (ep) {
```
wirebirdlabs: +1 for attention to detail on the style guide for nanomsg

src/core/global.c
gdamore: We pass in the socket structure because we've already looked it up.  There is no need to have the fragility of a second lookup here.  This actually slightly simplifies the code, and makes it more robust and clear that I'm using a structure rather than some index into the array.

More to the point, the only place I use the array index (well, apart from the special statistics socket initialization bit) is when I also *validate* the array index and place a hold on the structure.  Everywhere else I use the structure I know that I can safely do so because I've already gotten a shared reference (hold) on the structure.

src/core/sock.c
```diff
@@ -1088,3 +1161,24 @@ void nn_sock_stat_increment (struct nn_sock *self, int name, int64_t increment)
...
+int nn_sock_hold (struct nn_sock *self)
+{
+    switch (self->state) {
+    case NN_SOCK_STATE_ACTIVE:
+    case NN_SOCK_STATE_INIT:
+        self->holds++;
+        return 0;
+    case NN_SOCK_STATE_ZOMBIE:
+        return -ETERM;
+    }
```
wirebirdlabs: There's not much the compiler can do for us here, going on pure C convention with no typedefs, but I find myself wondering ... "what other cases are there??"

Said another way ... it's hard to read this code and feel certain it's not wrong. It cannot explicitly tell me it's not wrong. Therefore, it's hard to reason whether it's correct.

gdamore: The other cases are NN_STATE_STOPPING, STOPPING_EP, and FINI.

All EBADF. :-)

wirebirdlabs: Perhaps, explicitly enumerate all expected cases, then nn_unreachable assert on default case?

src/utils/efd_eventfd.inc
```diff
@@ -61,18 +69,25 @@ void nn_efd_signal (struct nn_efd *self)
 {
     const uint64_t one = 1;
     ssize_t nbytes;
+    int fd = self->efd;
```
wirebirdlabs: Is this for readability, or functionality? Either way, it looks good, but is a slight departure from the rest of the codebase.

src/utils/efd_win.inc
```diff
@@ -201,14 +202,35 @@ wsafail3:
     return -rc;
 }
 
+void nn_efd_stop (struct nn_efd *self)
+{
+    int rc;
+    SOCKET s = self->w;
+    self->w = INVALID_SOCKET;
```
wirebirdlabs: It is OK to not do an atomic swap here? (i.e., is it properly MT mutexed?)

src/core/sock.c
gdamore: Not very long.  Less than a second normally.  Probably 10-20 msec.

tests/async_shutdown.c
```diff
@@ -0,0 +1,76 @@
...
+int main ()
+{
+    int sb;
+    int i;
+    struct nn_thread thread;
+
+    for (i = 0; i != TEST_LOOPS; ++i) {
+        sb = test_socket (AF_SP, NN_PULL);
+        test_bind (sb, SOCKET_ADDRESS);
+        nn_sleep (100);
```
wirebirdlabs: Using `nn_sleep` here is a known race condition that I wrote; it should be refactored with proper semaphore eventually.

src/core/sock.c
gdamore: The global mutex is NOT held.  This is a big difference from before where it was held here. This shouldn't affect anything other than the thread calling close.

src/utils/efd_eventfd.inc
gdamore: Readability, performance (fewer dereferences), and consistency.

src/core/sock.c
gdamore: I can do that.  Not sure it really helps readability. But maybe.

src/utils/efd_win.inc
gdamore: Good question.  We should be in the outer context. So should be safe.

tests/async_shutdown.c
gdamore: I'm not sure its worth fixing in this test case.  You need to wait for TCP connections to be established. We lack port events to get this right and this lame delay is adequate.

src/core/global.c
wirebirdlabs: Got it. I understand these terms now.

src/core/global.c
wirebirdlabs: Assuming such gross failures on `getsockopt()` and `setsockopt()` are both 1) infrequent and 2) acceptable to take a few instructions longer to fail than failing early, consistency is more important; I'd vote to keep as you put it.

src/core/sock.c
wirebirdlabs: Sweet.

src/core/sock.c
wirebirdlabs: > Not sure it really helps readability

100% agree, but it sorta combats unreadabilty. Where, combatting unreadability with `nn_assert` on a `default` case feels ... not awesome.

src/utils/efd_win.inc
wirebirdlabs: Atomic swap would be a helpful wrapper utility. We have atomic increment/decrement utility functions: https://github.com/nanomsg/nanomsg/blob/master/src/utils/atomic.c

tests/async_shutdown.c
wirebirdlabs: This is another topic altogether :-\ For now we'll let it be.

src/core/sock.h
```diff
@@ -78,6 +79,9 @@ struct nn_sock
     /*  Next endpoint ID to assign to a new endpoint. */
     int eid;
 
+    /*  Count of active holds against the socket. */
+    int holds;
```
wirebirdlabs: How about making this an atomic counter, using atomic increment/decrement? Is it possible that this alone could elimiate other mutexing? (probably not, but I have not thought through the problem; you're in the zone and could probably give a quick answer)

Could we gain anything else from using atomic counter here, further than a pedantic semantic? Conversely, do there exist new potential bugs from not using atomic counter?

gdamore: I don't think atomic adds anything. I've been really careful to use this only under the ctx lock.  If anything adding an atomic may be misleading by drawing attention away from the real protection.

src/utils/efd_win.inc
gdamore: I don't think it needs to be atomic though.

src/core/sock.c
gdamore: So I'm adding the explicit cases along with default.  For default I put it to EBADF as well.  I'm not a huge fan of littering the code with explicit checks / assertions; in fact I think a lot of the fragility we've seen in nanomsg in the past has been the result of too 'stringent' checks on things like errnos.

In this case, if its not a state that I explicitly covered already, then I want to return EBADF.

src/core/global.c
gdamore: Nope the comment is where it belongs, as I come back to this today.

src/core/sock.h
wirebirdlabs: Seems legit. If the "real protection" you speak of could be decomposed into lower-level protections such as this, there's merit in considering it, but 1) that's too big a bite to take right now, and 2) it doesn't jump out as something that would be obviously successful (since it feels like the heavy-handedness of a blanket protection here is justified)

src/core/global.c
wirebirdlabs: Sounds good.

src/core/global.c
wirebirdlabs: Sounds good. Acquire/release feels a little more conventional, but if hold/release is kernel-speak then let's stick with it.

wirebirdlabs: > Acquire/release feels a little more conventional

Wait a minute.... acquire/release is a different semantic than hold/release ... yes? They are not synonyms.

Acquire/release is for mutexing; hold/release is for reference counting.

Don't mind my comments on terminologies so much on this review; I need to just grok a bit better. At this point, it feels the code is correct and I'm needing to catch up.

src/core/sock.c
wirebirdlabs: I like this strategy.

And on the record, I feel that the @sustrik design decision to put `assert()` all over the place was absolutely correct and has made for a better library. That's another topic ... but to tie into what you say, libnanomsg is definitely at the tipping point nearing v1.0 where `assert()` should no longer be sprinkled, but a run-time error code instead.

That said, I'd be in favor of a new error code -- say ENANOMSG or such -- whose existence in the field would be distinct from EBADF.

The semantic of ENANOMSG alerts nanomsg library devs that something is wrong and needs to be fixed, whereas EBADF alerts library users (application developers) that they have made a logic error.

This could tie in with work @TTimo pushed for long ago, turning `nn_assert` into a logger rather than a panic stop.

Thoughts?

src/core/global.c
gdamore: Yep, you've got it.  I was going to comment that acquire is usually exclusive, whereas hold is more like a shared lock; although it isn't a lock at all -- its just something to keep the system from tossing the memory out from underneath us while we're using it. :-)

src/core/sock.h
gdamore: It can't really be decomposed.  Really, this is the best solution -- and its a paradigm that has been well tested and battle proven in operating system kernels for decades.

src/core/sock.c
gdamore: I have different opinions about assert()  -- its useful for documenting conditions that the programmer is certain of; external errnos don't fall into that category.

EBADF is used only *here* because its the most appropriate semantic to use.  We do have EFSM already which I could use for the default case, if it really bothers you.  (My feeling is that any other states that may get added should be treated as not-ready in this case, i.e. closing or not initialized yet, therefore EBADF.)

I don't want to create a new magic errno.  EBADF has particular semantics here -- the file handle (socket descriptor) is invalid.  (Usually because it is closed or somesuch.)

nn_assert() needs to *always* be a panic stop.  Just logging & continuing makes it something much different than an assertion.  The Gnome people went that way years ago, and as a result gnome programs spew assertions like crazy -- because they assert things that clearly are false or not assertions, and yet the program continues onward.  That's abuse of assertions.

assert() is like crashing a kernel.  Its to be used only when you've got some fundamental assumption, and if that assumption is broken then the very worst thing you can do is try to continue, because it means you have no idea what the state of the program is.

If asserts are firing and causing problems, then those are critical bugs that need to be fixed.  In many case we just need to remove the assertions.  (For example, handling of errnos -- instead of an assertion we should inject correct handling to deal with unexpected system errors.)

Anyway, this is way out of scope for this change set.

wirebirdlabs: Agree with all you say, and for clarification, I still delineate the important difference you mention between panic and logging.

The point I was trying to make is this -- on usage of `nn_assert()` currently in libnanomsg in fairly benign failure modes -- let's say a small, infrequent memory leak -- those instances would be changed to a new `nn_log_failure()` or such.

For true sanity checks, where insanity presents a grievous failure mode, `nn_assert()` keeps the exact same behavior as today, and remains in the codebase.

Ideally, this is all moot, with the number of problems resolved in recent weeks!

gdamore:
I rebased and merged this.

PR #551, opened 2015-11-23
Providing test to demonstrate hang in nn_close on websocket transport

wirebirdlabs:
This test demonstrates an occasional hang (at least on Windows) on line 100 when the PUB socket is closed. Creating PR in order to kick off automated builds on other platforms to see if they fail here as well, or if it is limited to just Windows.

This only *appears* to hang on the WebSocket transport. This could be just a problem with the WebSocket transport implementation (the most likely scenario), or it could be because the WebSocket transport incidentally requires more time to connect and perhaps disconnect sockets, as compared to say TCP, and is thus more likely to trigger a race condition.

### DIFF
--- a/tests/ws_async_shutdown.c
+++ b/tests/ws_async_shutdown.c
@@ -63,6 +63,8 @@ static void routine (NN_UNUSED void *arg)
 
 int main ()
 {
+    int i;
+    int j;
     int s;
     int sb;
     int rcvtimeo = 10;
@@ -70,14 +72,14 @@ int main ()
     int sockets [TEST_THREADS];
     struct nn_thread threads [TEST_THREADS];
 
-    for (int i = 0; i != TEST_LOOPS; ++i) {
+    for (i = 0; i != TEST_LOOPS; ++i) {
         
         sb = test_socket (AF_SP, NN_PUB);
         test_bind (sb, SOCKET_ADDRESS);
         test_setsockopt (sb, NN_SOL_SOCKET, NN_SNDTIMEO,
             &sndtimeo, sizeof (sndtimeo));
 
-        for (int j = 0; j < TEST_THREADS; j++){
+        for (j = 0; j < TEST_THREADS; j++){
             s = test_socket (AF_SP, NN_SUB);
             test_setsockopt (s, NN_SOL_SOCKET, NN_RCVTIMEO,
                 &rcvtimeo, sizeof (rcvtimeo));
@@ -92,7 +94,7 @@ int main ()
 
         test_send (sb, "");
         
-        for (int j = 0; j < TEST_THREADS; j++) {
+        for (j = 0; j < TEST_THREADS; j++) {
             test_close (sockets [j]);
             nn_thread_term (&threads [j]);
         }
--- a/tests/ws_async_shutdown.c
+++ b/tests/ws_async_shutdown.c
@@ -1,0 +1,104 @@
+/*
+    Copyright (c) 2012 Martin Sustrik  All rights reserved.
+    Copyright (c) 2015 Jack R. Dunaway. All rights reserved.
+
+    Permission is hereby granted, free of charge, to any person obtaining a copy
+    of this software and associated documentation files (the "Software"),
+    to deal in the Software without restriction, including without limitation
+    the rights to use, copy, modify, merge, publish, distribute, sublicense,
+    and/or sell copies of the Software, and to permit persons to whom
+    the Software is furnished to do so, subject to the following conditions:
+
+    The above copyright notice and this permission notice shall be included
+    in all copies or substantial portions of the Software.
+
+    THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+    IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+    THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+    LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+    FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+    IN THE SOFTWARE.
+*/
+
+#include "../src/nn.h"
+#include "../src/pubsub.h"
+
+#include "testutil.h"
+#include "../src/utils/thread.c"
+
+/*  Test condition of closing sockets that are blocking in another thread. */
+
+#define SOCKET_ADDRESS "ws://127.0.0.1:5558"
+#define TEST_LOOPS 10
+#define TEST_THREADS 10
+
+static void routine (NN_UNUSED void *arg)
+{
+    int s;
+    int rc;
+    char msg[1];
+
+    nn_assert (arg);
+
+    s = *(int *)arg;
+    
+    while (1) {
+        rc = nn_recv (s, &msg, sizeof(msg), 0);
+        if (rc == 0) {
+            continue;
+        }
+
+        nn_assert (rc == -1);
+
+        /*  A timeout is OK since PUB/SUB is lossy. */
+        if (nn_errno () == ETIMEDOUT) {
+            continue;
+        }
+        break;
+    }
+    /*  Socket is expected to be closed by caller.  */
+    errno_assert (nn_errno () == EBADF);
+}
+
+int main ()
+{
+    int s;
+    int sb;
+    int rcvtimeo = 10;
+    int sndtimeo = 0;
+    int sockets [TEST_THREADS];
+    struct nn_thread threads [TEST_THREADS];
+
+    for (int i = 0; i != TEST_LOOPS; ++i) {
+        
+        sb = test_socket (AF_SP, NN_PUB);
+        test_bind (sb, SOCKET_ADDRESS);
+        test_setsockopt (sb, NN_SOL_SOCKET, NN_SNDTIMEO,
+            &sndtimeo, sizeof (sndtimeo));
+
+        for (int j = 0; j < TEST_THREADS; j++){
+            s = test_socket (AF_SP, NN_SUB);
+            test_setsockopt (s, NN_SOL_SOCKET, NN_RCVTIMEO,
+                &rcvtimeo, sizeof (rcvtimeo));
+            test_setsockopt (s, NN_SUB, NN_SUB_SUBSCRIBE, "", 0);
+            test_connect (s, SOCKET_ADDRESS);
+            sockets [j] = s;
+            nn_thread_init (&threads [j], routine, &sockets [j]);
+        }
+
+        /*  Allow all threads a bit of time to connect. */
+        nn_sleep (100);
+
+        test_send (sb, "");
+        
+        for (int j = 0; j < TEST_THREADS; j++) {
+            test_close (sockets [j]);
+            nn_thread_term (&threads [j]);
+        }
+
+        test_close (sb);
+    }
+
+    return 0;
+}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -106,6 +106,7 @@ add_libnanomsg_test (zerocopy)
 add_libnanomsg_test (shutdown)
 add_libnanomsg_test (cmsg)
 add_libnanomsg_test (bug328)
+add_libnanomsg_test (ws_async_shutdown)
 
 #  Build the performance tests.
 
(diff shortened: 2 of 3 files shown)


gdamore:
Is this restricted to Winsock?

gdamore:
Oh, I see you answered that.

wirebirdlabs:
Yikes; lots of syntax errors on `Visual Studio 11` generator -- will fix these and commit on the same branch. That said, NMake shows the failure.

gdamore:
Note that I've noticed on Windows TCP slow start takes ~forever.  Like 200 msec is not enough!!  (But 1 second is.  I've not tried to narrow it further.  On Linux, 100 msec is usually but not always enough, and 200 is always enough.  100 is always enough on Darwin.

gdamore:
I wonder if this is TCP delayed ack -- 200 msec is the delayed ack time.  We need to become TCP_NODELAY by default, I think, and actually enable that.

wirebirdlabs:
> Windows TCP slow start takes ~forever

Perhaps .... but if this is the case, we would see just as many failures in this test case submitted for TCP transport as WS transport, yes?

tests/ws_async_shutdown.c
```diff
@@ -0,0 +1,104 @@
...
+            s = test_socket (AF_SP, NN_SUB);
+            test_setsockopt (s, NN_SOL_SOCKET, NN_RCVTIMEO,
+                &rcvtimeo, sizeof (rcvtimeo));
+            test_setsockopt (s, NN_SUB, NN_SUB_SUBSCRIBE, "", 0);
+            test_connect (s, SOCKET_ADDRESS);
+            sockets [j] = s;
+            nn_thread_init (&threads [j], routine, &sockets [j]);
+        }
+
+        /*  Allow all threads a bit of time to connect. */
```
gdamore: Try increasing this to 1000.  Then what happens?

wirebirdlabs:
Check it -- Travis reports a failure on CLang: https://travis-ci.org/nanomsg/nanomsg/jobs/92822123#L710

tests/ws_async_shutdown.c
wirebirdlabs: Will try before fixing syntax in next commit on this PR ...

wirebirdlabs: Tried increasing to 1000msec -- that did not help! This does not appear to be a problem with TCP slow start.

wirebirdlabs:
Fixed the C89 compile error and repro'd with MSVC: https://ci.appveyor.com/project/gdamore/nanomsg/build/1.0.190/job/516sh4pag3876gx0#L1547

By the way, @gdamore -- I've been testing all your `nn_close` improvements, and this PR represents the *final* issue remaining that I'm currently able to characterize and provide a minimal test case for -- thank you kindly for so much work flushing out all the other little shutdown issues!! Everything else is working quite well.

wirebirdlabs:
OK, we've got an NMake failure (https://ci.appveyor.com/project/gdamore/nanomsg/build/1.0.190/job/ib0e3elum2j4tufa#L692)

And this time, CLang reported no error ...

... but 6minutes in, gcc is about to fail.

This is definitely a problem with the WebSocket transport that periodically affects all platforms.

gdamore:
Argh.  Ok, I guess I get to have fun debugging this.  The one error QUEUE_NOT_INQUEUE is known.  Closing that is going to take some time since its very very hard to reproduce.  If this Winsock is easier to repro, it will help.

gdamore:
It does indeed appear that you can repro it some times on Linux, since the Travis failure looks like a hang in this test case.  Oh joy.

nanomsg is like the redheaded adopted special-needs step-child.  It keeps demanding my attention, pulling it from a zillion other things (natural children, etc.) I'd rather be working on (including, btw, mangos).  The adoption seemed like maybe a good idea at the time, but now over a year into it, I'm having regrets.

I sure wish we'd get some other qualified volunteers to pitch in more. :-/

tests/ws_async_shutdown.c
gdamore: So the thing that would help me the most, is a bit more information about the nature of the hang.  What's stalling out?  It doesn't make sense exactly that nn_close() is hanging, and if it is I'd love to have a more complete stack trace -- including for all threads.

Getting this on Windows is something I'm not qualified to do -- I know *nothing* about the Windows debugging tools, and I prefer to remain blissfully ignorant.  :-)

wirebirdlabs:
Stack trace of hang on Windows:

```
ntdll.dll!NtWaitForSingleObject() + 12 bytes	Unknown
[Frames below may be incorrect and/or missing, no symbols loaded for ntdll.dll]	
KernelBase.dll!WaitForSingleObjectEx() + 183 bytes	Unknown
KernelBase.dll!WaitForSingleObject() + 18 bytes	Unknown
nanomsg.dll!nn_sem_wait(nn_sem * self=0x0047e378) Line 122	C
nanomsg.dll!nn_sock_term(nn_sock * self=0x0047e0f0) Line 219	C
nanomsg.dll!nn_close(int s=0) Line 591	C
ws_async_shutdown.exe!test_close_impl(char * file=0x0014a8b8, int line=100, int sock=0) Line 119	C
ws_async_shutdown.exe!main(...) Line 100	C
ws_async_shutdown.exe!__tmainCRTStartup() Line 626	C
ws_async_shutdown.exe!mainCRTStartup() Line 466	C
```

and there's only 1 other thread -- an `nn_worker_routine`:

```
ntdll.dll!NtRemoveIoCompletionEx() + 12 bytes	Unknown
[Frames below may be incorrect and/or missing, no symbols loaded for ntdll.dll]	
KernelBase.dll!GetQueuedCompletionStatusEx() + 132 bytes	Unknown
nanomsg.dll!nn_worker_routine(void * arg=0x61626b40) Line 168	C
nanomsg.dll!nn_thread_main_routine(void * arg=0x61626b60) Line 30	C
msvcr120d.dll!_beginthreadex() + 417 bytes	Unknown
```

gdamore:
Ok thanks.  Efd bug most likely.

gdamore:
Good news, it reproduces on MacOS X, which means I won't have to go through insane hurdles to debug.

gdamore:
So I have a theory.  The theory is that we aren't waiting long enough for the nn_socket to bind/connect, and we're stuck waiting for that due to a left over item from the earlier invocation.  I think the problem here is that the transports may leave the TCP port in a mildly unusable state.

gdamore:
Interestngly enough, waiting a second after the nn_close of the sb before restarting the loop seemed to have alleviated the problem.

gdamore:
but maybe not conclusively.

wirebirdlabs:
> we aren't waiting long enough for the nn_socket to bind/connect, and we're stuck waiting for that due to a left over item from the earlier invocation

One significant difference between TCP and WS is that TCP will attempt to rebind to and already-bound socket: https://github.com/nanomsg/nanomsg/commit/1936472d557d5b7a23768873a4b2c86a85abcab7

Comparing, since this failure mode does not seem to affect TCP transport, just WS.

gdamore:
Its a statemachine failure I think.  I'm pretty sure I've found it.  The state machines here are a twisty maze, and no wonder there are bugs here.  Argh.

gdamore:
Interesting.  So mangos reconnects winsock as well as TCP.  This is a deficiency in the winsock implementation, IMO.  Yet another reason to not use this implementation of the transport in production.

wirebirdlabs:
> mangos reconnects winsock as well as TCP

I don't really understand this statement; are you saying this specific problem is on re-using a port?

I'm not exactly confident that port re-use is the problem here. In a separate application test suite, where I first found this bug, a similar stress test was choosing ephemeral ports rather than re-using ports.

wirebirdlabs:
> choosing ephemeral ports

And before this raises more questions than answers ... it chose this ephemeral port at a higher layer than a bona fide * on `nn_bind()` (since libnanomsg does not yet support this), meaning there does exist a known race condition between when the port chosen and whether it's still available by the time `nn_bind()` was invoked. But for all intents and purposes, I think we can assume there were no collisions.

gdamore:
No, I meant that mangos will keep trying to connect or bind on failure.  We do that in mangos for all transports. not just a select few.  The problem here is that in libnanomsg we have to implement a state machine (from hell) for every transport.  So there is almost no code sharing and every transport has to figure the reconnect strategy out for themselves.  Which is why you don't have it in winsock.

Port reuse can be a problem depending on which side disconnects, and if there are problems there with address use.  Google around for EADDRINUSE for a bit more detail here; the TCP state machine requires in some circumstances up to 2*RTT to ensure that no outstanding sequence numbers from an earlier connection can come back around and confuse things.

wirebirdlabs:
> Which is why you don't have it in winsock

You keep saying winsock. Do you mean WebSocket?

wirebirdlabs:
Of course. This thread makes a lot more sense now reading thru. Taking a look thru https://github.com/nanomsg/nanomsg/commit/1936472d557d5b7a23768873a4b2c86a85abcab7, it appears that it wouldn't be hard for me to port this to the websocket transport; unless you're already down this path, I'll take a look into this next couple days.

gdamore:
sorry, yes websocket.  Argh.

gdamore:
I am trying to debug this at present.  Believe me, nothing would please me more than to have someone else take point on this.

gdamore:
The sem_wait is on the termsem, its waiting for the EPS to all post their completion.  Some one isn't doing that.  I don't know why yet.

gdamore:
So looking at this, I see that aws state machines hanging off the aws for each of those connected sockets; at first blush it looks like the far end disconnect (close) didn't wind up causing the socket to close.  Or, perhaps we didn't close.  I'm not sure yet.

gdamore:
So another difference is that with websocket, the server only recvs, and doesn't try to send until it gets a client recv, whereas TCP sends a message in both directions.  This can leave a connected socket hanging.  The FSM probably doesn't detect a closed socket properly.   This is one of the major drawbacks to *not* having a proper blocking recv on the socket.  Eventing the close of the underlying TCP socket is a pain.  (Simple single threaded recv is *much* simpler.)

wirebirdlabs:
That said ... perhaps it would help to probe the statistics structure while the socket is hung? Will try this to see if anything jumps out ...

wirebirdlabs:
screen shot 2015-11-23 at 9 43 47 pm

wirebirdlabs:
10 `accepted_connections` and 10 `broken_connections` would tend to indicate that aws detected the break at this line: https://github.com/nanomsg/nanomsg/blob/master/src/transports/ws/aws.c#L272

gdamore:
My brain is tired from tracing through FSM hell.  I'm going  to declare this is a Websocket bug and leave it in your court to fix.  WebSocket is experimental, and it needs to remain that way until at least it can re-establish failed connections (from the client side).

gdamore:
That said, the mangos WebSocket transport has proven itself to be quite robust and flexible.  So its not the nature of WebSocket that is a problem, but of the libnanomsg implementation limitations.

wirebirdlabs:
> brain is tired from tracing through FSM hell

Déjà vu and sympathy pains here, flashback 1yr ago. That said. I'll look into this issue.

wirebirdlabs:
OK, I've effectively implemented parity in WebSocket with TCP in terms of NN_RECONNECT_IVL and backoff timers.

This does not seem to yet help this issue, so it will be coming in as a separate PR unrelated to this hopefully later today.

That said, while poring over differences, there are a couple things I noticed that could potentially help both TCP and WebSocket.

Stay tuned ...

wirebirdlabs:
So ... it might appear that, unceremoniously, this is the culprit for this hang:

https://github.com/nanomsg/nanomsg/blame/master/src/transports/ws/aws.c#L316

Compare that to the corresponding lines in TCP:

https://github.com/nanomsg/nanomsg/blame/master/src/transports/tcp/atcp.c#L293-L294

I seem to recall making that change for a specific reason (it's been >1yr now, that memory is rusty) -- but it was in order to properly close the TCP connection in order to best satisfy Autobahn TestSuite.

Without making the best effort to shut down the connection properly, Autobahn would throw failures/warnings for its peer-under-test not performing clean shutdowns.

So that said ... it's not entirely clear how to move forward, but it *feels* like the best thing to do is tend to migrate WebSocket (https://github.com/nanomsg/nanomsg/blame/master/src/transports/ws/aws.c#L316) over to TCP (https://github.com/nanomsg/nanomsg/blame/master/src/transports/tcp/atcp.c#L293-L294), rather than the trivial vice-versa that simply seems to fully satisfy the test in this PR #551.

@gdamore this could be related to the root cause of why it appears Windows takes >1sec to recover from some of the more brutal tests -- nanomsg may not be currently shutting down TCP in the best way possible.

gdamore:
I don't think this is related to the long Windows TCP connect times.  It feels like something else there.

wirebirdlabs:
This PR is superseded by https://github.com/nanomsg/nanomsg/pull/563 and will be closed once #563 is acknowledged.

wirebirdlabs:
Superseded by #565. Closing.

PR #553, opened 2015-11-24
fixes #324 inproc doesn't transport message headers

### LINKED ISSUE #324: inproc doesn't transport message headers
Make a REQ/REP socket pair connected via inproc. Send a message to the REQ socket. The body of the message arrives at REP. Not so message headers.

gdamore:
This basically disables the zero-copy optimization in inproc (which was incorrect), and makes it just do a data copy.  The result is that it works though.


gdamore:
I integrated this already, after rebasing it.

Issue #556, opened 2015-11-25
Support security attributes for Windows IPC


kostyay:
In the following scenario:
- Service process runs as windows service or (as an elevated process), binds to a channel such as `ipc://test`
- Client process that runs as a weak user tries to connect to `ipc://test`, it will fail.

This is because when the named pipe is created by the parent process in `usock_win.inc:510` `NULL` is passed at the last parameter for lpSecurityAttributes. This causes the pipe to be created with the parent process permissions.

I can see that there are some TODOs there that mention this.
Any plans to fix that?

gdamore:
There was a PR submitted for this #348.  I closed it as the author of it has stopped working on libnanomsg, and the code doesn't merge cleanly, and I am not enough of a Windows guy to take it on myself.  If you want to pick this work up, clean it up, and submit a new PR, I'd be happy to accept it.

In the meantime, I recommend you use TCP instead, if you can.

kostyay:
Thanks, I will look at it.

And yeah, we switched to TCP for now.

kostyay:
@gdamore here, take a look https://github.com/nanomsg/nanomsg/pull/557

PR #557, opened 2015-11-25
Support Pipe Security attributes in Windows [updated]

kostyay:
Basically I rebased and fixed some of the styling comments of https://github.com/nanomsg/nanomsg/pull/348.

Not sure about this comment: https://github.com/nanomsg/nanomsg/pull/348/files#r42952274

### DIFF
--- a/src/transports/ipc/aipc.c
+++ b/src/transports/ipc/aipc.c
@@ -81,6 +81,7 @@ int nn_aipc_isidle (struct nn_aipc *self)
 
 void nn_aipc_start (struct nn_aipc *self, struct nn_usock *listener)
 {
+    size_t sz;
     nn_assert_state (self, NN_AIPC_STATE_IDLE);
 
     /*  Take ownership of the listener socket. */
@@ -88,6 +89,14 @@ void nn_aipc_start (struct nn_aipc *self, struct nn_usock *listener)
     self->listener_owner.src = NN_AIPC_SRC_LISTENER;
     self->listener_owner.fsm = &self->fsm;
     nn_usock_swap_owner (listener, &self->listener_owner);
+
+#if defined NN_HAVE_WINDOWS
+    /* Get/Set security attribute pointer*/
+    nn_epbase_getopt (self->epbase, NN_IPC, NN_IPC_SEC_ATTR, &self->usock.sec_attr, &sz);
+
+    nn_epbase_getopt (self->epbase, NN_IPC, NN_IPC_OUTBUFSZ, &self->usock.outbuffersz, &sz);
+    nn_epbase_getopt (self->epbase, NN_IPC, NN_IPC_INBUFSZ, &self->usock.inbuffersz, &sz);
+#endif
 
     /*  Start the state machine. */
     nn_fsm_start (&self->fsm);
--- a/tests/win_sec_attr.c
+++ b/tests/win_sec_attr.c
@@ -1,0 +1,152 @@
+/*
+    Copyright (c) 2015 Timothee "TTimo" Besset  All rights reserved.
+
+    Permission is hereby granted, free of charge, to any person obtaining a copy
+    of this software and associated documentation files (the "Software"),
+    to deal in the Software without restriction, including without limitation
+    the rights to use, copy, modify, merge, publish, distribute, sublicense,
+    and/or sell copies of the Software, and to permit persons to whom
+    the Software is furnished to do so, subject to the following conditions:
+
+    The above copyright notice and this permission notice shall be included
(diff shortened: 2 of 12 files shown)


src/transports/ipc/aipc.c
```diff
@@ -89,6 +90,14 @@ void nn_aipc_start (struct nn_aipc *self, struct nn_usock *listener)
     self->listener_owner.fsm = &self->fsm;
     nn_usock_swap_owner (listener, &self->listener_owner);
 
+#if defined NN_HAVE_WINDOWS
+    /* Get/Set security attribute pointer*/
+    nn_epbase_getopt (self->epbase, NN_SOL_SOCKET, NN_SEC_ATTR, &self->usock.sec_attr, &sz);
```
gdamore: So these should be transport level options, not NN_SOL_SOCKET, as they apply to the IPC transport rather than to the socket as a whole.

tests/win_sec_attr.c
```diff
@@ -0,0 +1,98 @@
...
+    nn_assert (ret);
+
+    xa.grfAccessPermissions = GENERIC_READ | GENERIC_WRITE;
+    xa.grfAccessMode = SET_ACCESS;
+    xa.grfInheritance = SUB_CONTAINERS_AND_OBJECTS_INHERIT;
+    xa.Trustee.TrusteeForm = TRUSTEE_IS_SID;
+    xa.Trustee.TrusteeType = TRUSTEE_IS_WELL_KNOWN_GROUP;
+    xa.Trustee.ptstrName  = (LPSTR) &SIDAuthUsers;
+    ret2 = SetEntriesInAcl (1, &xa, NULL, &pACL);
+    nn_assert (ret2 == ERROR_SUCCESS);
```
gdamore: While this asserts that you can create this security attribute, I don't see how it validates that the IPC is actually created with that.  I'm not a windows guy, so I don't understand how you could even attempt to validate that -- how can you check the permissions of an Named Pipe?  I'd really like to see that verified though.

gdamore:
Thanks for cleaning this up.  See my comments on the test suite, though, and you should change to using a transport socket option instead of a socket layer option. 

For an example, look at how we set NN_WS_MSGTYPE in the websocket transport.

gdamore:
Oh yeah, eventually this need to be rebased into a single commit, too.   I try to have each integration into the code base be code complete; we don't expose the sausage factory behind each feature or bug fix to downstreams.  (This also means, btw, that we're trying to get to an "always release ready" state, where any arbitrary bisect or checkout can be used to get a functional nanomsg.)

src/transports/ipc/aipc.c
kostyay: Is it ok if they will be still defined in `sock.h` and the processing will be done in `sock.c`?
I'm not familiar with the project structure that much.

gdamore: These options should be defined in ipc.h.  (I'm less sure about the buffer options, but I think so, there as well.)  The handling  here in aipc.c is the right place.

kostyay:
@gdamore 
I made the fixes as you asked.
Please have a look.
As for the test I tried to improve it abit.
I will check tomorrow how its possible to query a pipe for its security attributes and improve the test.

kostyay:
How do I merge all the commits into a single one?

gdamore:
Do an interactive rebase and you can squash the commits.

reqshark:
and also you'll need to force push back to your branch after the `git rebase -i`

kostyay:
@gdamore 
It took me a while to figure this out but I wrote a good tests that verifies the pipe was created with the proper security descriptor.
I also squashed the commits so everything should be ready to go now.

kostyay:
Any chance for a merge? Its been 8 days now since I fixed everything and no review :(

gdamore:
I'm looking at this now.  Sorry, but I'm not paid to work on this, so it takes time sometimes.

gdamore:
I've merged this in, after editing the commit comments:

commit c621fb490699dbae8594135f324edb13fb8dc64d
Author: Timothee 'TTimo' Besset <ttimo@ttimo.net>
Date:   Tue Sep 16 09:10:17 2014 -0500

    fixes #556 Support security attributes for Windows IPC
    
    Allow passing security descriptor to the ipc channel in Windows.
    The tunables NN_IPC_SEC_ATTR, as well as NN_IPC_OUTBUFSZ and
    NN_IPC_INBUFSZ are exposed on Windows when using the named pipe
    based IPC.

kostyay:
Thanks :)

Issue #528, opened 2015-11-17
NN_QUEUE_NOTINQUEUE fix incomplete


gdamore:
Just saw this on Travis:

Assertion failed: self->next == NN_QUEUE_NOTINQUEUE (../src/utils/queue.c:102)
/bin/bash: line 5: 13356 Aborted                 (core dumped) ${dir}$tst

Arrgh.  This was the ipc_shutdown test.

gdamore:
Well, this seems hard to hit.  I'm several thousand invocations into this, and I can't seem to reproduce.  Aargh.

wirebirdlabs:
Link to failure: https://travis-ci.org/nanomsg/nanomsg/jobs/91501857#L667

It's a stretch, but to help brainstorm: is it possible that https://github.com/nanomsg/nanomsg/blob/master/tests/ipc_shutdown.c#L82-L84 ignoring of the SIGPIPE error is now hurting us?

gdamore:
I don't think so.  That signal shouldn't have been delivered on Linux ever because we are MSG_NOSIGNAL

gdamore:
Wow.  I've been running this test in a while loop for *hours*, and haven't been able to reproduce it.

wirebirdlabs:
> in a VM

Idea: drastically throttle the VM; 1 core, 1GB memory ... try to repro what are probably relatively "low resource" conditions on Travis.

gdamore:
I think my vm is kind of constrained.

gdamore:
I'm starting to think this might be related to #554 -- that is I think nn_term() races against the REQ/REP state machine for example.

gdamore:
That said, ipc_stress uses PUB/SUB.

JackDunaway:
Bump. This is continually failing in Travis CI -- here's the latest failure: https://travis-ci.org/nanomsg/nanomsg/jobs/123991735#L642

This bug should be bumped to one of the highest priorities, especially since it's still not well-understood what causes the issue.

gdamore:
Yes, I need to spend time working on it again.  :-)  Trying to get the
other stuff stabilized first.

gdamore:
I have a hope/theory that the changes I made today to make nn_close() and nn_term() more resilient may have fixed this.  Investigation pending -- lets keep an eye out for this failure since I've not seen it in quite a while now.

gdamore:
This is at least a release stopper -- we need to find out if it still happens.

[...]

PR #558, opened 2015-11-25
fixes #554

### LINKED ISSUE #554: weird hang/crash & race against nn_term
I think there is a problem with the req/rep state machines racing with nn_term.

https://ci.appveyor.com/project/gdamore/nanomsg/build/1.0.193/job/haswhuvgk3tv2lgr

I can't explain this crash, or hang, any other way; something doesn't add quite up.  The problem is seriously complicated to debug, due to the gazillion state machines involved. 

Interestingly enough, increasing the timeout for the connection request, which allows the TCP connections to be established properly, seems to clear this problem up.  (Its also relevant that this problem only was seen on Windows, which seems to require quite a bit longer for TCP to establish.)

wirebirdlabs:
Fingers crossed, this fixes #554 by releasing holds within the global lock. Even if not a full fix, it's anecdotally been helping with shutdown race conditions, especially in the WebSocket transport in and around failing connections with the proper RFC 6455 Close Handshake.

Although we typically strive to take things out of locks rather than put them in, these operations should be extremely fast, adding little penalty.

### DIFF
--- a/src/core/global.c
+++ b/src/core/global.c
@@ -578,12 +578,12 @@ int nn_close (int s)
         This is done with the glock held to ensure that two instances
         of nn_close can't access the same socket. */
     nn_sock_stop (sock);
-    nn_glock_unlock ();
 
     /*  We have to drop both the hold we just acquired, as well as
         the original hold, in order for nn_sock_term to complete. */
     nn_sock_rele (sock);
     nn_sock_rele (sock);
+    nn_glock_unlock ();
 
     /*  Now clean up.  The termination routine below will block until
         all other consumers of the socket have dropped their holds, and


gdamore:
This feels right to me.  And shame on me for screwing it up in the first place. Thanks for digging in and fixing.

gdamore:
I'm inclined to accept this PR, but I'm not yet convinced it is the fix for #554.  It may be, and it may also fix the QUEUE_NOTINQUEUE assertion failures too.  (In fact I think this is pretty likely either a fix, or a big improvement.)

So what I'd like to do is open a separate bug, indicating that the lock should be held (synopsis like "Socket release needs to be done under lock") and then wait and see if we can close the others based on more testing & time.

What do you think?

gdamore:
I think this may fix #528

gdamore:
I went ahead and commited this as

commit fd66ff55a5bad44ea0c3cca8bea345b6f02663bf
Date:   Wed Nov 25 16:26:08 2015 -0600

    fixes #560 Socket released without global lock

PR #562, opened 2015-11-29
attempt to fix device7 hang

gdamore:


gdamore:
Whoa.  Got a segfault in device7 this time.

wirebirdlabs:
So ... thoughts on this. The current implementation of `nn_term()` just pulls the rug from underneath a structured application shutdown, and I cannot imagine recommending the usage of this method in a production application.

Even in a non-production application, it's just as well to kill the process as to blanket destroy any and all active sockets, if quick-n-dirty makes more sense than taking the time to design a structured shutdown.

As for the use-case of shutting down a device -- the application recommendation should be to create the socket (or perhaps both sockets) in a local thread prior to passing its ownership to an async thread where the device is created to combine the two sockets.

This way, the application developer should always have a file descriptor to at least one socket part of the device in order to call `nn_close()`, which would effectively bring down the device cleanly.

Saying the same thing from a different perspective -- the `device7` test goes against best practices of creating at least one socket locally, but instead creates both sockets in the same async thread as the creation of the device: https://github.com/nanomsg/nanomsg/blob/master/tests/device7.c#L45-L51 When this best-practice is overlooked, "needing" (air-quotes) `nn_term()` in its current manifestation is contrived, and not a case we should test for or even support.

I propose that we officially deprecate the `nn_term()` method immediately, and especially prior to v1.0.

gdamore:
I'd had much the same thought.  However my concern is library and runtime consumes that may desire a handy way to just shut everything down as other wise they need to keep track of every socket. 

Application code should do this sure.  But it's those frameworks that I'm worried about.  I'm not quite ready to throw in the towel on nn_term.

wirebirdlabs:
You're right.

JackDunaway:
Closing this PR for now due to a new strategy for shutdown with #694. @gdamore -- feel free to reopen if necessary. Thanks!

PR #563, opened 2015-12-01
WebSocket Robustification

wirebirdlabs:
This PR is bigger than normal. More on that later*

Major related improvements include:

1) Retry timer for `nn_bind()` implemented
2) `nn_close()` stability for multi-threaded environment
3) Bringing Autobahn test back up to date with latest WebSocket API changes
4) Better encapsulation of the Closing Handshake as an internal implementation in the library, rather than exposing this message to the libnanomsg end-user. Said another way, it is treated more like a protocol violation and handled at the transport layer, rather than leaking out to the SP or end-user layers (effectively building upon the work by @zakharov, but providing more of a best-effort to adhere to the RFC 6455 Closing Handshake)
5) Implementing NN_RCVMAXSIZE for WebSocket
6) Overall better test coverage

Significant parts of this commit are ports of work done previously for the TCP transport by @bnewbold and @vseryakov 

*It departs from a typical "fix" PR, but instead tackles several related issues at once that would be difficult (but not impossible) to fix separately. The commits are atomic enough at least to pull in the good parts, if we need to do a little restructuring of this PRsaurus.

Finally -- it's worth noting that the Autobahn test is currently hard-coded to run. My intention is to have this initial PR kick off the CI environments, make sure they all work, then turn this test off by default and upgrade the two CI definition files to enable only on the CI servers (including, downloading/installing the Autobahn package). This will come in the form of more commits on top of this PR. Stay tuned ...


wirebirdlabs:
Also -- it's expected that the `ws_stress` test will initially fail -- after verifying it fails, the first commit to roll in will be turning this test off by default to kick off the next test.

wirebirdlabs:
As expected, disabling `ws_stress` has enabled the tests to pass. Now, on to updating the CI specs to download/install Autobahn python package, and enable the test by default (this may not happen today... stay tuned)

wirebirdlabs:
OK, time to hang my hat up for a while, and also request some help:

Briefly, there is a CI environment variable (https://github.com/nanomsg/nanomsg/pull/563/files#diff-354f30a63fb0907d4ad57269548329e3R7) that needs to override a defined constant in the source code (https://github.com/nanomsg/nanomsg/pull/563/files#diff-1f8bb897866ba4f42b43d5b44a42197bR42), only in the context of the CI environments.

It seems this task will take just a small handful of strategically-placed or modified lines somewhere in the chain between (and including) these two places, very likely delving into autoconf-land and touching `configure.ac`, but this is an area where I'm just not experienced.

Help? We're close.

wirebirdlabs:
This is nearly working on Linux. Currently, it's failing here: https://github.com/nanomsg/nanomsg/pull/563/files#diff-1f8bb897866ba4f42b43d5b44a42197bR250

Which is not surprising. Will have to figure out a clever way to resolve this.

gdamore:
So I've not looked this all over yet, and its going to take some time.

Really, I'd like to have the autobahn test suite as a separate PR from the other fixes.  I realize you're doing this work together, but once you've got it working, how hard do you think it would be to separate those?

I've got some pretty mixed feelings about enabling autobahn by default.  Having to run sudo for the tests means that CI is going to take longer since it cannot use containers.  Its also going to make life harder for other folks.

At the end of the day, strict autobahn conformance is one of those things that seems "nice to have", but not entirely necessary, provided we interoperate with other websocket implementations.

wirebirdlabs:
> At the end of the day, strict autobahn conformance is one of those things that seems "nice to have", but not entirely necessary, provided we interoperate with other websocket implementations

It's not really about strict Autobahn conformance, or even strict RFC 6455 conformance -- it's about using Autobahn in order to test the code paths, especially things like chunked transfers, opening handshake compliance, what would happen if browsers/agents decide to ping ... basically a lot of things that are outside the scope of the peer's nanomsg implementation, but are discretions of the underlying transport itself.

> I've got some pretty mixed feelings about enabling autobahn by default

It's only enabled by default on CI environments (well, getting there...), and only if explicitly-requested if you are building from source.

> Having to run sudo for the tests means that CI is going to take longer since it cannot use containers.

True; it adds time (roughly +90sec), but once enabled, the WebSocket transport should be the most comprehensively-tested transport in libnanomsg (e.g., we don't have a fuzzer sending known-bad packets to TCP). I think the tradeoff is worthwhile for the quality.

I would like to investigate how we can get away from needing `sudo` so we can move back to containerized environments, and also cache the dependencies.

> Its also going to make life harder for other folks

It shouldn't -- this was an important design goal when adding this test. It's opt-in only, using `--enable-autobahntest=yes` on the `configure` step.

wirebirdlabs:
@gdamore -- looks like AppVeyor configuration is currently controlled thru your web interface at https://ci.appveyor.com/project/gdamore/nanomsg rather than `appveyor.yml` in the repo root -- would you mind stubbing out the current settings into a new file?

For reference, here's the template provided by AppVeyor docs: http://www.appveyor.com/docs/appveyor-yml

wirebirdlabs:
> @gdamore -- looks like AppVeyor configuration is currently controlled thru your web interface

That said ... it appears the default settings for a project (looking at my own AppVeyor account) allow appveyor.yml to take precendence by default ... I'll go ahead and try committing one...

wirebirdlabs:
It worked! @gdamore, I can take it from here for the time being (not blocked by your admin account ownership after all). Will also trace thru old build logs to recreate the proper settings.

wirebirdlabs:
@gdamore I will spend time today squashing these 62 commits into distinct, coherent PR's.

I hope this makes your job merging much faster.

wirebirdlabs:
> Having to run sudo for the tests means that CI is going to take longer since it cannot use containers

FYI, this may be a temporary limitation: https://docs.travis-ci.com/user/migrating-from-legacy/?utm_source=legacy-notice&utm_medium=banner&utm_campaign=legacy-upgrade#Using-sudo-isn%E2%80%99t-possible-(right-now)

gdamore:
As long as they run Docker, it isn't going to get fixed, because Docker isn't true isolation (unlike e.g. Solaris zones.)

I'd still prefer to avoid requiring sudo.  Why does autobahn require this?  Is it for binding to port 80/443?  (And can that be changed, if it is?)  Or is it something else?

wirebirdlabs:
> Why does autobahn require this?

Only a package installation: https://github.com/nanomsg/nanomsg/pull/563/files#diff-354f30a63fb0907d4ad57269548329e3R12

Travis CI recommends building an executable binary, storing that to your own server (or repository), then copying it into the environment, as one solution that doesn't require `sudo`.

Additionally, that would cut appreciable time off the build.

gdamore:
Does installation require root for some reason (why)?  I mean, its fairly typical that we can install software in the Docker instance without problems...  You might need to select a different destination directory, but usually that's not too hard to do...

gdamore:
Check this: http://kazhack.org/?post/2014/12/12/pip-gem-install-without-sudo

wirebirdlabs:
> Check this

Great tip. I'll look into that, and trigger a build here in a minute that removes the `sudo` in order to observe the failure mode.

In the meantime, check out the latest commit -- since it may be relevant for issues you've been chasing: https://github.com/nanomsg/nanomsg/commit/d15451743ed645c4b83a23865e4e2f7164b88359

and a link to the line number in the log of the associated failure mode: https://travis-ci.org/nanomsg/nanomsg/jobs/94521740#L877

wirebirdlabs:
Wow ... `--user` flag is all it took to get back to container environment -- we're back in business! No `sudo`, faster performance.... even Travis CI recommends this: https://docs.travis-ci.com/user/languages/python#pip

Well, that was a freebie.

A refreshing treat, after dealing with persnickety, mind-numbing autotools ...

wirebirdlabs:
OK, the rebasing begins -- everything is working quite well, except for two tests -- `tcp_shutdown`, and also `device7`

The Autobahn test has been consistently passing on both platforms, as with the `ws` test.

wirebirdlabs:
Superseded by #565. So long, infernal thread of autotools strife. Closing.

PR #564, opened 2015-12-03
Tentative fix to a periodically-observed failure in ipc_shutdown

wirebirdlabs:
Link to one such failure, observed periodically while kicking off dozens of builds developing WebSocket transport and CI build process: https://travis-ci.org/nanomsg/nanomsg/jobs/94521740#L877

### DIFF
--- a/src/transports/utils/streamhdr.c
+++ b/src/transports/utils/streamhdr.c
@@ -259,7 +259,9 @@ invalidhdr:
         switch (src) {
 
         case NN_STREAMHDR_SRC_USOCK:
-            /*  It's safe to ignore usock event when we are stopping  */
+            /*  It's safe to ignore usock event when we are stopping, but there
+                is only a subset of events that are plausible. */
+            nn_assert (type == NN_USOCK_ERROR);
             return;
 
         case NN_STREAMHDR_SRC_TIMER:
@@ -286,6 +288,12 @@ invalidhdr:
 /******************************************************************************/
     case NN_STREAMHDR_STATE_STOPPING_TIMER_DONE:
         switch (src) {
+
+        case NN_STREAMHDR_SRC_USOCK:
+            /*  It's safe to ignore usock event when we are stopping, but there
+                is only a subset of events that are plausible. */
+            nn_assert (type == NN_USOCK_ERROR);
+            return;
 
         case NN_STREAMHDR_SRC_TIMER:
             switch (type) {


wirebirdlabs:
@gdamore any problems merging this one? It's hard to keep making forward progress juggling branches away from mainline.

gdamore:
So how have you confirmed this fixes the problem?  I've not had time yet to analyze the code, although your change *seems* low risk to me.

wirebirdlabs:
So, confirmed, no, because this issue is so infrequently reproduced.

I spent some time trying to trace out all the code paths to confirm this fix is correct, but retracing my (original) inductive reasoning in order to give you a definitive answer, I'm not 100% confident that this patch is not just masking the real problem.

The reasoning goes something like this:

When we do observe the failure, it's clear what failed, because of the log. (link in original PR, here's a copy for convenience: `Unexpected source: state=5 source=1 action=5 (../src/transports/utils/streamhdr.c:306)`)

To dereference those numbers:

state=5 is `NN_STREAMHDR_STATE_STOPPING_TIMER_DONE`
source=1 is NN_STREAMHDR_SRC_USOCK
action=5 is NN_USOCK_ERROR

How could we have gotten to the `NN_STREAMHDR_STATE_STOPPING_TIMER_DONE` state? From only one location, where USOCK must have just posted the `NN_USOCK_RECEIVED` action: https://github.com/nanomsg/nanomsg/blob/9a0cb281695759fde449244b5f664ece5075d9f5/src/transports/utils/streamhdr.c#L227

Is it possible that the USOCK could immediately post `NN_USOCK_ERROR ` just after `NN_USOCK_RECEIVED` but before `NN_STREAMHDR_SRC_TIMER` posts `NN_TIMER_STOPPED`? Conceivably -- and if so, that would be evidenced as the assertion in this case.

That said ... it's hard to trace the logic between the timer and the usock to prove whether this case could definitely exist.

gdamore:
So given that response, I'm inclined not to rush this in, but to wait until I have some more time to do a complete analysis to verify that this is indeed the correct fix.

wirebirdlabs:
understandable. and agree.

gdamore:
Looking at this, I can see ways this can happen.  For example, if the remote peer closes the connection, or someone else closes the connection, just after a receive is done.  Given that this is a shutdown/race close most of the time, this aligns with the expectation.  I'm going to go ahead and merge this, because the changes look good.  The only thing I've not been 100% certain about was whether the assertions you've included are actually true.  I think they, and if they're not, it will be good to know.  (If someone triggers either of those, we can consider other actions, which may include removing the assertion, fixing the bug, or backing this change out.)

gdamore:
I've updated the comments, and pushed as:

commit c0e92dd84f0263fb4f2ad504dd7ccbdf42a24c21
Date:   Wed Dec 2 20:47:30 2015 -0600

    fixes #568 ipc_shutdown still fails sometimes

wirebirdlabs:
> The only thing I've not been 100% certain about was whether the assertions you've included are actually true. I think they, and if they're not, it will be good to know

Precisely my logic as well. It's not awesome to be *adding* assertions rather than taking them away from the library, but it felt weird/wrong to blanket-ignore all events, without defining a reasonable subset.

PR #565, opened 2015-12-03
WebSocket mega-squash

wirebirdlabs:
This single PR is a pretty coherent packaging of 82 commits made during the integration of Autobahn TestSuite integration with Travis CI and AppVeyor in order to fully test the WebSocket transport. Each individual commit is designed to be digestible for code review, and building upon each previous commit should apply cleanly against the master.

PR #569, opened 2015-12-09
WebSocket (supersedes #565)

JackDunaway:
#565 no longer cleanly merges with master -- here's the same work, just with no merge conflicts, and two more squashed commits


wirebirdlabs:
With regards to the test failure: https://travis-ci.org/nanomsg/nanomsg/jobs/95899479#L1095

That's a late-breaking known issue that will likely be resolved shortly: https://github.com/crossbario/autobahn-testsuite/issues/55

@gdamore if you prefer we can disable the test in the meantime by changing one line in both CI specs:

Travis CI: https://github.com/wirebirdlabs/featherweight-nanomsg/commit/28f6c213f903d94ceb5e7754fb248a78722b6fd9#diff-354f30a63fb0907d4ad57269548329e3R13

AppVeyor: https://github.com/wirebirdlabs/featherweight-nanomsg/commit/28f6c213f903d94ceb5e7754fb248a78722b6fd9#diff-180360612c6b8c4ed830919bbb4dd459R26

wirebirdlabs:
Merge yet?

As a reminder this notification still exists in the documentation `THIS IS AN EXPERIMENTAL FEATURE. DO NOT USE. THE FUNCTIONALITY IS SUBJECT TO CHANGE WITHOUT PRIOR NOTICE.`: https://github.com/nanomsg/nanomsg/blob/master/doc/nn_ws.txt#L19-L20

Removing that notice will be the cherry on top of this PR.

gdamore:
I confess I'd really like to break out the autobahn tests into a separate PR.  That's the main hesitancy here. I've not gone through it all yet.

wirebirdlabs:
see #578

wirebirdlabs:
> I'd really like to break out the autobahn tests into a separate PR

Understandable -- once #578 is mainlined, I'll clean up this PR to be only Autobahn TestSuite, fully squashed into a single commit.

That said -- what are options for mainlining this "big" dependency that optimally 1) helps the library by being run yet 2) remains transparent in order to not burden the rest of the dev process?

Like the Documentation and also Diagrams in this repository, which are opt-in based on whether you have the additional dependencies already installed, so is this test.

What about informing Travis CI and AppVeyor to only run this WebSocket test on a certain branch? Or only on PR's? Or only when tagging to create a release? (or some combination thereof)

gdamore:
So I noticed that in #578 you still didn’t quite achieve the 1 bug = 1
commit.  I’m a bit unhappy with that — as part of good hygiene, we should
be able to point at a single commit when determining whether a bug is
“fixed” or not.  Its nicer when we don’t have multiple issues in the same
commit, but frankly bundling multiple issues into a single commit is less
bad than spreading a single issue over multiple commits.

Maybe you should file separate issues for your test suites?  I’ve not
reviewed the separation of the commits, so I’m not sure whether that’s
better, or if its better to squash them.  Each commit needs to be
“well-formed” and “complete” though, so that any point can be used for a
bisect meaningfully.

On Tue, Dec 15, 2015 at 8:51 AM, Wirebird Labs LLC <notifications@github.com

wirebirdlabs:
> a bit unhappy

Can you live with it? The opportunity cost here is momentum. I generally agree with the tracking policy of commits and bugs, but recent development in these commits is more of a continuation of initial development that stagnated, rather than a patch for a defect or upgrade to a stable feature (where that sort of traceability is more valuable).

gdamore:
Well, let me put it this way.  I want to get these changes, and I’d like
them to reflect the actual changes.  Let me give you my thoughts in more
depth:

first commit is a clear bug of its own: synopsis should be “allow for
different test-specific timeouts”
second commit: (which Imo should be added *after* the fix it tests for),
should be “Provide async nn_close test for websocket”
third commit: “fix occasional deadling during nn_shutdown in websocket”
fourth commit: can stand on its, but could be squashed into the third
commit — except it seems nothing actually uses it?  Maybe we should reject
this change until a consumer exists for the change?  Oh wait, the next
commit includes a test for it:
fifth commit: that’s the main websocket fix
sixth commit: squash together with the fifth

So I’d squash commits 4, 5, and 6, and have the others stand apart.

I can go ahead and do this myself, except you’ve indicated that this would
cause you grief…

wirebirdlabs:
In the spirit of forward progress, I won't argue fix-before-test order, or the value of squashing 4, but feel free to cherry pick and change commit messages up thru the third (or fourth) -- i'll be glad to clean up squashing 5+6 afterward in a new PR if it helps.

JackDunaway:
Closing this PR without merging since it is superseded by #578; we'll resume conversation about Autobahn integration at a later date.

PR #578, opened 2015-12-14
WebSocket improvements (same as #569, but without Autobahn TestSuite)

JackDunaway:


gdamore:
Thanks. I will review and merge (assuming no issues found) shortly.

CMakeLists.txt
```diff
@@ -55,59 +55,60 @@ target_link_libraries (nanocat nanomsg)
...
+add_libnanomsg_test (timeo 5)
+add_libnanomsg_test (iovec 5)
+add_libnanomsg_test (msg 5)
+add_libnanomsg_test (prio 5)
+add_libnanomsg_test (poll 5)
+add_libnanomsg_test (device 5)
+add_libnanomsg_test (device4 5)
+add_libnanomsg_test (device5 5)
+add_libnanomsg_test (device6 5)
+add_libnanomsg_test (device7 120)
```
gdamore: This should be smaller. Lots smaller.

CMakeLists.txt
```diff
@@ -55,59 +55,60 @@ target_link_libraries (nanocat nanomsg)
...
-add_libnanomsg_test (tcp_shutdown)
-add_libnanomsg_test (ws)
-add_libnanomsg_test (tcpmux)
+add_libnanomsg_test (inproc 5)
+add_libnanomsg_test (inproc_shutdown 5)
+add_libnanomsg_test (ipc 5)
+add_libnanomsg_test (ipc_shutdown 30)
+add_libnanomsg_test (ipc_stress 5)
+add_libnanomsg_test (tcp 5)
+add_libnanomsg_test (tcp_shutdown 120)
```
gdamore: Why so long?

src/transports/ws/sws.c
```diff
@@ -679,7 +686,75 @@ static void nn_sws_validate_utf8_chunk (struct nn_sws *self)
...
+        close_code == NN_SWS_CLOSE_ERR_WUT ||
+        close_code == NN_SWS_CLOSE_ERR_INVALID_FRAME ||
+        close_code == NN_SWS_CLOSE_ERR_POLICY ||
+        close_code == NN_SWS_CLOSE_ERR_TOOBIG ||
+        close_code == NN_SWS_CLOSE_ERR_EXTENSION ||
+        close_code == NN_SWS_CLOSE_ERR_SERVER ||
+        (close_code >= 3000 && close_code <= 3999) ||
+        (close_code >= 4000 && close_code <= 4999)) {
+        /*  Repeat close code, per RFC 6455 7.4.1 and 7.4.2 */
+        nn_sws_fail_conn (self, (int) close_code,
```
gdamore: Not a fan of the cute message.  App may not want to advertise nanomsg consumption. And there can be security reasons to avoid self finger printing.

CMakeLists.txt
wirebirdlabs: It's a moving target how/why this test failed, but I observed that sometimes it would *pass* in the 60-120sec range. It truly seems like some sort of resource contention based on limited AppVeyor systems, most likely due to the single-core.

Basically, what we have is a test that really SLAMS the system, that works well with multiple cores, but seems to take a long, long time on single cores. Maybe the resolution here is to back off on the severity of the test. (??)

CMakeLists.txt
wirebirdlabs: Agree.

src/transports/ws/sws.c
wirebirdlabs: agree; squashed into a new PR, or commit on top of this one? your choice

CMakeLists.txt
wirebirdlabs: The other option here is to detect in the script whether an environment variable is set to APPVEYOR or similar (this is definitely possible) and give it more time. On an "average" desktop, I think we can drop this all the way to 10sec or even 5sec

wirebirdlabs: Final note: whether or not this test passes seems correlated to the time of day and day of week the test is run. Anecdotally, it's more likely to pass during what are assumed to be "off-peak" hours for AppVeyor workload, and pretty much fails 100% of the time during "busy" hours.

gdamore: Do we seriously think this is that load dependent?  I'm doubtful. 

I'd rather fail more aggressively. I think we still don't understand the reason for the failures.  Perhaps we are spinning when we should not?

src/transports/ws/sws.c
gdamore: Add to this.  I'm gonna squash them all into one anyway. Perhaps against multiple issues.

CMakeLists.txt
gdamore: That correlation note is very interesting indeed.  Maybe we are spinning or the test needs to be changed to be less sensitive to load?

wirebirdlabs: > Do we seriously think this is that load dependent? I'm doubtful.

For clarity, I'm not correlating with the load of the *machine*, but rather the load of the AppVeyor *cluster*, which might conditionally provision more/less resources (this is conjecture). It's hard to draw any conclusions though -- even just 25days ago, the AppVeyor documentation went from 2cores to 1core: https://github.com/appveyor/website/commit/349b45d307045ecfed57382c9606895972f206a8

I can repro on a Windows development VM when configuring that VM to launch with just 1 core. I have not been able to repro on a dev machine with 2+ cores.

Do you have a dev environment in a VM you could configure to just 1 core to see if it's just a Windows problem? (I have OS X on metal, but nothing handy as a VM that could be throttled)

src/transports/ws/sws.c
```diff
@@ -679,7 +686,74 @@ static void nn_sws_validate_utf8_chunk (struct nn_sws *self)
...
+        close_code == NN_SWS_CLOSE_ERR_WUT ||
+        close_code == NN_SWS_CLOSE_ERR_INVALID_FRAME ||
+        close_code == NN_SWS_CLOSE_ERR_POLICY ||
+        close_code == NN_SWS_CLOSE_ERR_TOOBIG ||
+        close_code == NN_SWS_CLOSE_ERR_EXTENSION ||
+        close_code == NN_SWS_CLOSE_ERR_SERVER ||
+        (close_code >= 3000 && close_code <= 3999) ||
+        (close_code >= 4000 && close_code <= 4999)) {
+        /*  Repeat close code, per RFC 6455 7.4.1 and 7.4.2 */
+        nn_sws_fail_conn (self, (int) close_code, "");
```
wirebirdlabs: Reason is now empty. BTW -- it's fine to squash this final commit 7607653e3980b7736b145cf4abe9e4f969c61730 with 28b08fd9879edc81e40e21e9c36b16038614e207, but any further squashing 1) removes history and ability for others to cherry pick into branches and 2) makes it unecessarily hard as a contributor to then reconcile my own branches. The commits as exist make sense, where the unnecessary noise has been squashed already (compared to the first PR, which was simply used as a sandbox for its CI)

gdamore:
Any commits you don't want squashed need to have issues clearly associated with them.  I don't like to have a given issue associated with more than a single commit.

wirebirdlabs:
In order of the 6 commits on this PR: #579, #505, #505, #580, #581, #581.

The two 505's are separate commits to 1) provide test then 2) provide fix, in order that other may independently verify the bug from the fix (testing alternate fixes, etc), so there's a 2:1 commit/issue ratio here.

The two #581's could be squashed, but that would make fast-forwarding hard (for me) and would erase commit log history from the code review.

gdamore:
I am seeing failures in the ws.c test:

garrett@Triton{233}% ./tests/ws
Assertion failed: rc < 0 (../tests/ws.c:225)
Abort

Am investigating now.  (It looks like this may be specific to macos X, I'm not sure.)

gdamore:
It passes sometimes... not others.

gdamore:
I think maybe it depends on speed... if the peer takes longer to close the connection, then the async send may actually pass.  Need to inject a slight delay, I think.

gdamore:
Well, 100 msec isn't enough...

gdamore:
Right, the test code is broken.  The problem is that you're trying to capture an error on send, and that isn't going to pass.  Due to the async nature, we can pass only if the send occurs fast enough to pick up the case where we've marked the state as disconnecting, but before the state has actually disconnected.

Send errors are not reported reliably for a variety of reasons, and code really shouldn't depend on them.  However, in our case I think we can improve this by ensuring that the queue size on send is zero.

gdamore:
Oh wait, queue size is a mangos thing, not nanomsg.  Argh.

gdamore:
Well, crap, I'm not ready to merge this because the test/ws fails, and its going to take more effort to understand the nature of the failure.  It looks like the code that attempts to test error codes for send is broken/fragile, because it can depend on the status of the state machines.

Please run manual tests if you can.  As I said, this reproduces not every time, but one in 5 or so runs.

I think if we try to send before the disconnect occurs, it causes the problem, but I've not found that timing fixes this.

The other thing is, are we sure that the disconnects are actually occurring?

gdamore:
So rather than sending me a single big PR, lets take these one item at a time; yes its more work, but at least that way we can make progress.

wirebirdlabs:
I'll try to repro on OS X -- I've not noticed it on the Travis or AppVeyor environments. Any specifics on your OS X environment that shows the failure?

gdamore:
I’m running MacOS 10.11.2.  Otherwise, pretty stock.

On Wed, Dec 16, 2015 at 5:36 PM, Wirebird Labs LLC <notifications@github.com

wirebirdlabs:
10.11.2 here too ... I can't get it to repro after many runs (ran 10+ times now):

```
============================================================================
Testsuite summary for nanomsg 0.8-beta-18-g7607653
============================================================================
# TOTAL: 40
# PASS:  40
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
============================================================================
Jack-MacBookAir13-Mid2013:featherweight-nanomsg jack$ git rev-parse HEAD
7607653e3980b7736b145cf4abe9e4f969c61730
```

gdamore:
I see you’re on an MB Air.  I’m using an high end iMac 27.  I wonder if the
issue is that I have extra oomph available.  Since it seems timing related,
I think that might be plausible.

On Wed, Dec 16, 2015 at 6:09 PM, Wirebird Labs LLC <notifications@github.com

wirebirdlabs:
> So rather than sending me a single big PR, lets take these one item at a time; yes its more work, but at least that way we can make progress

I hear you ... but it's just too tedious trying to get this stuff in. There may still be loose ends, but these PR's absolutely tie up known issues -- and the new failure you're able to show may be new, it may be an existing issue. There's a long history of AppVeyor and Travis CI passes for all the work in this PR -- I'd strongly recommend to just get it in, and start sourcing out testing to others besides yourself.

Again, if for no other reason -- this library is still <1.0, and there's still a huge warning not to even use the WebSocket transport on the public website: http://nanomsg.org/v0.8/nn_ws.7.html

Once these commits are mainlined, it's arguable that WebSocket transport in this library is on par with the TCP transport.

I'll look into updating the Travis CI spec to include OS X in the matrix.

gdamore:
I get maybe one error in 5.

wirebirdlabs:
> I’m using an high end iMac 27

I have one of those too, but never saw it when testing there either (during development).

wirebirdlabs:
Got the error, finally!

```
Assertion failed: rc < 0 (tests/ws.c:225)
```

That particular part of the WebSocket test does more checks than the TCP test.

TCP: https://github.com/nanomsg/nanomsg/blob/master/tests/tcp.c#L183-L199

WebSocket: https://github.com/wirebirdlabs/featherweight-nanomsg/blob/pr-ws-mega-squash-3/tests/ws.c#L199-L233

Here's the analysis -- 1) either the test is valid, and we might expect TCP to fail in the same manner if it were tested in this manner, or 2) the test is not valid, and is just subject to some inherent race condition, and the test should be improved.

Either way ... can we still move forward with merging? Even if that part of the test is broken, it's worthwhile to see it deleted in the commit history with rationale why it was wrong (rather than squashed)

gdamore:
Lets just remove that part of the test.  I don’t think it matters.
Reliance on sendmsg() error codes is just fragile testing.  Can you fix
that, so that I can move ahead?

Alternatively, I can cherry pick parts of your PR until we get to the
failing one.

I apologize for my caution here, I just don’t want to regress.  Things in
this code path are so complex, that accepting changes that I don’t feel
really confident in is probably not a great idea.  We can get into a very
bad state by letting too many of those in, so I prefer to be very defensive.

On Wed, Dec 16, 2015 at 6:34 PM, Wirebird Labs LLC <notifications@github.com

gdamore:
NB. The conflict is as a result of cherry picking these commits.  Stay tuned.

JackDunaway:
Closing this PR for now; it will be replaced in the near future with commits that merge cleanly.

JackDunaway:
Related: this branch (and others) will be pruned from the https://github.com/JackDunaway/featherweight-nanomsg repository, meaning some links above in GitHub may not work anymore.

gdamore:
Ok, I want to help with the cmake work too — so as soon as you have a PR
ready for testing, please let me know.

PR #586, opened 2016-01-10
Added missing errno value to nn_close(3) man page.

Snaipe:
`nn_close` seems to set errno to ETERM if the library has not yet been initialized, or if no open sockets are remaining.

I am unsure whether the function should set errno to EBADF instead (since at both of these points there is no way any integer could be a valid socket) or if the current behaviour should be documented.

In doubt, this PR fixes the man page for `nn_close(3)` to mention ETERM in the list of possible errno values.


gdamore:
Yes, ETERM is right here.  I'm going to go ahead integrate this.

gdamore:
The code is now changed, so it only returns EBADF.

PR #587, opened 2016-01-10
fixes #582 Added optional at-fork strategies

### LINKED ISSUE #582: Add an optional "at-fork" strategy
I was talking about this on the gitter channel with @gdamore, so I thought I'd write up my thoughts on the matter here.

(Sorry for the long wall of text, see the end for a tl;dr)
#### The context

I am maintaining an unit testing framework for C. One of its feature is that it runs each test in its separate address space, to make sure that in the not-so-unlikely event that the user code crashes, the test runner will be still alive and well.

Currently, I am gathering data from the workers through a pipe I set up. This was simple enough until now, although I had to make a portable abstraction layer over `pipe(2)` and `CreatePipe` to have it run on both posix & windows systems -- hence, the current structure of the runner is to create the pipe, then fork the workers, read data from the pipe, and fork new workers as old ones die.

While this was sufficient, it was a bit of a hassle to maintain and very limited for new purposes like having external workers (e.g. in an embedded device), along with a lot of other problems; this is why I considered using nanomsg for the rewrite of the IO layer.
#### The problem

Reading through the sources of nanomsg, I noticed that there was heavy usage of non-blocking I/O, and hence threads. This is usually a problem for programs using fork(), as the two are orthogonal to each other, so I went to see if there was a `pthread_atfork` handle.

It turns out there isn't (I don't blame you, this is an issue that cannot be trivially solved), which brings up the main topic of this issue: what should be done on the side of nanomsg if a process decides to fork() ?

I agree that there cannot be an "expected" behaviour and that it depends on the use case, which is
why I think nanomsg could have a function to set an atfork strategy before the first socket is ever opened.
#### The current stategy

Currently, this is what happens when forking:
1. The parent calls fork()
2. The child process spawns, all (possibly locked) mutexes and file descriptors are inherited from the parent.
3. The parent resumes.

This means that nanomsg is effectively in an unusable state in the child process after the fork.

Thankfully, nanomsg also sets `FD_CLOEXEC` on its sockets, so for the common fork()-then-exec() use case it's possible to reset the library to an usable state.
#### A possible new strategy

In my case, I don't call exec() after fork, because I still want to work with a clone of the current process. However, it turns out that I don't really need any of the sockets from the parent in the child process: I could easily connect in the child to the server socket in the parent. I would just need nanomsg to be in a "clean" state, as if no sockets existed.

Effectively, this would be equivalent to manually do what happens on exec(), but earlier: release all the locks and close all of nanomsg's file descriptors.

A naive implementation (in pseudo-code, as I am not familiar enough with the internals of nanomsg) could be:

``` c
void nn_prefork(void) {
    // for each worker thread
    for (struct nn_worker *w = ...; ...; ...) {
        // Lock its mutex
        nn_mutex_lock (&w->sync);
    }

    // lock the global lock
    nn_glock_lock ();
}

void nn_postfork_parent(void) {
    // unlock the global lock
    nn_glock_unlock ();

    // for each worker thread
    for (struct nn_worker *w = ...; ...; ...) {
        // Unlock its mutex
        nn_mutex_unlock (&w->sync);
    }
}

void nn_postfork_child(void) {
    // unlock the global lock
    nn_glock_unlock ();

    // for each worker thread
    for (struct nn_worker *w = ...; ...; ...) {
        // Unlock its mutex
        nn_mutex_unlock (&w->sync);

        // Clean up worker resources
        nn_timerset_term (&w->timerset);
        nn_poller_term (&w->poller);
        nn_efd_term (&w->efd);
        nn_queue_item_term (&w->stop);
        nn_queue_term (&w->tasks);
        nn_mutex_term (&w->sync);
    }

    // for each opened socket
    for (int *sock = ...; ...; ...) {
        nn_close(*sock);
    }
}

enum nn_option {
    NN_FORK_STRATEGY,
};

enum nn_fork_strategy {
    NN_FORK_NONE,
    NN_FORK_RESET
};

int nn_setopt (int option, const void *optval, size_t optvallen) {
    ...
    if (option == NN_FORK_STRATEGY && ((const int *)optval) == NN_FORK_RESET) {
        pthread_atfork(nn_prefork, nn_postfork_parent, nn_postfork_child);
    }
    ...
}
```

---

Maybe another "cleaner" possibility:

``` c
void nn_prefork(void) {
    // for each worker thread
    for (struct nn_worker *w = ...; ...; ...) {
        // Ask worker to suspend itself
        nn_mutex_lock (&w->sync);
        nn_queue_push (&w->tasks, &w->suspend);
        nn_efd_signal (&w->efd);
        nn_mutex_unlock (&w->sync);

        // Wait until the worker is suspended
        nn_wait_thread_suspended (&self->thread); // TODO

        // Reacquire the lock
        nn_mutex_lock (&w->sync);
    }
}

void nn_postfork_parent(void) {
    // for each worker thread
    for (struct nn_worker *w = ...; ...; ...) {
        // Wake worker up
        nn_thread_wakeup (w) // TODO

        // Unlock its mutex
        nn_mutex_unlock (&w->sync);
    }

    // lock the global lock
    nn_glock_lock ();
}

// Same nn_postfork_child and nn_setopt.
```

And modify the current posix worker to lock the sync mutex and wait on a `pthread_cond_t` or something like this.

Since waiting on a `pthread_cond_t` unlocks its associated mutex, one can lock again the mutex, fork, and in the parent tell the thread to wake up (which will cause the waiting on the `pthread_cond_t` to try and lock the mutex), then unlock the mutex to allow the worker to resume.
#### TL;DR
- There is no current atfork strategy, which means that it's impossible to use nanomsg in a child process after fork()ing
- The sockets are set to close on exec(), so the only way to reset the library to an usable state is by calling exec() after fork() or vfork().
- Rather than trying to solve the problem, the library could provide functions in its API to let the user set the atfork strategy, based on his use case.
- One of the possible strategies could be to do the whole library reset -- that is normally done by the OS on exec() -- right after fork().

Snaipe:
This addresses issue #582.

This PR is twofold:
1. It first adds the `nn_setopt(3)` function in the API, to set global nanomsg options that are not bound to a specific socket.
2. It adds at-fork handling to the library, as well as two strategies: `NN_FORK_NONE` (do nothing, current behaviour) and `NN_FORK_RESET` (terminate and cleanup completely nanomsg in the child process).

This is still a work in progress -- I am opening this for reviews, and most notably: 
- I want to know if adding `nn_setopt` is a good idea. I added it because I felt like adding a singular function like `nn_set_atfork_strategy` wouldn't follow the current design of the library, and there might be other "global" options in the future, but not everyone might agree with me.
- I provided the `NN_FORK_RESET` strategy as an option, rather than the default. It might have been more sensible to have it as the default strategy, seeing that the library simply fails horribly if it's used after fork with `NN_FORK_NONE`, but I thought that keeping the current behaviour would be less disrupting for existing programs that use `fork(2)` then `exec(2)`.
- I did the ground work for multiple fork strategies in the event that more than the two commited here would be added, but it might also not be justified.

If adding `nn_setopt` is OK, I will add a man page to cover it in `./doc`.

Last but not least, I added a test at `tests/fork.c` to test the behaviour of NN_FORK_RESET.

### DIFF
--- a/src/core/global.c
+++ b/src/core/global.c
@@ -1258,7 +1258,8 @@ int nn_global_postfork_cleanup ()
         for (i = 0; i != NN_MAX_SOCKETS; ++i)
             if (self.socks [i]) {
                 nn_sock_stop (self.socks [i]);
-                nn_sock_rele (self.socks [i]);
+                while (self.socks [i]->holds > 0)
+                    nn_sock_rele (self.socks [i]);
                 nn_sock_term (self.socks [i]);
                 nn_free (self.socks [i]);
                 self.socks [i] = NULL;
--- a/src/core/global.c
+++ b/src/core/global.c
@@ -1190,21 +1190,49 @@ int nn_setopt (int option, const void *optval, size_t optvallen)
     return -1;
 }
 
+void nn_global_lock_all_sockets (void)
+{
+    int i;
+    nn_assert(nn_is_glock_held());
+
+    nn_ctx_enter (&self.ctx);
+    if (self.socks && self.nsocks) {
+        for (i = 0; i != NN_MAX_SOCKETS; ++i)
+            if (self.socks [i])
+                nn_ctx_enter (nn_sock_getctx (self.socks [i]));
+    }
+}
+
+void nn_global_unlock_all_sockets (void)
+{
+    int i;
+    nn_assert(nn_is_glock_held());
+
+    if (self.socks && self.nsocks) {
+        for (i = 0; i != NN_MAX_SOCKETS; ++i)
+            if (self.socks [i])
+                nn_ctx_leave (nn_sock_getctx (self.socks [i]));
+    }
+    nn_ctx_leave (&self.ctx);
+}
+
 int nn_global_postfork_cleanup ()
 {
     int i;
+    nn_assert(nn_is_glock_held());
 
-    nn_glock_lock();
     self.forked = 1;
 
     if (self.socks && self.nsocks) {
         for (i = 0; i != NN_MAX_SOCKETS; ++i)
             if (self.socks [i]) {
+                nn_sock_stop (self.socks [i]);
+                nn_sock_rele (self.socks [i]);
+                nn_sock_term (self.socks [i]);
+                nn_free (self.socks [i]);
                 self.socks [i] = NULL;
                 self.nsocks--;
             }
     }
     nn_global_term();
-
-    nn_glock_unlock();
 }
--- a/src/core/global.c
+++ b/src/core/global.c
@@ -2,6 +2,7 @@
     Copyright (c) 2012-2014 Martin Sustrik  All rights reserved.
     Copyright (c) 2013 GoPivotal, Inc.  All rights reserved.
     Copyright 2016 Garrett D'Amore <garrett@damore.org>
+    Copyright (c) 2016 Franklin "Snaipe" Mathieu <franklinmathieu@gmail.com>
 
     Permission is hereby granted, free of charge, to any person obtaining a copy
     of this software and associated documentation files (the "Software"),
@@ -43,6 +44,7 @@
 #include "../utils/chunk.h"
 #include "../utils/msg.h"
 #include "../utils/attr.h"
+#include "../utils/fork.h"
 
 #include "../transports/inproc/inproc.h"
 #include "../transports/ipc/ipc.h"
@@ -132,6 +134,9 @@ struct nn_global {
     int state;
 
     int print_errors;
+
+    /* Forked status */
+    int forked;
 };
 
 /*  Singleton object containing the global state of the library. */
@@ -181,6 +186,9 @@ static void nn_global_init (void)
     /*  Check whether the library was already initialised. If so, do nothing. */
     if (self.socks)
         return;
+
+    /* Reset forked status */
+    self.forked = 0;
 
     /*  On Windows, initialise the socket library. */
 #if defined NN_HAVE_WINDOWS
@@ -248,6 +256,11 @@ static void nn_global_init (void)
     nn_global_add_socktype (nn_bus_socktype);
     nn_global_add_socktype (nn_xbus_socktype);
 
+#ifndef NN_HAVE_WINDOWS
+    /* Register atfork handlers */
+    errno_assert(nn_setup_atfork_handlers () == 0);
+#endif
+
     /*  Start the worker threads. */
     nn_pool_init (&self.pool);
 }
@@ -266,7 +279,8 @@ static void nn_global_term (void)
         return;
 
     /*  Shut down the worker threads. */
-    nn_pool_term (&self.pool);
+    if (!self.forked)
+        nn_pool_term (&self.pool);
 
     /*  Ask all the transport to deallocate their global resources. */
     while (!nn_list_empty (&self.transports)) {
@@ -1160,9 +1174,37 @@ void nn_global_rele_socket(struct nn_sock *sock)
 int nn_setopt (int option, const void *optval, size_t optvallen)
 {
     switch (option) {
+    case NN_FORK_STRATEGY:
+        nn_assert(optvallen == sizeof (int));
+        int idx = *(const int *) optval;
+        if (idx < 0 || idx >= NN_FORK_MAX_) {
+            errno = EINVAL;
+            break;
+        }
+        nn_fork_strategy = &nn_fork_strategies[idx];
+        return 0;
     default:
         errno = ENOPROTOOPT;
         break;
     }
     return -1;
+}
+
+int nn_global_postfork_cleanup ()
+{
+    int i;
+
+    nn_glock_lock();
+    self.forked = 1;
+
+    if (self.socks && self.nsocks) {
+        for (i = 0; i != NN_MAX_SOCKETS; ++i)
+            if (self.socks [i]) {
+                self.socks [i] = NULL;
+                self.nsocks--;
+            }
+    }
+    nn_global_term();
+
+    nn_glock_unlock();
 }
--- a/src/core/global.c
+++ b/src/core/global.c
@@ -1186,25 +1186,6 @@ void nn_global_rele_socket(struct nn_sock *sock)
     nn_glock_unlock();
 }
 
-int nn_setopt (int option, const void *optval, size_t optvallen)
-{
-    switch (option) {
-    case NN_FORK_STRATEGY:
-        nn_assert(optvallen == sizeof (int));
-        int idx = *(const int *) optval;
-        if (idx < 0 || idx >= NN_FORK_MAX_) {
-            errno = EINVAL;
-            break;
-        }
-        nn_fork_strategy = &nn_fork_strategies[idx];
-        return 0;
-    default:
-        errno = ENOPROTOOPT;
-        break;
-    }
-    return -1;
-}
-
 void nn_global_lock_all_sockets (void)
 {
     int i;
--- a/src/core/global.c
+++ b/src/core/global.c
@@ -45,7 +45,7 @@
 #include "../utils/msg.h"
 #include "../utils/attr.h"
 #include "../utils/fork.h"
-#include "../utils/atomic.h"
+#include "../utils/cond.h"
 
 #include "../transports/inproc/inproc.h"
 #include "../transports/ipc/ipc.h"
@@ -137,9 +137,12 @@ struct nn_global {
     int print_errors;
 
 #ifndef NN_HAVE_WINDOWS
-    struct nn_mutex fork_sync;
     volatile int fork_lock;
-    struct nn_atomic io_operations;
+    struct nn_mutex fork_sync;
+    struct nn_cond fork_cond;
+    volatile uint32_t io_operations;
+    struct nn_mutex ioop_sync;
+    struct nn_cond ioop_cond;
 #endif
 };
 
@@ -166,10 +169,6 @@ static int nn_global_create_socket (int domain, int protocol);
 static int nn_global_hold_socket(struct nn_sock **sockp, int s);
 static int nn_global_hold_socket_locked(struct nn_sock **sockp, int s);
 static void nn_global_rele_socket(struct nn_sock *);
-
-/*  Pending fork locks  */
-static void nn_global_lock_if_forked (void);
-static void nn_global_unlock_if_forked (void);
 
 int nn_errno (void)
 {
@@ -265,7 +264,10 @@ static void nn_global_init (void)
     /* Register atfork handlers */
     errno_assert(nn_setup_atfork_handlers () == 0);
     nn_mutex_init (&self.fork_sync);
-    nn_atomic_init (&self.io_operations, 0);
+    nn_mutex_init (&self.ioop_sync);
+    nn_cond_init (&self.fork_cond);
+    nn_cond_init (&self.ioop_cond);
+    self.io_operations = 0;
     self.fork_lock = 0;
 #endif
 
@@ -606,8 +608,6 @@ int nn_bind (int s, const char *addr)
     int rc;
     struct nn_sock *sock;
 
-    nn_global_lock_if_forked ();
-
     rc = nn_global_hold_socket (&sock, s);
     if (rc < 0) {
         goto failure;
@@ -620,11 +620,9 @@ int nn_bind (int s, const char *addr)
     }
 
     nn_global_rele_socket (sock);
-    nn_global_unlock_if_forked ();
     return rc;
 
 failure:
-    nn_global_unlock_if_forked ();
     errno = -rc;
     return -1;
 }
@@ -633,8 +631,6 @@ int nn_connect (int s, const char *addr)
 {
     int rc;
     struct nn_sock *sock;
-
-    nn_global_lock_if_forked ();
 
     rc = nn_global_hold_socket (&sock, s);
     if (nn_slow (rc < 0)) {
@@ -648,11 +644,9 @@ int nn_connect (int s, const char *addr)
     }
 
     nn_global_rele_socket (sock);
-    nn_global_unlock_if_forked ();
     return rc;
 
 failure:
-    nn_global_unlock_if_forked ();
     errno = -rc;
     return -1;
 }
@@ -661,8 +655,6 @@ int nn_shutdown (int s, int how)
 {
     int rc;
     struct nn_sock *sock;
-
-    nn_global_lock_if_forked ();
 
     rc = nn_global_hold_socket (&sock, s);
     if (nn_slow (rc < 0)) {
@@ -677,11 +669,9 @@ int nn_shutdown (int s, int how)
     nn_assert (rc == 0);
 
     nn_global_rele_socket (sock);
-    nn_global_unlock_if_forked ();
     return 0;
 
 failure:
-    nn_global_unlock_if_forked ();
     errno = -rc;
     return -1;
 }
@@ -731,11 +721,8 @@ int nn_sendmsg (int s, const struct nn_msghdr *msghdr, int flags)
     struct nn_cmsghdr *cmsg;
     struct nn_sock *sock;
 
-    nn_global_lock_if_forked ();
-
     rc = nn_global_hold_socket (&sock, s);
     if (nn_slow (rc < 0)) {
-        nn_global_unlock_if_forked ();
         errno = -rc;
         return -1;
     }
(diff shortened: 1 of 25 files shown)


src/core/global.c
```diff
@@ -1465,3 +1483,69 @@ void nn_global_rele_socket(struct nn_sock *sock)
...
+            break;
+        }
+#ifndef NN_HAVE_WINDOWS
+        nn_fork_strategy = &nn_fork_strategies[idx];
+#endif
+        return 0;
+    default:
+        errno = ENOPROTOOPT;
+        break;
+    }
```
wirebirdlabs: The break/return strategy is a little mixed here. Perhaps introduce `int rc` and just use breaks, with a single return at the end of the function?

That said ... not sure if `setopt` is a good idea yet, compared to say a build-time flag configuration option... thoughts?

src/core/global.c
```diff
@@ -1465,3 +1483,69 @@ void nn_global_rele_socket(struct nn_sock *sock)
...
+int nn_global_lock_all_sockets (void)
+{
+    int i;
+
+    if (self.socks && self.nsocks) {
+        for (i = 0; i != NN_MAX_SOCKETS; ++i)
+            if (self.socks [i])
+                nn_ctx_enter (nn_sock_getctx (self.socks [i]));
+    }
+}
```
wirebirdlabs: This might be a multi-threading race; let's check into whether `nn_glock_lock()` needs to be held. Specifically -- in the `if` test, consider if the socket is destroyed in a parallel thread just after the test but before the next line it's guarding.

src/nn.h
```diff
@@ -342,6 +342,18 @@ NN_EXPORT  struct nn_cmsghdr *nn_cmsg_nxthdr_ (
...
+
+/*  Fork strategies.                                                          */
+/*  ABI warning: do not reorder, add new entries before NN_FORK_MAX_.         */
+enum nn_fork_strategy_kind {
+    NN_FORK_NONE = 1,
+    NN_FORK_RESET,
+
+    NN_FORK_MAX_
+};
+
```
wirebirdlabs: Going from memory here, but this might be the first use of an `enum` instead of a simple `int` in the nanomsg library here. (There exist merits for both design decisions, and it's worth considering the design decisions and precedents)

src/nn.h
```diff
@@ -352,6 +364,7 @@ NN_EXPORT int nn_setsockopt (int s, int level, int option, const void *optval,
     size_t optvallen);
 NN_EXPORT int nn_getsockopt (int s, int level, int option, void *optval,
     size_t *optvallen);
+NN_EXPORT int nn_setopt (int option, const void *optval, size_t optvallen);
```
wirebirdlabs: Again, not sure if a run-time function is better than compile-time configuration option ... do there exist use cases where it even makes sense to set at run-time?

src/core/global.c
Snaipe: In the current usage, the global lock is already held -- see [src/utils/fork.c:42](https://github.com/nanomsg/nanomsg/pull/587/files#diff-394e5ecf05f93b4c8488b977beb6bed6R42), but I agree that it's not immediately apparent. Should I add a word of warning that the global lock needs to be held, or should I make it so this function explicitely calls `nn_glock_lock()` then `nn_glock_unlock()`?

src/core/global.c
Snaipe: I was thinking about a build time flag, but I'm not sure about how this should be handled for packaged distributions. Do we assume that everyone wants `NN_FORK_RESET` as the "best" fork strategy?

src/nn.h
Snaipe: One *could* want to set the strategy to NN_FORK_RESET, then reset it to NN_FORK_NONE in the child after fork, notably if the child itself forks then execs as the runtime cost of resetting the library could be non-negligible.

src/core/global.c
wirebirdlabs: > the global lock is already held

Indeed.

> make it so this function explicitely calls `nn_glock_lock()` then `nn_glock_unlock()`

Perhaps not, if the semantic required here is to remain locked until the symmetric work is done also, or if there is additional initial work that needs to be done along with the call to this. That said; your mind is better wrapped around what is required right now; consider my comments at this point "drive by" considerations.

> word of warning that the global lock needs to be held

Comments alone are weak; thinking aloud, it would be perfect if there could be created a macro `nn_assert_global_lock_held();` to capture this thought in code. That could be used a few different places, probably (notably ... pretty much anywhere the global lock is held, but *especially* in locations where the `unlock` does not immediately follow the `lock` in the same code block/scope)

tests/fork.c
```diff
@@ -0,0 +1,93 @@
...
+    test_recv (sb_alt, "0123456789012345678901234567890123456789");
+    test_send (sb_alt, "0123456789012345678901234567890123456789");
+
+    /* Fork with a pair on each end */
+
+    sb = test_socket (AF_SP, NN_REQ);
+    test_bind (sb, SOCKET_ADDRESS);
+
+    pid_t pid = fork();
+    assert(pid != -1);
```
wirebirdlabs: Minor nit -- use `nn_assert()` rather than `assert()

tests/fork.c
```diff
@@ -0,0 +1,93 @@
...
+    test_send (sc_alt, "0123456789012345678901234567890123456789");
+    test_recv (sb_alt, "0123456789012345678901234567890123456789");
+    test_send (sb_alt, "0123456789012345678901234567890123456789");
+
+    /* Fork with a pair on each end */
+
+    sb = test_socket (AF_SP, NN_REQ);
+    test_bind (sb, SOCKET_ADDRESS);
+
+    pid_t pid = fork();
```
wirebirdlabs: (Looks like this test will break on Windows compiling with MSVC)

src/nn.h
wirebirdlabs: This does not feel like a common use-case, which would make me lean toward the compile-time option. The cost of introducing this as a run-time option is increasing the surface area of the libnanomsg API, without a substantial ROI on the benefit.

`nn_setopt` might be revisited if there existed more run-time options that added even more value to the existence of `nn_setopt`, but for now doesn't seem like the best option.

It is of course intentionally limiting to set this at compile time -- but trading flexibility for usability *seems* more desirable here.

src/core/global.c
wirebirdlabs: A packaged distribution connotes some level of curation of "best" strategy. What would you say is the best strategy?

tests/fork.c
Snaipe: I did not add it to the CMake tests for this reason, as it is really only relevant to run it on posix. Should I put some guards on the file checking against NN_HAVE_WINDOWS to short circuit the logic?

src/core/global.c
Snaipe: There really is no best strategy I'm afraid.

The current behaviour (aka `NN_FORK_NONE`) is fast and useful for children that call exec() after fork, while `NN_FORK_RESET` forces a reset of the library, can sometimes be fairly slow (and might even be slower if nanomsg implements multiple workers in the poller), and should really be a last-resort solution.

I was thinking of introducing a 3rd strategy after the initial work was done, that would give up on all pending I/O events and tasks in the child, effectively leaving the current opened sockets in a valid, non-working state. Admitedly, this could be more versatile than the two strategies introduced by this commit, but I wouldn't qualify it "better" in its own sense.

src/nn.h
Snaipe: I did this to avoid incrementing the value of NN_FORK_MAX_ if another value was to be added, but I can use macro definitions instead.

I'm curious, though -- what is the rationale of not using enums originally? I was led to believe that integral constants should generally be represented as enums because of the type information they convey, but I certainly am not aware of everything that led to these design decisions.

src/core/sock.c
```diff
@@ -242,9 +243,51 @@ int nn_sock_term (struct nn_sock *self)
...
 
+    nn_sock_unsafe_cleanup(self, NN_CLEAN_DEFAULT);
+
+    return 0;
+}
+
+void nn_release_endpoint (struct nn_list_item *it)
+{
+    struct nn_ep *ep = nn_cont (it, struct nn_ep, item);
+    /* TODO: force cleanup the endpoint */
```
Snaipe: This needs a hook inside the endpoint to release all of its resources at once without caring about safely terminating.

src/core/sock.c
```diff
@@ -242,9 +243,51 @@ int nn_sock_term (struct nn_sock *self)
...
+}
+
+void nn_sock_unsafe_cleanup (struct nn_sock *self, enum nn_cleanup_opt opts)
+{
+    int i;
+
+    /*  Cleanup. */
+    if (opts & NN_CLEAN_EMPTY) {
+        /* TODO: Force destroy the socket base */
+        /* self->sockbase->vfptr->destroy (self->sockbase); */
```
Snaipe: Same as the endpoints, this needs a hook inside the socket base to release everything. The destroy function simply fails because it checks that the state is valid, which cannot happen during a fork, before terminating.

src/nn.h
Snaipe: Right. I'm going to revert 4331b77 and introduce a configure option to set the fork strategy, but what should I make the default? The current behaviour?

Snaipe:
I'm having troubles fighting the library where *all* (or almost all) termination functions expect the state to be valid.

This makes it hard to force nanomsg to simply "give up" when there is nothing that could be done to restore the state to a valid one.

This seems unlikely to be done right now, but maybe the validity checks should be decoupled from the actual destructors -- this way, upon normal termination the library code would call those validity checks, while leaving an internal way to give up and destroy everything without aborting the user process.

Snaipe:
Another issue I noticed comes from the behaviour of the file descriptors returned by `kqueue(2)`: those are [not inheritable in children](https://www.freebsd.org/cgi/man.cgi?query=kqueue&sektion=2), meaning that `close(2)`'ing them is an error.

In a nutshell, this makes all forks immediately die because during the force-cleanup process, a file descriptor returned by kqueue gets closed, `close` returns -1, and nanomsg aborts because errno is `EBADF`.

I am hesitating if I should yet again duplicate the cleanup code, even as a short-term solution.

Snaipe:
To address the problem mentioned in my two above commands, I added a cleanup parameter to convey the intent of the destruction function. This has led to a lot of changes on the codebase, and while current behaviour is maintained with no fork strategies, there could be a few failure points that may hit users of the reset fork strategy on some protocols/transports.

I am going to add a more complete test for fork, over multiple transports and protocols.

The second thing that should raise the confidence that everything works as expected would be to add OS X builds to the travis configuration (this should speed up the development process as well for future PRs).

gdamore:
Historically, MacOS X was not part of the free Travis plan.

Snaipe:
Seeing that there were almost 1400 changed lines over almost all sources, I decided to try another approach for this fix: Instead of immediately clean up with forceful/unsafe variants of `*_term` functions, I restore the library to a valid state by resurrecting the worker and the poller and locking the library in a more stable state before forking, before simply calling `nn_sock_stop` and `nn_sock_term` on all sockets.

This approach is much less intrusive on the existing codebase and has the distinct advantage of using the "normal" cleanup provided by the library rather than trying to hack through all destructors because all of them expect a valid state.

This works quite well on my end, tested with a heavy workload on a ipc+reqrep setup. I expect it to work everywhere since this approach is much more general than the old one, but I'll throw in some more tests to make sure.

gdamore:
This sounds like a very same approach.

Snaipe:
@gdamore I'd say it roughly has the same effect. The new approach simply restores nanomsg to an idling state before terminating the library, while the old approach tried to terminate it while it was in an invalid state.

wirebirdlabs:
I'll extrapolate here, that @gdamore meant to type "sane" rather than "same" from a mobile phone -- in other words, much closer to a compliment than a jab, in case that didn't come across @Snaipe :-)

Even if that's not the case, I'll go ahead and say it -- much more sane indeed.

gdamore:
Yes that's what I meant!!!  Now your reply makes much more sense to me.  Argh tiny phone keyboards.

reqshark:
heh, @wirebirdlabs i didnt even read "same" until you wrote that

Snaipe:
Ah, thanks for clarifying, I was a bit confused here. :)

gdamore:
Looking at this, I still have to review the code.  But I think we should consider a few things.

First off, I don't think we should ever support the notion of libnanomsg sockets "crossing" a fork boundary.  That is, I think all nn_sockets should be considered to be "illegal" in a post-fork situation -- automatically closed, with no access to the socket permitted in the child process.

Given that, I think a reasonable course of action would be to simply close() any file descriptors that were open in the child, and to toss any state, ignoring mutexes altogether.  Probably a single global variable can be used to assist the worker threads, that will get errnos if they were running, to notice that they are in a child and need to just bail.  (This should only be a concern on systems like Solaris or illumos, where forkall() or forkallx() or legacy fork() semantics are in play, and all threads are duplicated in child.  POSIX semantics only allow a single thread to be created in the child on fork.)

I have some ideas on how to set this up, and I think we should do this automatically on behalf of the application at nn_global_init() time.

The interesting bit here is dealing with allocations.  I'm not sure how to make fork not leak resources (memory) -- more consideration is needed.

Snaipe:
> First off, I don't think we should ever support the notion of libnanomsg sockets "crossing" a fork boundary. That is, I think all nn_sockets should be considered to be "illegal" in a post-fork situation -- automatically closed, with no access to the socket permitted in the child process.

This is the behaviour introduced by the PR. There is just no way to sanely resume socket work in the child after a fork, so we just bail by briefly reviving any worker threads to only process termination events and calling a synchronous version of `nn_term()`.

> Given that, I think a reasonable course of action would be to simply close() any file descriptors that were open in the child, and to toss any state, ignoring mutexes altogether. Probably a single global variable can be used to assist the worker threads, that will get errnos if they were running, to notice that they are in a child and need to just bail. (This should only be a concern on systems like Solaris or illumos, where forkall() or forkallx() or legacy fork() semantics are in play, and all threads are duplicated in child. POSIX semantics only allow a single thread to be created in the child on fork.)

The gist of what the PR does is:

1. During prefork, it tells the worker thread to go into a paused state. The worker then processes the work queue until it's empty, then waits on a condition variable. We also lock all global/socket/thread mutexes.
2. During postfork:

    1. In the parent, the condition variable is signaled and the worker resumes its work. All mutexes locked during prefork are unlocked.

    2. In the child, we destroy the mutex & condition variable the (now dead) worker thread was waiting on. Then, a "dumb" worker thread gets spawned in place of the dead one to process any postfork event before resetting, all mutexes locked during prefork are unlocked, and the library gets terminated.

> The interesting bit here is dealing with allocations. I'm not sure how to make fork not leak resources (memory) -- more consideration is needed.

Thus far, I'm sometimes having some issues with the way to deal with memory allocated by `nn_recv` -- the rest being gracefuly handled by the `nn_term` -- since it's very possible for a fork to happen while nn_recv blocks in another thread.

What first comes to mind to fix this would be to register the newly allocated message to a list somewhere, then remove it when the function is done with it. The child would then iterate through this list and free each block.

A cleaner solution would be to make all allocations in a memory pool, either by using something like [talloc](https://talloc.samba.org/talloc/doc/html/index.html) or implementing one ourselves. This would assert that no leaks would be possible during a fork, ever, since we will be able to simply destroy the pool and any nodes associated with it.

JackDunaway:
> Thus far, I'm sometimes having some issues with the way to deal with memory allocated by nn_recv -- the rest being gracefuly handled by the nn_term -- since it's very possible for a fork to happen while nn_recv blocks in another thread

Can you link line numbers to some of the more problematic allocations?

Snaipe:
@JackDunaway 

For `nn_recv`: https://github.com/nanomsg/nanomsg/blob/master/src/core/global.c#L953
For `nn_send`: https://github.com/nanomsg/nanomsg/blob/master/src/core/global.c#L809

Both allocate a chunkref that gets lost in limbo if another thread forks.

grodtron:
What is the status of this PR? Is it likely to be addressed at some point in the future?

gdamore:
Yes, I have a plan to address it.  I’m not sure the PR is ready as it
stands though.

Snaipe:
The PR works but is not ready for merging. There are a few points to address, mainly:

1. Make the whole thing simpler
2. Address potential leaks when using nn_recv with NN_MSG in a multithreaded context

gdamore:
I'm closing this -- the discussion was noted elsewhere - basically we cannot safely do this because pthread_atfork() does not permit the child side to use any async-signal-unsafe functions.  So no mutexes, no nothing.

I may revisit this to permit an intentional reinitialization (explicitly done in the child), or to panic() on reentrance by the child process.  If I do, I'll deal with that separately.

PR #590, opened 2016-01-15
Fixed 'A call of the 'memset' function will lead to overflow of the b…

stevenhoving:
…uffer '& self->sec_attr'.'


Snaipe:
This does not fix the intent of the original line, which was to zero-fill the security attributes.

A proper fix would be `memset (self->sec_attr, 0, sizeof (SECURITY_ATTRIBUTES));` (or with `sizeof (*self->sec_attr)`), and introducing a null check.

stevenhoving:
Yes, you are correct... I will prep a new pr tomorrow.

gdamore:
Integrated alternate fix in #635

PR #591, opened 2016-01-17
Added cmake options to disable builds of tools and tests

Snaipe:
This adds the `NN_TOOLS` and `NN_TESTS` cmake options to disable at build-time tests and tools.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,7 +31,6 @@ include (CheckLibraryExists)
 include (CheckCSourceCompiles)
 
 project (nanomsg C)
-enable_testing ()
 
 set (ISSUE_REPORT_MSG "Please consider opening an issue at https://github.com/nanomsg/nanomsg with your findings")
 
@@ -82,9 +81,11 @@ endif()
 
 # User-defined options.
 
-option (NN_ENABLE_NANOCAT "Enable building nanocat utility." ON)
 option (NN_STATIC_LIB "Build static library instead of shared library." OFF)
 option (NN_ENABLE_GETADDRINFO_A "Enable/disable use of getaddrinfo_a in place of getaddrinfo." ON)
+option (NN_TESTS "Build and run nanomsg tests" ON)
+option (NN_TOOLS "Build nanomsg tools" ON)
+option (NN_ENABLE_NANOCAT "Enable building nanocat utility." ${NN_TOOLS})
 
 #  Platform checks.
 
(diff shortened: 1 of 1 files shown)


JackDunaway:
CC'ing @gdamore for visibility to discuss this PR further, especially in the context of #641

**It makes sense to standardize on `options` flag conventions.** Here in #591, `NN_TOOLS` and `NN_TESTS` are introduced, and in #641 three other flags are introduced: https://github.com/nanomsg/nanomsg/pull/641/files#diff-af3b638bc2a3e6c650974192a53c7291R84

  - shall we use the `NN_` prefix? I'm ambivalent, except strongly prefer a yes or no versus mixed prefixes (this, having introduced mixed prefixes myself in #591!)
  - shall we place all `options` near the top of the script, rather than near the code it cases? (I prefer seeing them all at the top, rather than hunting through CMakeLists.txt to discover all the options it exposes)
  - Here in #591, `NN_TOOLS` overlaps with `ENABLE_NANOCAT` introduced in #641. Again a prefix question -- `ENABLE_` is somewhat of a CMake convention ... do we use it? Would the best option be `NN_ENABLE_TOOLS` or `ENABLE_TOOLS` or `NN_NANOCAT` or ... ? (Again, ambivalent here, except that it be uniform)

Snaipe:
@JackDunaway When I started using cmake, I didn't prefix my options with anything and went with the `ENABLE_FOO` convention, but as I started working with dependencies, it turns out that not prefixing options with anything results in more of a problem than anything.

The fact is, even though people should be using cmake's `ExternalProject_Add` to build dependencies, they often end up doing `add_subdirectory` on the project root for many reasons (mainly because ExternalProject is a hassle to use by itself, doesn't reuse caches, and almost doubles the build time due to mid-build reconfigurations). In that situation, not adding prefixes to options may lead to conflicts (although you may argue that something like `ENABLE_NANOCAT` is *very* unlikely to be a source of conflicts).

Again, this is more of a preference more than anything, but not having to fix a project's cmake configuration is a big plus when the price to pay is a three character prefix.

JackDunaway:
> not having to fix a project's cmake configuration is a big plus when the price to pay is a three character prefix

Good point. `NN_` prefix seems like it should be the standard here.

CMakeLists.txt
```diff
@@ -137,7 +147,9 @@ install (FILES src/pipeline.h DESTINATION include/nanomsg)
 install (FILES src/survey.h DESTINATION include/nanomsg)
 install (FILES src/bus.h DESTINATION include/nanomsg)
 
-install (TARGETS nanocat RUNTIME DESTINATION bin)
+if (NN_TOOLS)
```
gdamore: We need this in the current work.

JackDunaway: It is, just not with this name. The question is whether it makes sense to keep the current flag strategy of a one-flag-per-tool basis, or to lump them all under a single `TOOLS` flag.

It's a bit moot right now, since there's just one tool -- but I now lean toward keeping per-tool enables rather than lumping.

Snaipe: You could have both. I am currently bundling nanomsg inside a project, and passing `-DNN_TOOLS=OFF` to disable all tools would be more convenient than specifying a list of `-DNN_ENABLE_XXX=OFF` (and update it if any tool gets added in the future). I agree that this is very far fetched since there is only nanocat, though.

On a complexity standpoint, adding a `NN_TOOLS` option with specific `NN_ENABLE_XXX` options isn't doing much: You'd just have to specify the default value of every `NN_ENABLE` option with `${NN_TOOLS}`.

CMakeLists.txt
```diff
@@ -82,9 +81,11 @@ endif()
 
 # User-defined options.
 
-option (NN_ENABLE_NANOCAT "Enable building nanocat utility." ON)
 option (NN_STATIC_LIB "Build static library instead of shared library." OFF)
 option (NN_ENABLE_GETADDRINFO_A "Enable/disable use of getaddrinfo_a in place of getaddrinfo." ON)
+option (NN_TESTS "Build and run nanomsg tests" ON)
+option (NN_TOOLS "Build nanomsg tools" ON)
+option (NN_ENABLE_NANOCAT "Enable building nanocat utility." ${NN_TOOLS})
```
JackDunaway: Nice; this is the best of both worlds with a clean, simple syntax.

JackDunaway:
@gdamore, any issues here? If no, I can mainline; looks good to me.

gdamore:
Looks good to me.  I'm not too thrilled about having both enable for tools and for nanocat separately -- but its not really worth holding this up for.  (In the future I think we should stick with just one of these.)

JackDunaway:
fast-forward merged; thank you, @Snaipe!

PR #595, opened 2016-01-21
fix some typos and indentation anomalies

paddor:


gdamore:
I appreciate the effort you spent here, but I'm probably not going to accept this.  The RFCs themselves need to be reformatted, updated in some cases, and moved into a separate repo.

The new format will be asciidoc, so we don't have to deal with the xml2rfc nonsense.

PR #601, opened 2016-02-06
fixes #632 GCC int-conversion warning with MinGW-w64

### LINKED ISSUE #632: GCC int-conversion warning with MinGW-w64
When compiling nanomsg with mingw32-w64-gcc, the following warning is issued.

```
In file included from C:/projects/criterion/dependencies/nanomsg-patched/src/utils/err.h:33:0,
                 from C:/projects/criterion/dependencies/nanomsg-patched/src/aio/worker_win.inc:27,
                 from C:/projects/criterion/dependencies/nanomsg-patched/src/aio/worker.c:26:
C:/projects/criterion/dependencies/nanomsg-patched/src/aio/worker_win.inc: In function 'nn_worker_routine':
C:/projects/criterion/dependencies/nanomsg-patched/src/utils/fast.h:27:38: warning: passing argument 1 of '__builtin_expect' makes integer from pointer without a cast [-Wint-conversion]
 #define nn_fast(x) __builtin_expect ((x), 1)
                                      ^
C:/projects/criterion/dependencies/nanomsg-patched/src/aio/worker_win.inc:176:17: note: in expansion of macro 'nn_fast'
             if (nn_fast (entries [i].lpOverlapped)) {
                 ^
```

This happens because `entries [i].lpOverlapped` is a pointer, and `__builtin_expect` compares this value to 1, an int value.

Snaipe:
This happened because __builtin_expect compares the pointer to an int.

### DIFF
--- a/src/aio/worker_win.inc
+++ b/src/aio/worker_win.inc
@@ -173,7 +173,7 @@ static void nn_worker_routine (void *arg)
         for (i = 0; i != count; ++i) {
 
             /*  Process I/O completion events. */
-            if (nn_fast (entries [i].lpOverlapped)) {
+            if (nn_fast (entries [i].lpOverlapped != NULL)) {
                 op = nn_cont (entries [i].lpOverlapped,
                     struct nn_worker_op, olpd);
 


gdamore:
This change looks obviously correct.  Have you filed a bug on this -- we like to use a specific comment format in our commits...

fixed #XXX bug sysnopis

Which means you need an XXX and a bug synopsis.  I can do this for you if you like, but when I commit the comments will change.

PR #608, opened 2016-02-23
Update global.c

tbmale:
In this case gmtime_r, having mingw doesn't exclude windows.


gdamore:
I'm rejecting this PR, because I'm going to just rip out the offending code altogether.  The statistics socket that used this code is getting ripped out.

PR #612, opened 2016-03-10
Introducing nn_allocmsg_ptr and a set of changes to accomodate it

wavesoft:
Hi all,

in my project I had the requirement to have true zero copy from a pointer that I received from the system, therefore the `nn_allocmsg` was not really working for me. However, I realized that `nn_chunk` is only used in `nn_chunkref` and it's structure is quite straightforward to extend. So I added an additional support to be able to carry a pointer instead of the raw data itself.

I tried to be as compatible as possible, but I am looking for your feedback on this.

In addition, for a [new transport that I am developing](https://github.com/wavesoft/nanomsg-transport-ofi) I need to perform memory registration before sending or receiving data, and therefore I wanted some detailed insight on the private structures of `nn_chunk`. However, since that part of the structure is critical, I introduced a new function: `nn_chunk_describe` in order to get the chunk description (base address, length).

Finally, since there is a chance of a memory region to be de-allocated and re-allocated on the same address I introduced an additional field in the `nn_chunk` structure : the `id` that changes every time a new message is allocated. This helps again with memory registration, since I can identify changes and re-register the re-allocated memory regions.

I will also comment on my source code lines.  So, briefly noting:
- Adding support to carry pointer in src/utils/chunk.h
- Adding nn_chunk_describe to assist memory registration functions
- Adding nn_chunk_alloc_ptr to allocate chunk by pointer
- Adding nn_chunk_deref to unify the access to the underlaying data
- Adding new support to src/utils/chunkref.h


src/utils/chunkref.c
```diff
@@ -82,14 +82,14 @@ void *nn_chunkref_getchunk (struct nn_chunkref *self)
 
     if (self->u.ref [0] == 0xff) {
         ch = (struct nn_chunkref_chunk*) self;
-        self->u.ref [0] = 0;
+        // self->u.ref [0] = 0;
```
wavesoft: I could not find the reason why this is here. It looks that when somebody calls `nn_chunkref_getchunk` the internal reference is just released. This makes no sense, since there is no other way to get the internal chunk and there might be more than one call needed.

src/utils/chunk.c
```diff
@@ -31,9 +32,15 @@
 #include <string.h>
 
 #define NN_CHUNK_TAG 0xdeadcafe
+#define NN_CHUNK_TAG_PTR 0xdeadd00d
```
wavesoft: I am trying to reuse as much as possible from the original chunk structure, with the smallest memory impact possible.

In order to differentiate a classic chunk vs a pointer chunk I could have used two approaches:
- Introduce a `type` field in the `nn_chunk` structure (adds an extra byte in memory)
- Use a different tag (needs an extra tag check, but requires fewer tests in order to get the underlying pointer)
I chose the second.

wavesoft:
Looking into the errors reported by CI tests...

wavesoft:
Hi all, just a heads-up regarding the new changes:

- For semantical correctness I changed `nn_chunkref_getchunk` to only _**get** the chunk_ and I introduced `nn_chunkref_popchunk` to _**get** the chunk and **reset** the reference_.
- I also added a test for the `nn_chunk_alloc_ptr` function that I introduced

I am looking forward to your feedback :smile:

wavesoft:
Summarising all the updates, including the original commits and the modifications ever since:

### Changelog

* **FIX:** Making sure `nn_chunkref_mv` clean-ups the source after moving data to destination.
* **FIX:** Introducing the `nn_chunkref_popchunk` to semantically differentiate from `nn_chunkref_getchunk`. The first should reset the chunk reference after returning/creating the referred chunk, while the latter should keep it unharmed.
* **ADD:** Introducing `nn_chunk_alloc_ptr` to allocate zero-copy messages from existing data + user-provided deallocator function. 
* **ADD:** Adding user pointer in `struct nn_chunk`, passed to the chunk de-allocator function.
* **ADD:** Adding `nn_chunk_replace_free_fn` that that allows replacing of the free function of a chunk. This combined with the user pointer allows chaining of free functions.
* **ADD:** Adding `nn_chunk_deref` that must be used in order to receive the pointer to the chunk. This either returns the base address to the chunk data, or the pointer the chunk carries.
* **ADD:** Adding `nn_chunk_reset` that resets the contents of the chunk to it's original state (ex. removing empty space) and optionally hard-replace the size of the chunk. This is particularly useful for allocating a chunk only once and re-using it.

wavesoft:
I made an important decision that I wanted the potential users of this function to know: 
> :warning:  _Please do not use this function inside the transport or protocol!_

Let me elaborate on this: If you allocate a message using `nn_chunk_alloc_ptr` and pass it to the pipebase receive function, the user will end-up with a pointer to a **pointer** and not to the data. This means that the following code won't work:

```c
void *buf = NULL;
nbytes = nn_recv (s, &buf, NN_MSG, 0);

/* ERROR: buf is a pointer to a pointer (void**) not pointer to data! */
printf( "Data: %s\n", buf );

/* This would work, but this breaks the API, since the user does not
   know if buf is a data pointer or a pointer to a pointer */
printf( "Data :%s\n", *((char**)buf));
```

### Changelog

* **ADD:** Merged #620 in order to avoid relying on `nn_chunk_alloc_ptr` in the transport in order to allocate page-aligned memory.

gdamore:
I apologize, but I've probably completed busted this.

At this point, I'm probably not going to be accepting new transports for nanomsg ...

BUT the new nng library is a totally different matter, and for that I'd like to look at handling allocations and such to support RDMA, shared memory, and so forth.

gdamore:
Conflicts left and right.... my fault.  But as I said, long term I think this isn't going to be accepted upstream into nanomsg... instead we should look at supporting your new transport in nng.  I will be far more engaged in helping you with that, if you're so inclined.

PR #616, opened 2016-04-04
cmake: Add initial support for linux

xor-gate:
This adds to initial use of cmake build system on Linux systems. 
Tested under debian 8 amd64.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -66,5 +66,4 @@ Tobias Peters <tobias.peters@kreativeffekt.at>
 Victor Guerra <vguerra@gmail.com>
 Yonggang Luo <luoyonggang@gmail.com>
 Zoltan Boszormenyi <zboszor@pr.hu>
-
-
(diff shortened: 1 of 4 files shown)


josephsmeng:
the author of these CMake files seems like adding a space after each commands, so I think maybe it would be better for as to do the same for the consistency. some lines does not obey this rule, like the `install` command from the src/CMakeLists.txt, I think you can fix this in your pulls. thanks for your codes.

xor-gate:
@josephsmeng you are right, i have improved this and also platform checks are now way better. Now we can use `poll2` syscall if it exists.

xor-gate:
I leave this now as-is because it seems some people already tried to integrate cmake for cross-platform. Probably this can be a good base to have full cross-platform support on QNX,OSX,Solaris etc.

gdamore:
I'm gonna close this, and we'll hope for @JackDunaway to submit a PR addressing this shortly.

PR #620, opened 2016-04-11
Page-aligned chunk allocator

wavesoft:
Since some more advanced transports (such as one using RDMA) require page-boundaries alignment, I thought it might be a good idea to have such support available in the core.

This commit introduces the `NN_ALLOC_PAGEALIGN` alloc type that can be used on any chunk operation in order to align the chunk's memory on page boundaries. For example:

``` c
void *msg = nn_allocmsg( 1024, NN_ALLOC_PAGEALIGN );
```

This function will return `-ENOSYS` if aligned-alloc is not supported by your system. Currently that's windows, OSX, and *nix platforms with older POSIX api. 

_NOTE: This will overlap with #612, however I wanted to make a stand-alone pull request in case it's of a general interest_

### DIFF
--- a/doc/nn_allocmsg.txt
+++ b/doc/nn_allocmsg.txt
@@ -26,7 +26,13 @@ default one, however, individual transport mechanisms may define their
 own allocation mechanisms, such as allocating in shared memory or allocating
 a memory block pinned down to a physical memory address. Such allocation,
 when used with the transport that defines them, should be more efficient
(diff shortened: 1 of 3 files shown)


gdamore:
This is not ready, and at this point I'm punting on it for nanomsg.  If we need this kind of behavior for nng, please let me know.  (I am considering the issue there.)

PR #624, opened 2016-04-14
fixes #623 Programmatic access to statistics desired

### LINKED ISSUE #623: Programmatic access to statistics desired
We should expose statistics programmatically.

This means exposing all the NN_STAT_xxx defines (including the nn_symbol machinery, where we should declare NN_UNIT_COUNTER and NN_UNIT_MESSAGES to help.  (The COUNTER type indicates a stat that only increases, where as we can use UNIT_NONE for stats that represent a current level, such as NN_STAT_CURRENT_CONNECTIONS

One wrinkle in all this is that we need to expose a 64-bit integer type to applications.  I'm choosing to create NN_UINT64 for this purpose.  For now.  I'd really really rather use standard uint64_t -- but that might not exist everywhere. 

While here I've also cleaned up the symbol declarations to make them easier to read and maintain.

gdamore:
This is a nice API for accessing stats.  Integration of this should be considered a pre-requisite for removal of the stats socket and printing functionality.

Note that I've included a test of this functionality as well.


src/core/symbol.c
```diff
@@ -38,190 +39,143 @@
...
+    NN_SYM(NN_TYPE_NONE, OPTION_TYPE, NONE, NONE),
+    NN_SYM(NN_TYPE_INT, OPTION_TYPE, NONE, NONE),
+    NN_SYM(NN_TYPE_STR, OPTION_TYPE, NONE, NONE),
+
+    NN_SYM(NN_UNIT_NONE, OPTION_UNIT, NONE, NONE),
+    NN_SYM(NN_UNIT_BYTES, OPTION_UNIT, NONE, NONE),
+    NN_SYM(NN_UNIT_MILLISECONDS, OPTION_UNIT, NONE, NONE),
+    NN_SYM(NN_UNIT_PRIORITY, OPTION_UNIT, NONE, NONE),
+    NN_SYM(NN_UNIT_BOOLEAN, OPTION_UNIT, NONE, NONE),
+    NN_SYM(NN_UNIT_COUNTER, OPTION_UNIT, NONE, NONE),
```
gdamore: Crap.  NN_UNIT_MESSAGES is missing here.

src/nn.h
```diff
@@ -32,6 +32,17 @@ extern "C" {
 #include <errno.h>
 #include <stddef.h>
 
+/*  64-bit integer type for statistics.  */
+#if defined(_WIN32)
+typedef	 __uint64 NN_UINT64
```
gdamore: Apparently I got this wrong for windows.

src/core/symbol.c
```diff
@@ -38,190 +39,143 @@
...
-        NN_TYPE_NONE, NN_UNIT_NONE},
-    {NN_PULL, "NN_PULL", NN_NS_PROTOCOL,
-        NN_TYPE_NONE, NN_UNIT_NONE},
-    {NN_SURVEYOR, "NN_SURVEYOR", NN_NS_PROTOCOL,
-        NN_TYPE_NONE, NN_UNIT_NONE},
-    {NN_RESPONDENT, "NN_RESPONDENT", NN_NS_PROTOCOL,
-        NN_TYPE_NONE, NN_UNIT_NONE},
-    {NN_BUS, "NN_BUS", NN_NS_PROTOCOL,
-        NN_TYPE_NONE, NN_UNIT_NONE},
+#define	NN_SYM(sym, namespace, typ, unit) \
```
JackDunaway: Is the lack of space between `#define` and `NN_SYM` in the source or just a rendering problem?

Also: generally not a fan of preprocessor breaking IDE's ability to "Find All" on a symbol, but it appears this one is localized and contained to this single source file, which makes it not too bad (if not, actually kinda nice). Thoughts? Seems reasonable here.

src/nn.h
JackDunaway: Yes, even without a compiler in front of me, this feels wonky; I think `uint64_t` is defined by `<stdint.h>` on Windows for MSVC compiler.

src/core/symbol.c
gdamore: Yeah, that's a formatting bug in github. Go figure.

There's a real benefit here, which is that the stringifcation is now automated, eliminating a rather large opportunity for human error.  (The symbol names will fail a compile check if they are wrong. :-)

The compact representation makes it easier to see what's going on.  I understand your gripe about breaking the IDE search function, but I think the benefits outweigh the cons in this case.  (In fact, you'll still find the reference nearby in this file, because the symbols that are "abbreviated" are listed in full in their own entries in this array.)

src/nn.h
```diff
@@ -387,9 +401,35 @@ NN_EXPORT int nn_device (int s1, int s2);
...
+
+/*  The socket-internal statistics  */
+#define NN_STAT_MESSAGES_SENT           301
+#define NN_STAT_MESSAGES_RECEIVED       302
+#define NN_STAT_BYTES_SENT              303
+#define NN_STAT_BYTES_RECEIVED          304
+/*  Protocol statistics  */
+#define	NN_STAT_CURRENT_SND_PRIORITY    401
+
+NN_EXPORT NN_UINT64 nn_get_statistic (int s, int stat);
```
JackDunaway: I don't disagree with getting stats one at a time, but to get ahead of potential question -- why did you decided that statistics can only be retrieved one at a time?

What would the developer experience look like to be able to iterate through all known statistics (without needing to maintain a list of constants to the available stats)? Is there such a facility to discover all available statistics?

src/nn.h
gdamore: In modern MSVC (2010 edition or later) this is true.  Older versions don't have it.  I fixed it for now, but I hate this.  If we can just state that MSVC 2010 or newer is required, that would be simpler.  (See also the __VMS hackery for OpenVMS.  Puke.)  What I don't know is whether requiring MSVC 2010 or newer will cut off an additional round of platforms (Vista or Windows 7) that would otherwise be supported.  I have tried (mostly successfully) remain blissfully unaware of Microsoft platforms up until now...

src/nn.h
gdamore: Turns out this is possible (I had already thought about it). Using the nn_symbol() capability, you can iterate over all symbols, and just look for the ones that are in the NN_NS_STATISTIC namespace.  So everything is discoverable.

Offering up all of them would have been a bit trickier to do in an interface that is meant to evolve (as new stats get added, and possibly old ones deprecated).  This way we can evolve without breaking binaries due to incorrectly compiled structure sizes or somesuch.

And the whole thing is discoverable without me having to invent a new API, which I love.  (Not that I really have any love of the nn_symbol() API -- but in this case it solves the problem for us.)

tests/stats.c
```diff
@@ -0,0 +1,95 @@
...
+
+    nn_assert (nn_get_statistic(rep1, NN_STAT_ACCEPTED_CONNECTIONS) == 1);
+    nn_assert (nn_get_statistic(rep1, NN_STAT_ESTABLISHED_CONNECTIONS) == 0);
+    nn_assert (nn_get_statistic(rep1, NN_STAT_CURRENT_CONNECTIONS) == 0);
+
+    test_close (rep1);
+
+    return 0;
+}
+
```
JackDunaway: In the past, statistics have been pretty heavily marked as being for informational use only, and not intended to control program flow -- even within libnanomsg itself.

Is this new public function, `nn_get_statistic()`, now officially provided in order to help steer application program flow?

Or does it still come with the strong disclaimer "for debugging purposes only"?

The reason I ask, is this significantly affects the previous surface area of the libnanomsg API, exposing what used to be encapsulated, private data.

I make no claim at this point whether it's good or bad, or whether the decision should be applied on a per-statistic basis.

It seems helpful, but those are the types of things that could later prove hurtful. At least, it's worth heavy consideration how this new ability is branded and documented.

src/core/symbol.c
JackDunaway: Ah, the formatting issue could be this (now seeing an alternate rendering within my email client) -- is it a TAB instead SPACE?

Which brings up what should be a new issue -- I feel we should adopt a standard code linter with specified configuration of the linter, to lower the barrier as much as possible for contributions. A nit should never gate a contribution.

tests/stats.c
gdamore: Valid point.  I think we should advise caution.  While the API itself can be considered stable, some of the statistics may be considered unstable, and relying upon their values programmatically for purposes other than visibility or debugging is at least strongly discouraged.

That said, this is a level of visibility that can be *useful*.  Imagine knowing that you could monitor connections statistic to determine when to start flooding data... etc.  I do imagine that I'd wind up using this information for the express purpose of test cases, since now we have visibility into some state that we previously lacked.  But I wouldn't want to do any of that in production.

gdamore: In fact, I've forgotten to write a man page for this API.  Argh.  Later.

gdamore: Btw, I apologize for the coding style screw ups.  I actually don't like the nanomsg coding style -- mostly because its different from the BSD/illumos coding style that I use everywhere else.  I occasionally make screw ups (like failing to insert whitespace before leading parenthesis.)  I'd actually love to change the style, and add some verification tools (e.g. illumos' cstyle checking program) to ensure that we are conforming to the style...  something we can discuss later.

src/nn.h
JackDunaway: Win7 works perfectly fine even thru the latest-greatest MSVC 2015. Speaking as one stakeholder on this platform, if MSVC 2010 is the lowest common denominator for this particular feature (among perhaps others), I find it perfectly acceptable to make >=2010 the cutoff.

WinXP -- I cannot speak for, because my business officially does not support this. It's >2yrs out of Microsoft's support at this point.

Feel free to put out a call whether versions *prior* to MSVC 2010 should still be supported, but from my point of view, there should be no problem making >=2010 a requirement.

tests/stats.c
JackDunaway: > for the express purpose of test cases .... But I wouldn't want to do any of that in production

Ah ... that's exactly what threw the flag when looking at the test case. A test case is app code ... yes? (Suppress -- or rather, cue -- religious debate ...)

Again, I'm still reserving judgment applying "good" or "bad" to each element of "this particular set" of statistics.

The obvious path right now is this: **keep the API, since it's good, and mark it's usage as strictly informational and subject to change prior to further notice**.

JackDunaway: > I apologize for the coding style screw ups

It's a nit. Who cares? The linter (yet to be chosen); that's who.

PR #631, opened 2016-04-16
fixes #621 Remove experimental tcpmux transport

### LINKED ISSUE #621: Remove experimental tcpmux transport
The tcpmux transport was an experiment, but ultimately it turns out not be very useful given the websocket transport.

TCP mux also has portability problems, and really to be useful requires root privilege (so it runs on port 1), which makes it mostly useless for a large number of scenarios.  The websocket transport doesn't suffer from this problem at all.

gdamore:
This removes the tcpmux transport.  Better to use websocket.  (Hmm... we need to uplevel the capabilities of the websocket transport too, and promote it out of experimental status.)


gdamore:
Merged via rebase:

commit 5bb109305b5bd738576f9010f7264e4136bd1456
Date:   Sat Apr 16 08:28:59 2016 -0700

    fixes #621 Remove experimental tcpmux transport

diff --git a/CMakeLists.txt b/CMakeLists.txt

PR #636, opened 2016-04-16
fixes #574 Cross-platform CMake Build

### LINKED ISSUE #574: Cross-platform support for CMake builds
This is a new open case for the continuation of #210, which was originally closed due to lack of contributions.

Here's a snapshot of work done toward fully supporting this feature request, nearly ready for a pull request: https://github.com/wirebirdlabs/featherweight-nanomsg/compare/pr-ws-mega-squash...wirebirdlabs:ftw-kernel

One blocker before creating this PR is merging #569; especially the CMakeLists.txt upgrade in https://github.com/wirebirdlabs/featherweight-nanomsg/commit/03f4dc2f29c8b1947f10f22898e3630221924fd1

The CMake script is currently passing on Windows environments in AppVeyor, and also the default Linux environment in Travis CI.

This patch aims to provide for CMake users the full capabilities of the current autotools toolchain. Those opting to switch from autotools to CMake are kindly requested to help suss out details or missing features.

Some high-level changes in this patch:

- platform-specific source files (e.g., `usock_win.h` versus `usock_posix.h`) are now conditionally included. This should provide a better developer experience within the IDE, when say, searching for a symbol
- ability to build static library with the configure option `NN_STATIC_LIB=ON`
- abitility to enable/disable nanocat build with `ENABLE_NANOCAT=ON`
- standardization of the `NN_HAVE` versus `NN_USE` nomenclature, both in source and in the configure script
- grouping of source files into a hierarchical folder structure in the IDE using CMake's `source_group` feature

The inclusion of this patch should not affect current autotools users, except I plan to makes minor upgrades to those script also to better conform to the `NN_HAVE` vs `NN_USE` definitions.

JackDunaway:
This PR is the first attempt at squashing and rebasing CMake work from several branches that has diverged from mainline for several months. It is meant to unify all platforms into a single CMake build system, in order to eventually deprecate the autotools toolchain.

This PR includes some commits from @shiva (notably, lines 52-81 in `CMakeLists.txt`) that were squashed as an unfortunate byproduct of consolidating all these branches; the original commits may be found here: https://github.com/shiva/cmakeify-nanomsg/commits/cmakeify

@gdamore, we'll likely need to update both the Appveyor and Travis dashboard settings. Also, we'll probably need to add a new package as part of the Travis pre-build, since IIRC that environment does not have CMake 2.8.12 installed by default. First though, let's see what happens when the build is triggered on this PR ...


JackDunaway:
Looks like these few follow-on commits are getting Travis CI to build.

Action items:
1. @gdamore -- turn on AppVeyor to obey `appveyor.yml` instead of dashboard settings
2. @JackDunaway -- tomorrow, re-squash this PR -- ideally properly attributing commits by @shiva
3. @shiva -- note this familiar failure: https://travis-ci.org/nanomsg/nanomsg/jobs/123621308#L264 and the "fix" the suppressed this issue: https://github.com/nanomsg/nanomsg/pull/636/commits/e9dc4e91e816a6062ac945e5cd46d4a793555e2e -- any ideas on a more sustainable fix?
4. @gdamore -- we'll need to look into packaging and deployment on tagged commits.

I'm away for the rest of today, with limited availability chatting tomorrow (Sunday), but will be back looking into this ~40hrs from now on Monday.

JackDunaway:
@shiva -- if you work any more with this, begin work from this branch instead of the old one. Note though, that I'm going to try to re-do some squashes to preserve your attribution, meaning this branch will even be reset from where it current sits -- meaning any further work you do may get similarly shuffled around :-)

shiva:
don't fret too much about attribution. Do what you need to get it in.

I'll fork again and continue to work on it. 

Will get back on the other issue in a couple of days.

JackDunaway:
I am currently investigating why the AppVeyor builds are failing: https://ci.appveyor.com/project/gdamore/nanomsg/build/0.8.392

It appears that Travis CI building on Ubuntu is healthy: https://travis-ci.org/nanomsg/nanomsg/builds/123624427

As part of this process, I will also turn on OS X builds on Travis CI. The goal is to get as much mileage on as many platforms as possible to flush out issues, since each environment is built each time a commit is added to a PR.

After sussing out issues with what could be quite a few pushes to this branch, I'll then squash all commits into a cogent PR and close this one out.

Snaipe:
@JackDunaway if that helps, I'm using [this CMakeLists.txt](https://github.com/Snaipe/nanomsg/blob/master/CMakeLists.txt) on my own fork to compile nanomsg on Linux, OS X and Windows (MSVC + MinGW-w64)

.travis.yml
```diff
@@ -6,10 +6,10 @@ matrix:
       compiler: gcc
     - os: linux
       compiler: clang
-#    - os: osx
-#      compiler: gcc
-#    - os: osx
-#      compiler: clang
+    - os: osx
+      compiler: gcc
```
Snaipe: FYI, 'gcc' is aliased to clang/llvm on OS X workers, if you really want gcc, you need to specify gcc-4.8 for instance (I don't remember if it is installed by default, but you can still install it through `brew`)

.travis.yml
```diff
@@ -1,12 +1,29 @@
...
-- clang
-- gcc
+matrix:
+  include:
+    - os: linux
+      compiler: gcc
+    - os: linux
+      compiler: clang
+    - os: osx
+      compiler: gcc
```
Snaipe: FYI, 'gcc' is aliased to clang/llvm on OS X workers, if you really want gcc, you need to specify gcc-4.8 for instance (I don't remember if it is installed by default, but you can still install it through `brew`)

gdamore:
Ok, I am going to spend some time looking at this too…

.travis.yml
JackDunaway: Good to know; I'm taking a look through your CMake script now ...

JackDunaway:
> if that helps, I'm using this CMakeLists.txt on my own fork

It's particularly helpful! -- especially that you're able to build with CMake 2.8.7 instead of CMake 2.8.12!

Being able to lower this requirement would reduce setup time, since 2.8.7 is what ships with the default Travis CI environment. At some point during original development of this script, I recall introducing enough features to justify the upgrade to 2.8.12 from 2.8.7, but seeing what little effort you spent to drop back to 2.8.7 seems like the way to go.

I'll be soon cherry picking and pushing updates ...

gdamore:
CMake 3.4 is the latest, I believe.  Are there reasons we should be using
the older 2.x series?

JackDunaway:
> CMake 3.4 is the latest, I believe.  Are there reasons we should be using the older 2.x series?

Brief, glossed-over answer: yes.

gdamore:
That was *really* too brief. :-/

JackDunaway:
Longer answer -- based on my last in-depth technical investigation, which is now a few months dated and always a moving target, it appeared the options were:

* **CMake 2.8.7** - seems this is the latest-pegged "stable" version in trusty and LTS linux builds, including the Travis CI environment with Ubuntu 12.04 LTS: http://packages.ubuntu.com/precise/cmake
* **CMake 2.8.12** - most up to date CMake release in the 2.x line, notably including commands to more precisely specify target properties rather than directory properties 
* **CMake 3.x** - a moving target (seems to be updated quite often), and not particularly accessible within the Travis CI environment, especially because Travis CI does not allow `sudo` privilege for containerized builds.

The new piece of information I'm learning today from @Snaipe is that he's successfully been back at 2.8.7, which feels definitely worth investigating, since it lowers CI build time appreciably and removes script complexity.

At some point, it could be worth investigating ugprading to 3.x, but at this point, 2.8.7 makes the most sense (or, 2.8.12 if it turns out that requirement remains)

gdamore:
Ok.  I use 3.4 (or 3.5rcX) myself, because i have to install it myself on
OSX.  If its easy to make the old stuff work, and doesn’t break folks
working with newer CMake, then I agree.

.travis.yml
```diff
@@ -1,12 +1,29 @@
...
+    packages:
+    - cmake
 script:
-- ./autogen.sh
-- ./configure
-- make distcheck
+  - mkdir build
+  - cd build
+  - cmake --version
+  - cmake -DCMAKE_C_COMPILER=$CC ..
```
Snaipe: `CC` is already picked up by cmake IIRC

JackDunaway:
Beginning to study and remember the differences between CMake 2.8.7 and 2.8.12 -- I'm recalling some specific issues when a project such as nanomsg is included as a static library in a parent build, where the parent build uses [`add_subdirectory`](https://cmake.org/cmake/help/v2.8.7/cmake.html#command:add_subdirectory).

The issues dealt with the abilities to override and also to inherit properly target settings of such sub-projects from the main project, and hinges around these changes made by @Snaipe to support 2.8.7: https://github.com/Snaipe/nanomsg/commit/2a3dc72ed15e83e507835b70967833ea557f0077#diff-95e351a3805a1dafa85bf20b81d086e6L369

@Snaipe or others -- does this sound familiar? Is it a technical reason why 2.8.12 would be a requirement over 2.8.7, where the patch linked above precludes behavior required by `add_subdirectory`? At the moment, I'm continuing the technical investigation, but asking in hopes of short-circuiting the investigation ...

JackDunaway:
Aha ... I can answer my own question here.

The limitations of `add_definitions` compared to `target_compile_options` indeed boils down to inheritance of values and being able to override from higher levels.

However, in CMake versions prior to `target_compile_options` being available -- including 2.8.7 -- you can achieve the same behavior by providing one degree of indirection between variables set at a higher layer, and target properties set at the lower (nested project) layer.

Apparently, as I was working on developing the CMake script, this pattern incidentally emerged, with the addition of this line: https://github.com/nanomsg/nanomsg/pull/636/commits/a15356b79e47e0c6ff48ba08e4dff538c7750844#diff-af3b638bc2a3e6c650974192a53c7291R86

The variable `NN_STATIC_LIB` provides that one degree of indirection that may be set upstream in a parent `CMakeLists.txt`. This variable is then read later: https://github.com/nanomsg/nanomsg/pull/636/commits/a15356b79e47e0c6ff48ba08e4dff538c7750844#diff-95e351a3805a1dafa85bf20b81d086e6R355 which then applies the target compile option: https://github.com/nanomsg/nanomsg/pull/636/commits/a15356b79e47e0c6ff48ba08e4dff538c7750844#diff-95e351a3805a1dafa85bf20b81d086e6R357

And here's the tricky bit. The variable passed in is called `NN_STATIC_LIB`, and the compile option is also called `NN_STATIC_LIB`, but they're not the same thing.

When initially developing this script, in my mind I must have been realizing that this compile option needed to change based on parent project logic, and must have been trying to set it in a more direct manner. But then this degree of indirection grew around it anyway, meaning 2.8.7 available functions would suffice.

**tl;dr** -- I think 2.8.7 will work. 2.8.12 functions that "felt like they were required" aren't actually because of an incidental design pattern that grew up around them, and turned out to be a little cleaner anyway.

shiva:
+1 to trying to go 2.8.7. 

Although most folks in open-source, and those using cmake for personal use tend to be on latest greatest, most production environments (\*cough\* large co \*cough\*) tend to use older more stable cmake versions due to stability issue with cmake itself. Further, it is harder to justify upgrading build tools, just to adopt one library. Much appreciate the efforts here to go to 2.8.7!

.travis.yml
JackDunaway: I think you're correct, but recall adding this explicitly at some point in the past for a reason ... that reason may no longer hold ... for this reason, it's better to remove this line and see if it still works in all cases. It can be revisited and re-added later if it's a problem.

JackDunaway:
> Much appreciate the efforts here to go to 2.8.7!

Thanks! Good to hear validation it's worthwhile; I see the value also, and was excited to see @Snaipe had achieved this. Seems to be passing the smoke test so far.

Snaipe:
> Thanks! Good to hear validation it's worthwhile; I see the value also, and was excited to see @Snaipe had achieved this. Seems to be passing the smoke test so far.

Glad to help. Quite a few number of my users are still stuck with 2.8.7, so being able to build nanomsg with it was a requirement. Nice to see it being integrated in the main branch!

CMakeLists.txt
```diff
@@ -179,6 +173,19 @@ if (NN_HAVE_PIPE2 OR NN_HAVE_ACCEPT4 OR NN_HAVE_GETADDRINFO_A)
...
+    int main()
+    {
+        int64_t i;
+        i = 0;
+        return 0;
+    }
+" NN_HAVE_STDINT)
+if (NN_HAVE_STDINT)
+    add_definitions (-DNN_HAVE_STDINT)
+endif ()
```
gdamore: Just remove this work... I'm not checking the definition of the macro anyway, and if it fails... ETOOBAD.

JackDunaway: Legit.

.travis.yml
```diff
@@ -1,12 +1,37 @@
...
+  # Print all environment variables to aid in CI development
+  - printenv
+  # Print version and available CMake generators to aid in CI development
+  - cmake --version
+  - cmake --help
+  # Perform out-of-source build
+  - mkdir build
+  - cd build
+  # Perform CMake backend generation, build, and test
+  - cmake -DCMAKE_C_COMPILER=$CC ..
```
gdamore: Is the definition of -DCMAKE_C_COMPILER required here?  (It doesn't pick up $CC automatically?)

CMakeLists.txt
```diff
@@ -22,35 +23,216 @@
...
+    find_package (Threads REQUIRED)
+elseif (CMAKE_SYSTEM_NAME MATCHES "OpenBSD")
+    add_definitions (-DNN_HAVE_OPENBSD)
+    find_package (Threads REQUIRED)
+elseif (CMAKE_SYSTEM_NAME MATCHES "Solaris|SunOS")
+    add_definitions (-DNN_HAVE_SOLARIS)
+    find_package (Threads REQUIRED)
+    check_library_exists (socket socket "" NN_HAVE_SOLARIS_SOCKET)
+    check_library_exists (nsl gethostbyname "" NN_HAVE_SOLARIS_GETHOSTBYNAME)
+elseif (CMAKE_SYSTEM_NAME MATCHES "QNX")
```
gdamore: So it turns out that while a HAVE_QNX macro was defined, nothing in the code uses it.  I'm unconvinced that QNX actually "works".  TBH, I'm not thrilled with this long explicit system listing.

gdamore: Clearly we need OSX and Windows checks.  We might need one for NetBSD (see poller_kqueue.inc), but the other system checks should not be needed.

We do need library checks for socket and gethostbyname (nsl).  I'm inclined to believe that those should not be tied to a specific platform.

CMakeLists.txt
```diff
@@ -22,35 +23,216 @@
...
+    add_definitions (-DNN_HAVE_HPACC)
+elseif (CMAKE_C_COMPILER_ID MATCHES "MSVC")
+    add_definitions (-DNN_HAVE_MSVC)
+elseif (MINGW)
+    message (FATAL_ERROR "ERROR: MinGW is officially no longer supported.")
 else ()
-    message (FATAL_ERROR "ERROR: CMake build system is intended only to generate MSVC solution files.\nUse autotools build system for any other purpose." )
+    message (AUTHOR_WARNING "WARNING: This compiler may or may not be supported: ${CMAKE_C_COMPILER_ID}")
+    message (AUTHOR_WARNING "${ISSUE_REPORT_MSG}")
 endif ()
```
gdamore: I don't like any of these compiler checks ... I think they serve to arbitrarily limit the source portability.  I'd remove this entire stanza.

CMakeLists.txt
```diff
@@ -22,35 +23,216 @@
...
-#  Build the library itself.
+check_function_exists (gethrtime NN_HAVE_GETHRTIME)
+check_function_exists (socketpair NN_HAVE_SOCKETPAIR)
+check_function_exists (eventfd NN_HAVE_EVENTFD)
+check_function_exists (pipe NN_HAVE_PIPE)
+check_function_exists (pipe2 NN_HAVE_PIPE2)
+check_function_exists (accept4 NN_HAVE_ACCEPT4)
+check_function_exists (epoll_create NN_HAVE_EPOLL)
+check_function_exists (kqueue NN_HAVE_KQUEUE)
+check_function_exists (poll NN_HAVE_POLL)
```
gdamore: Do any of the above need to be checked in a location other than libc?  I don't *think* so, but I'm not sure.

accept4 and pipe2 might need _GNU_SOURCE flags to be defined too.  To be honest its kind of unclear to me that pipe2() and accept4() are worth the support overhead of keeping them -- they reduce the number of system calls (eliminating an fcntl) required to establish a connection/pipe, but most systems should not have huge numbers of pipes connecting and disconnecting -- this isn't a web server with shortlived connections after all.

CMakeLists.txt
```diff
@@ -22,35 +23,216 @@
...
+
+if (NN_HAVE_SEMAPHORE_RT OR NN_HAVE_SEMAPHORE_PTHREAD)
+    add_definitions (-DNN_HAVE_SEMAPHORE)
+endif ()
+
+if (NOT ENABLE_GETADDRINFO_A)
+    add_definitions (-DNN_DISABLE_GETADDRINFO_A)
+endif ()
+
+if (NN_HAVE_PIPE2 OR NN_HAVE_ACCEPT4 OR NN_HAVE_GETADDRINFO_A)
```
gdamore: I think you may have missed this by not having the define in your test above.  Note that its unclear that this is always required -- netbsd offers at least pipe2 without this flag.

CMakeLists.txt
```diff
@@ -22,35 +23,216 @@
...
+    {
+        volatile uint32_t n = 0;
+        __sync_fetch_and_add (&n, 1);
+        __sync_fetch_and_sub (&n, 1);
+        return 0;
+    }
+" NN_HAVE_GCC_ATOMIC_BUILTINS)
+if (NN_HAVE_GCC_ATOMIC_BUILTINS)
+    add_definitions (-DNN_HAVE_GCC_ATOMIC_BUILTINS)
+endif ()
```
gdamore: It strikes me that it may be time to check for C11 atomics, now that we have a standard for this... (and add references for them).

CMakeLists.txt
```diff
@@ -109,13 +291,17 @@ add_libnanomsg_test (zerocopy 5)
 add_libnanomsg_test (shutdown 5)
 add_libnanomsg_test (cmsg 5)
 add_libnanomsg_test (bug328 5)
-add_libnanomsg_test (win_sec_attr 5)
+
+# Platform-specific tests
+if (WIN32)
+    add_libnanomsg_test (win_sec_attr 5)
+endif()
```
gdamore: Better yet, lets just change win_sec_attr so it returns success (skipped) on non-Windows platforms?

gdamore: Btw, we should carefully go through the tests to make sure that the autoconf test list is fully represented here.  I think I recall seeing some tests were missing from the CMake build.  Probably through oversight.

src/CMakeLists.txt
```diff
@@ -284,20 +253,133 @@ set (NN_SOURCES
...
+        utils/efd_socketpair.inc
+    )
+elseif (NN_HAVE_WINSOCK)
+    add_definitions (-DNN_USE_WINSOCK)
+    list (APPEND NN_SOURCES
+        utils/efd_win.h
+        utils/efd_win.inc
+    )
+else ()
+    message (AUTHOR_WARNING "ERROR: could not determine socket signalling mechanism.")
```
gdamore: I think "signaling" (one l)

src/CMakeLists.txt
```diff
@@ -284,20 +253,133 @@ set (NN_SOURCES
...
+    message (AUTHOR_WARNING "${ISSUE_REPORT_MSG}" )
+endif ()
+
+# Provide same folder structure in IDE as on disk
+foreach (f ${NN_SOURCES})
```
gdamore: I don't understand this block at all.

appveyor.yml
```diff
@@ -0,0 +1,22 @@
+version: 0.8.{build}
```
gdamore: As discussed, we should change this file to .appveyor.yml.  -- also we need to think about version numbering handling here.

appveyor.yml
```diff
@@ -0,0 +1,22 @@
...
+environment:
+  matrix:
+    # array of all environments used to test builds
+    - GENERATOR: Visual Studio 14 2015
+      CFG: Debug
+    - GENERATOR: Visual Studio 12 2013
+      CFG: Debug
+    - GENERATOR: Visual Studio 14 2015 Win64
```
gdamore: This is cool -- getting new warnings from the build now, which we should fix.

src/CMakeLists.txt
```diff
@@ -130,10 +103,6 @@ set (NN_SOURCES
     utils/sleep.c
     utils/thread.h
     utils/thread.c
-    utils/thread_posix.h
```
gdamore: Lets talk about the changes in this file -- as well as poller.inc -- previously we had this conditional logic in macros in the source files, but now you're moving it to the cmake build.  I'm not sure if that's an excellent idea.

src/nn.h
```diff
@@ -43,12 +44,14 @@ typedef unsigned __int64 NN_UINT64;
 #define NN_UINT64 uint64_t
 #endif
 
-/*  Handle DSO symbol visibility                                             */
+/*  Handle DSO symbol visibility. */
```
gdamore: I filed a bug to clean this up too.  NN_EXPORT should be extern on all non-windows systems, we can clean this crap up.

CMakeLists.txt
JackDunaway: I do not necessarily like the list of system checks either. Or compilers. Instead, I feel we should specifically check for **features**, which come and go from OS's and compilers.

libnanomsg should be concerned with symbols that define whether a feature exists ... not inferring whether that feature exists based on platform.

Note that your comment here, and a few below, are simple artifacts of a straight port from autotools.

In this case, here is the relevant autotools portion: https://github.com/nanomsg/nanomsg/blob/master/configure.ac#L57-L141

CMakeLists.txt
JackDunaway: Agree that we should try to move away from casing off of compiler rather than features. This section is a direct port from autotools: https://github.com/nanomsg/nanomsg/blob/master/configure.ac#L57-L141

src/utils/efd.h
```diff
@@ -1,6 +1,7 @@
 /*
     Copyright (c) 2012-2013 Martin Sustrik  All rights reserved.
     Copyright 2015 Garrett D'Amore <garrett@damore.org>
+    Copyright (c) 2015-2016 Jack R. Dunaway.  All rights reserved.
```
gdamore: Just a heads up about All rights reserved.  This sentence means *nothing* in modern copyright law, and hasn't since 2000 (when the last hold out to Berne -- Nicaragua -- finally signed the Berne convention.)  That's why I never list it.

Similarly "(C)" has no meaning, and never has.  Its an attempt to replace &copy; with an ASCII-ifcation, but even &copy; means nothing in the presence of the word "copyright" or the abbreviation "copyr."  The symbolic form was intended to allow a short symbol *instead* of the word, rather than in addition to.  The ASCII version has never had any legal meaning whatsoever in a copyright notice, only the actual symbol or the word "copyright" does.

All of which is why my notices are always just "Copyright <year> <name>".  (I add my email too, but that's optional.)  The <year> is the most recent date of publication; adding the first year of publication is also optional but carries no particular weight of law behind it.

gdamore: All that said, the copyright notice is yours, so you can make it what you want (within limits).  The above message was mostly intended as an educational FYI.  :-)

CMakeLists.txt
JackDunaway: A lot of this is ported directly from autotools, and could be stale. Let's try to get CMake mainlined today -- even along with cruft that is ported wholesale from autotools -- and then continue with the fine-tooth comb once mainlined.

Good strategy?

CMakeLists.txt
JackDunaway: > I think I recall seeing some tests were missing from the CMake build. Probably through oversight

It's possible, especially for tests that have been added in the past 6months. I'll do a quick audit now ...

shiva:
> 3. @shiva -- note this familiar failure: https://travis-ci.org/nanomsg/nanomsg/jobs/123621308#L264 and the "fix" the suppressed this issue: e9dc4e9 -- any ideas on a more sustainable fix?

@JackDunaway This commit (https://github.com/shiva/cmakeify-nanomsg/commit/8159d2166d0b8a5849ba7c04d1b2e14486952ebb), calculates the version, and guarantees a value for  NN_PACKAGE_VERSION. At least, that was the intention.

If this is not working as expected, I can troubleshoot more. Let me know.

(Note: I just noticed that I hadn't used my github identity for my commits and hence had to re-write history. You can safely ignore my commits now that you have squashed all of them into your PR branch.)

CMakeLists.txt
gdamore: Oh sheesh.  Yeah, lets take this opportunity to fix & improve rather than just "porting".  acc -- I forgot about that HPUX crap.  (There was also acc on SunOS 4.1.3... I'm showing my age now though...)

CMakeLists.txt
gdamore: Tend to agree, but for the stuff that's easy and obvious to clean up -- the compiler testing above for example -- lets do that now.

CMakeLists.txt
gdamore: This should just be a future RFE; lets have a plan to support C11 atomics, and threading too, in the future.

gdamore:
Once we get the commit ready — we’ll add a note to the commit comments
listing all the contributors — Git sadly doesn’t record more than one
author in the commit itself, but there’s no reason you shouldn’t get credit
for contributions in the git log.  We can also add copyright statements to
relevant files if you like.

src/CMakeLists.txt
JackDunaway: The reason I chose this method is to prune other-platform source code from the backend-generated project file -- in my case this is a Visual Studio Solution. By omitting, e.g., `worker_posix.inc`, this means when I search for symbols, it doesn't return duplicates in non-relevant platforms -- only `worker_win.inc` was included and relevant. It makes the IDE experience better.

That said ... there's merit in understanding other platforms, even if it's not relevant to your own current platform. In which case, I would prefer a rearchitecture of the source code to be along the lines of this: https://github.com/nanomsg/nanomsg/blob/master/src/utils/clock.c#L71-L121

Either way ... much of this entire CMake rearchitecture is a stepping stone to get from where we are to where we want to be, and in the meantime, it's likely to carry along some cruft that can then be hit later.

src/utils/efd.h
JackDunaway: IANL :D

src/CMakeLists.txt
gdamore: In this case, maybe rearchitecting the stuff *now* is something to wait
for; keep the existing solution used in autoconf and old cmake (which seems
to work well enough), and we can explore alternative better ideas in the
future.  This would have the benefit of reducing the scope of change
required for cmake support.

src/utils/efd.h
gdamore: Neither am I.  But I *have* been advised by one.

src/CMakeLists.txt
JackDunaway: For clarity; this is working. I'm not sure exactly what you're saying here.

I'm saying that as it's written here, it's a stepping stone from where we are to where we want to be, and it currently works with both CMake and the existing autotools. Once completely switching over to CMake, we can then begin to rearchitect source code and CMake together, without having to worry how it affects the now-gone autotools.

gdamore: What I was saying, or suggesting, was that I'd revert this portion of the change.  I think the old stuff worked, didn't it?  (Even when applied to cmake?)

My goal is to have the changeset required to enable cmake be the smallest change we can easily do), since and then refine it going forward.

JackDunaway: I see. That said, the number of source files touched in this commit is strictly limited to what's required based on what were discrepancies (bugs) in the previous autotools and CMake build, and changes this far don't really get into the type of refactoring we're talking about. I don't know if it would explicitly break from simply removing these changes from the commit, but it doesn't seem worth it to remove it.

CMakeLists.txt
JackDunaway: All tests seem accounted for.

JackDunaway:
Closing; this PR is superseded by #641.

PR #637, opened 2016-04-18
fixes #629 Go ahead and require stdint.h and uint64_t

### LINKED ISSUE #629: Go ahead and require stdint.h and uint64_t
The platforms that lack stdint.h are ... hard to find.

OpenVMS might be one of them.  But its unclear if its actually possible to build libnanomsg under OpenVMS -- no autotools support or CMake support there (yet!)  And C99 support for stdint.h exists in modern (VS 2010 and later) Windows build environments.

Its time to stop trying to pretend to support stuff we can't support -- heck we don't know if any of that stuff *works*, so it was always a hail mary shot. 

Let's focus on modernish platforms -- Windows, Mac, Linux, and POSIX compliant systems.  This should also help with the cmake work.

gdamore:
This removes our custom int64 and special header -- if you can't get a uint64_t and stdint.h in your environment, then it is broken and needs fixing.  (Note that OpenVMS may fall into this category, but we believe that it is probably impossible to build nanomsg in openvms right now anyway owing to insufficient support for either autotools or cmake.)

While here, remove explicit header checks that we never bother to reference or use.


JackDunaway:
Note related work where I am converging toward this same goal: https://github.com/nanomsg/nanomsg/pull/636/commits/bb2d45724bcdd5c75c098bd6e6d305da0acf2104

JackDunaway:
Though, it appears I should change https://github.com/nanomsg/nanomsg/pull/636/commits/bb2d45724bcdd5c75c098bd6e6d305da0acf2104#diff-af3b638bc2a3e6c650974192a53c7291R179 to `uint_64`, and also forgo actually exporting NN_HAVE_STDINT as a symbol, and instead throw a CMake error in case the compile check fails.

gdamore:
I'd be inclined to just punt on explicit checks for stdint.h or int64 -- if the compile breaks it will be obvious.  I'm not interested in enabling environments where sane definitions of int64_t, or uint64_t, do not exist, nor do I care to support those without stdint.h.

JackDunaway:
> punt on explicit checks for stdint.h or int64

See: https://github.com/nanomsg/nanomsg/pull/636/commits/95b664f66e89dfbcc99877d3132a54b420c69cf5

PR #641, opened 2016-04-18
fixes #574 Migrate to CMake for all environments

### LINKED ISSUE #574: Cross-platform support for CMake builds

JackDunaway:
This pull request is meant to unify all platforms into a single CMake build system, in order to eventually deprecate the autotools toolchain.

This PR is the second attempt at squashing and rebasing CMake work from several branches that has diverged from mainline for several months, and includes feedback discussed on GitHub from @gdamore, @Snaipe, and @shiva. Discussion may be found here: https://github.com/nanomsg/nanomsg/pull/636

Substantial portions of this commit are attributed to both @shiva and @Snaipe, whom have both expressed their understanding squashing all work into a single commit while wrangling together this work from disparate sources. Thank you kindly for helping on this!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,28 @@
 language: c
 sudo: false
-compiler:
-- clang
-- gcc
+matrix:
+  include:
+    - os: linux
+      compiler: gcc
+    - os: linux
+      compiler: clang
+#    - os: osx
+#      compiler: gcc
+    - os: osx
+      compiler: clang
 script:
-- ./autogen.sh
-- ./configure
-- make distcheck
+  # Print all environment variables to aid in CI development
+  - printenv
+  # Print version and available CMake generators to aid in CI development
+  - cmake --version
+  - cmake --help
+  # Perform out-of-source build
+  - mkdir build
+  - cd build
+  # Perform CMake backend generation, build, and test
+  - cmake ..
+  - cmake --build .
+  - ctest --output-on-failure -C Debug
 deploy:
   provider: releases
   api_key:
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,7 @@
 #
 #   Copyright (c) 2012 Martin Sustrik  All rights reserved.
 #   Copyright (c) 2013 GoPivotal, Inc.  All rights reserved.
+#   Copyright (c) 2015-2016 Jack R. Dunaway. All rights reserved.
 #   Copyright 2016 Garrett D'Amore <garrett@damore.org>
 #
 #   Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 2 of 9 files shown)


JackDunaway:
@gdamore -- this PR is not perfect, but it incorporates most of the feedback today from yourself, @Snaipe, and @shiva.

I say we go ahead and mainline as long as this passes the smoke test, which gives us all the opportunity to prune our disparate branches. Further, it'll enable targeted, documented changes, rather than trying to lump them all into a single "switchover".

That said ... I'll be hands-off for long enough for you to review and merge, so feel free to make edits!

@gdamore -- the final bit is to update your AppVeyor console to point toward `.appveyor.yml` (notice the leading period)

gdamore:
Looks like I need to enable .appveyor.yml.

src/CMakeLists.txt
```diff
@@ -271,20 +240,133 @@ set (NN_SOURCES
 )
 
 if (WIN32)
-    LIST (APPEND NN_SOURCES
```
gdamore: I'm still pretty opposed to this approach.  Part of the reason is that it will be easy for folks to think they're modifying the file list for all platforms, and only be modifying them for one of them.  I agree we can refactor to better than what we had before this change,but I don't think this part of the change set represents an improvement.

src/CMakeLists.txt
```diff
@@ -271,20 +240,133 @@ set (NN_SOURCES
...
-install(TARGETS nanomsg
+if (WIN32)
+    target_link_libraries (${PROJECT_NAME} ws2_32)
+    target_link_libraries (${PROJECT_NAME} mswsock)
+    target_link_libraries (${PROJECT_NAME} advapi32)
+elseif (UNIX)
+    if(THREADS_HAVE_PTHREAD_ARG)
+        add_definitions (-pthread)
+    endif()
+    if(CMAKE_THREAD_LIBS_INIT)
```
gdamore: What's this -- I've never seen this before.

src/CMakeLists.txt
```diff
@@ -271,20 +240,133 @@ set (NN_SOURCES
...
+        SOVERSION "${NN_ABI_VERSION}")
+endif ()
 
-install(TARGETS nanomsg
+if (WIN32)
+    target_link_libraries (${PROJECT_NAME} ws2_32)
+    target_link_libraries (${PROJECT_NAME} mswsock)
+    target_link_libraries (${PROJECT_NAME} advapi32)
+elseif (UNIX)
+    if(THREADS_HAVE_PTHREAD_ARG)
```
gdamore: Is this a standard CMake test?

.travis.yml
```diff
@@ -1,12 +1,28 @@
...
+      compiler: clang
 script:
-- ./autogen.sh
-- ./configure
-- make distcheck
+  # Print all environment variables to aid in CI development
+  - printenv
+  # Print version and available CMake generators to aid in CI development
+  - cmake --version
+  - cmake --help
```
gdamore: This one ought to be commented out by default -- understand its useful during development....

gdamore:
We're getting close... I really however am pretty opposed to the changes that required you to modify the .c, .h, and .inc files.   I don't like putting this complexity in the CMake configuration -- I want the CMake machinery to be as simple as we can make it, because the subset of people who work on this code who understand CMake is going to be much less than the set who grok the C preprocessor (or so I postulate.)

I've updated .appveyor, and I'd like to take this PR and run with it myself, if thats ok with you?  I will probably revert the part that modifies the .c, .h, etc.

.travis.yml
JackDunaway: Agree; it's too verbose for the low amount of information it's used for. (as a comparison, `printenv` is as verbose, but pretty useful, and is on the borderline of keeping un-commented ... )

src/CMakeLists.txt
JackDunaway: > it will be easy for folks to think they're modifying the file list for all platforms, and only be modifying them for one of them

This specifically is a weak argument, but I'm generally OK with removing these changes in favor of a later, more sustainable rearchitecture.

Keep in mind though -- there *are* functional changes in the *.h and *.c files in this PR -- notably, the order of precedence into which `#if` a given system would fall. And the functional change within `nn.h` is simply to support a previously unsupported feature of CMake, and that is building nanomsg as a static lib on Windows.

src/CMakeLists.txt
JackDunaway: I'm >80% sure that this -- and the question about the CMake test just above -- are a port of this code from autotools: https://github.com/nanomsg/nanomsg/blob/master/configure.ac#L237-L241

JackDunaway:
> I want the CMake machinery to be as simple as we can make it, because the subset of people who work on this code who understand CMake is going to be much less than the set who grok the C preprocessor

Agree with this mostly ... the only contention is when divergence between platforms becomes so pronounced that it's better to account for the differences within CMake -- as has been done here. But I'm 100% OK with refactoring those areas first, and then judge whether they are divergent enough to conditionally include within CMake.

The tipping point is this -- are the frequency and expanse of `#ifdef`'s big enough within the codebase to make the codebase sufficiently unreadable? And is the pain great enough to warrant learning a small amount of CMake to alleviate the pain? It all depends, i guess.

> take this PR and run with it myself, if thats ok with you?

By all means!! Let's get it mainlined. Ideally, create new commits on top of this that individually address each issue. All the squashing and continued divergence from mainline is what made this PR so inefficient in the first place.

JackDunaway:
PS -- AppVeyor is having a tough time with changes here. It's possibly due to the `CFG` being passed in (see: https://github.com/nanomsg/nanomsg/pull/641/files#diff-11c909939117928998b102a1fff7d363R6), and possibly due to the lack of `VS_VERSION` being passed as a variable (see last successful build from mainline, without new `.appveyor.yml` changes: https://ci.appveyor.com/project/gdamore/nanomsg/build/1.0.401)

It does not appear related to `GENERATOR` being passed in. Seems to be related to either `CFG` or `VS_VERSION`.

src/CMakeLists.txt
gdamore: So I found this.  Its pretty sad that this exists, but I understand why.  It is indeed standard.  But the answer is to use CMAKE_THREAD_LIBS_INIT instead of explicitly putting this here, I think.  But really what is going on is the CMake people have used the magical -pthread convenience switch used in legacy environments (old Solaris for example) instead of expanding this to -lpthread (or whatever) at link time, and -D_REENTRANT at compile time.

I suspect we ought to just blindly declare -D_REENTRANT everywhere; I'm not aware of any platform where this causes harm, but I am aware of several where it it makes things better.

JackDunaway: Definitely deferring to your expertise here; by all means, feel free to do as is fit.

JackDunaway:
> frequency and expanse of `#ifdef`'s

Here's the strategy I'd be happy with.

First, audit all preprocessor conditionals in all the codebase, and get it to where we feel it's pretty sustainable. This also means evaluating file boundaries (e.g., with the `*.inc` files) -- perhaps collapsing files into one, or in other cases splitting into more than one.

Next, once we feel the preprocessor is about as good as it could be, evaluate if/where it makes sense to deal with any burdensome complexity at the CMake level.

As it stands  -- we can't make decisions on what CMake should or shouldn't do, because we've not taken the time to first pay down principal on tech debt accumulated over >4yr+ of development. Once the cruft is cleared, then it's worthwhile to revisit this conversation.

src/CMakeLists.txt
JackDunaway: addendum: This type of "magic" of expanding macros is the reason I became so fed up with autotools that I decided to invest fully into CMake. But now, seeing such "magic" repeated in CMake, I'm 100% inclined to see it go away in favor of something more explicit, such as a real compiler or linker flag as you suggest. Especially, if it means avoiding parsing the esoterica of the developer mailing lists to understand fully what `CMAKE_THREAD_LIBS_INIT` even buys us.

On the other hand, I respect that the compiler and linker flags and combinations thereof are sometimes made of magic anyway, and if those gods require we must chant this incantation, then I understand that too. The value proposition of macros such as this is not needing to understand that magic.

Where do we draw the line of our own understanding of what is needed? That's your expertise I defer to.

gdamore:
Going back through this, and looking at both CMake and my earlier work on CMake, I'm noticing a serious problem.  We don't actually *use* most of the things we test for.  That is, we're not going to enable linking against libsocket, or -lrt, etc.

We are declaring variables in CMake, but not causing those variables to have any impact on the compile environment in many of these cases.

I have macros that fix this by wrapping checks with changes to cmake variables *that* do impact the environment. 

I'm going to get this stuff done, and I'll verify that that everything builds correctly on all of illumos (omnios), MacOS X, Linux (via travis), and Windows (via appveyor).

Stay tuned.

JackDunaway:
> AppVeyor is having a tough time with changes here

I'm going to spend some time figuring out what is wrong here on a separate AppVeyor instance, unless you've already got it figured out @gdamore ... else, stay tuned ...

gdamore:
I think it's load on appveyor.  It works much better in the evening but falls over during daytime peaks.

CMakeLists.txt
```diff
@@ -138,8 +303,10 @@ install (FILES src/pipeline.h DESTINATION include/nanomsg)
 install (FILES src/survey.h DESTINATION include/nanomsg)
 install (FILES src/bus.h DESTINATION include/nanomsg)
 
-install (TARGETS nanocat RUNTIME DESTINATION bin)
+if (BUILD_NANOCAT)
```
JackDunaway: @gdamore , I expect your commit to be the next on top of this PR -- would you fix this bug? Rather than `BUILD_NANOCAT` it simply needs to be `ENABLE_NANOCAT` -- or, whatever we decide the name of this option should be over on this discussion on #591

JackDunaway:
> load on appveyor

I [have suspected this also](https://github.com/nanomsg/nanomsg/pull/578#discussion_r47570203) in the past -- but yesterday's performance was interesting.

It seemed to fail all the time on this particular branch for this PR -- but a commit to mainline passed.

Is your account a [Pro account that runs on the 2core VM](https://www.appveyor.com/docs/build-configuration#virtual-machine-configurations)? If so, the difference is explainable, assuming AppVeyor is running `master` branch on the Pro hardware, and PR branches on the lower-powered free-plan.

JackDunaway:
AppVeyor failures are likely due to a bug in CMakeLists.txt ... I'm troubleshooting this now ...

Snaipe:
By the way, If appveyor builds are slow to start (which seems to be the case reading the build times), you should open a ticket on their [support channel](http://help.appveyor.com/). I had issues with it (37 minutes builds where 7 minutes of which were actual work), and they upgraded my environment for free. Now, all of my builds start instantly and are performing quite fast.

JackDunaway:
@Snaipe, good tip.

And back to the failures here -- it's definitely a problem in the script itself -- the test executables are not properly linking to the nanomsg shared library. The failure mode on Windows manifests as a timeout, due to an unacknowledged dialog.

I'm looking into the most sustainable solution here for setting up the binary outputs so that they link properly. It's interesting that this failure doesn't happen on Travis ... at this time, I suspect it's a difference in the behavior of the CMake versions (2.8.7 on Travis, and 3.5.0 on AppVeyor)

src/CMakeLists.txt
```diff
@@ -350,6 +350,11 @@ else ()
         SOVERSION "${NN_ABI_VERSION}")
 endif ()
 
+# Set library outputs same as top-level project binary outputs
+set_target_properties (${PROJECT_NAME} PROPERTIES RUNTIME_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR})
+set_target_properties (${PROJECT_NAME} PROPERTIES LIBRARY_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR})
+set_target_properties (${PROJECT_NAME} PROPERTIES ARCHIVE_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR})
+
```
JackDunaway: @Snaipe, @gdamore, @shiva what do you think about this fix?

It's to resolve a linking problem on Windows that does not occur on *NIX systems. This bug is what has been causing AppVeyor to timeout and fail in all the tests on this PR.

Presumably, this bug is manifested from a bad interaction between `add_subdirectory` for `src/CMakeScripts.txt` and this output behavior unique to MSVC: http://stackoverflow.com/questions/7747857/in-cmake-how-do-i-work-around-the-debug-and-release-directories-visual-studio-2

It solves the problem by dumping the output `nanomsg.dll` (or `nanomsg.lib` for static libs) into the binary location specified in the parent `CMakeLists.txt` -- which is incidentally where all the test executables end up.

Even though this fix affects *NIX systems, I think it cleans up the structure of the artifacts a bit nicer (opinion), and for this reason didn't wrap this block in a `if (MSVC)` statement.

The other reason for doing it like this is to scope changes as narrowly as possible -- unlike the previous `src/CMakeScripts.txt` still currently in mainline that changed globally-scoped `CMAKE_...` variables -- https://github.com/nanomsg/nanomsg/blob/master/src/CMakeLists.txt#L24-L26

Thoughts on this patch? Any gotchas? Or a better way?

---

And two semi-related ideas I had while solving this issue:
1. collapse `src/CMakeScripts.txt` into `CMakeLists.txt`, obviating need for `add_subdirectory` and its scoping headaches (especially, in the bigger context of when nanomsg is a subproject of another project), or ...
2. ... going the opposite direction, and creating two more `CMakeLists.txt` for both the `tests` and `tools` directory, both for overall consistency, and also to meet composability problems such as these head-on by setting precedent of best practices on how to do so.

Thoughts here?

Snaipe: @JackDunaway Aaah, I haven't thought of that solution. What I usually do for my projects is to add the runtime output directory to the `PATH` (or install the library files in a known subdirectory and add the `bin/` directory to the path, which is what I currently do to in a project where I need to build and link multiple cmake dependencies). This has the advantage of working regardless of the layout and number of targets that I create.

I would be in favor of having multiple CMakeLists. When I started using CMake, I thought that having one CMakeLists per subdirectory was dumb, but now that I manage larger projects, I kind of see the point: the main CMakeLists gets easily cluttered and harder to read the more you add to it. This is mostly a question of aethetics and readability, more than functionality.

gdamore:
Quite a bit further work was done on this, and the result was squashed and merged:

commit 5b34eb6467975ff4c444e2f8e1267bf01aba4b36
Date:   Sat Apr 16 15:41:44 2016 -0500

    fixes #647 Initial cmake support for non-Windows platforms
    
    This change is the first step towards a unified CMake based build for
    all nanomsg platforms.  With this it is possible to build nanomsg
    using cmake on at least Linux, MacOS X, illumos, and Windows systems.
    However, there remains work to be done, particularly with respect to
    delivery of documentation and the nanocat binary.
    
    The eventual goal will be to remove support for the autotools, and
    settle on a single cmake based toolchain.
    
    This work is the result of a collaborative effort by several authors:
    Jack R. Dunaway (@JackDunaway), Garrett D'Amore (@gdamore),
    Franklin Mathieu (@Snaipe) and Shiva (@shiva).

gdamore:
Note that specifically #574 remains open because the work is not yet complete.

PR #642, opened 2016-04-18
fixes #640 clock_rdtsc is unsafe

### LINKED ISSUE #640: clock_rdtsc is unsafe
nanomsg's use of the TSC is not safe.  In particular, the TSC may not be supported on many systems, and even on those where it is supported there may be factors to consider in using it:

a) On older systems, the TSC is not "invariant" -- that is, it may not tick at a fixed rate.  In particular, the TSC may increment slower or faster depending on adjustments made to the system clock in response to power management.  Early Opterons had this problem.

b) On multicore / multicpu systems, the TSC may not be synchronized across all cores.  This can lead to surprising results when a process is rescheduled on a different core or cpu.

c) ON systems with dynamic reconfiguration of CPUs (rare!) the TSC won't be synchronized

Note that modern CPUs and systems probably don't have this problem -- usually the TSCs are all reinitialized at system start, and generally we don't see DR.  But nonetheless, the naive assumptions about the TSC are foolish.  To do this work correctly, you really need to look at the CPUID and check for a bunch of different things.  Time related code in UNIX kernels does this usually.

This is one more case of what I refer to as "premature optimization".  The reads performed of system clock time (generally equivalent to clock_gettime() or gethrtime()) are generally light weight.  (Sadly OSX needs special help here, it lacks a sane system timer except through mach-specific incantations) generally execute very quickly, using a "fast" system call (fasttrap or similar), or (as in the case of Linux) even just use a read of a shared kernel page.

The use of a TSC-specific timer is justified when submicrosecond resolution is required.  That's not our case.  We have millisecond resolution, and can make use of standardized services instead of inventing something special here.

Its entirely possible that this logic is the root of several problems seen working with multiCPU systems -- in virtualized environments its likely that the TSC is not entirely reliable.

gdamore:
While here, we can simplify the clock calls, since we don't need any state per clock.  Just return an absolute number of msec since arbitrary point in the past.

The main point of this change is to remove the broken premature optimizations around the TSC.  The operating systems in question are expected to provide sane, efficient, ways to determine the number of milliseconds since a point in the past (typically system boot).  Furthermore, we don't really need to have super accurate data here -- system clock granularity is more than sufficient since we aren't profiling or using this for super-resolution event timing; indeed even the bad case of 16 msec resolution some Windows systems have with GetTick64 would probably be sufficient for our needs.  (That said, we use QueryPerformanceCounter instead -- mostly because its more widely available than GetTick64.)

### DIFF
--- a/src/aio/timerset.c
+++ b/src/aio/timerset.c
@@ -24,18 +24,17 @@
 
 #include "../utils/fast.h"
 #include "../utils/cont.h"
+#include "../utils/clock.h"
 #include "../utils/err.h"
 
 void nn_timerset_init (struct nn_timerset *self)
 {
-    nn_clock_init (&self->clock);
     nn_list_init (&self->timeouts);
 }
 
 void nn_timerset_term (struct nn_timerset *self)
 {
     nn_list_term (&self->timeouts);
-    nn_clock_term (&self->clock);
 }
 
 int nn_timerset_add (struct nn_timerset *self, int timeout,
@@ -46,7 +45,7 @@ int nn_timerset_add (struct nn_timerset *self, int timeout,
     int first;
 
     /*  Compute the instant when the timeout will be due. */
-    hndl->timeout = nn_clock_now (&self->clock) + timeout;
+    hndl->timeout = nn_clock_ms()  + timeout;
 
     /*  Insert it into the ordered list of timeouts. */
     for (it = nn_list_begin (&self->timeouts);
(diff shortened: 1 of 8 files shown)


sustrik:
IIRC this way of doing things dates back to mid 2000's when we had problems with slow gettimeofday() on some (older?) systems. Good to hear it's not needed anymore.

gdamore:
Gettimeofday is still slow. But rarely needed as arbitrary times are much easier to collect. Gethrtime and clock_gettime(CLOCK_MONOTONIC) are fast. So is QueryPerformanceCounter.  These actually use the TSC when it is reasonable to do so.

gdamore:
rebased as: 

commit 4b88a3b68ed1bdd4c57ba80e2a52b31a4846831f
Date:   Mon Apr 18 16:32:35 2016 -0700

    fixes #640 clock_rdtsc is unsafe

Issue #639, opened 2016-04-18
Warnings seen in AppVeyor build


gdamore:
We should fix these... there are a number of warnings exposed in appveyr:

<pre>
"C:\projects\nanomsg\build\src\nanomsg.vcxproj" (default target) (4) ->
(ClCompile target) -> 
  C:\projects\nanomsg\src\core\global.c(176): warning C4101: 'addr': unreferenced local variable [C:\projects\nanomsg\build\src\nanomsg.vcxproj]
  c:\projects\nanomsg\src\aio\usock_win.inc(353): warning C4267: 'function': conversion from 'size_t' to 'int', possible loss of data [C:\projects\nanomsg\build\src\nanomsg.vcxproj]
  c:\projects\nanomsg\src\aio\usock_win.inc(412): warning C4267: 'function': conversion from 'size_t' to 'DWORD', possible loss of data [C:\projects\nanomsg\build\src\nanomsg.vcxproj]
  c:\projects\nanomsg\src\aio\usock_win.inc(462): warning C4267: 'function': conversion from 'size_t' to 'DWORD', possible loss of data [C:\projects\nanomsg\build\src\nanomsg.vcxproj]
  C:\projects\nanomsg\src\utils\efd.c(67): warning C4244: 'initializing': conversion from 'SOCKET' to 'int', possible loss of data [C:\projects\nanomsg\build\src\nanomsg.vcxproj]
  C:\projects\nanomsg\src\transports\ws\cws.c(173): warning C4244: '=': conversion from '__int64' to 'int', possible loss of data [C:\projects\nanomsg\build\src\nanomsg.vcxproj]
  C:\projects\nanomsg\src\transports\ws\sws.c(580): warning C4267: '=': conversion from 'size_t' to 'int', possible loss of data [C:\projects\nanomsg\build\src\nanomsg.vcxproj]
  C:\projects\nanomsg\src\transports\ws\ws_handshake.c(489): warning C4267: '=': conversion from 'size_t' to 'unsigned int', possible loss of data [C:\projects\nanomsg\build\src\nanomsg.vcxproj]
  C:\projects\nanomsg\src\transports\ws\ws_handshake.c(668): warning C4267: '=': conversion from 'size_t' to 'unsigned int', possible loss of data [C:\projects\nanomsg\build\src\nanomsg.vcxproj]
 
 
"C:\projects\nanomsg\build\ALL_BUILD.vcxproj" (default target) (1) ->
[...]

PR #643, opened 2016-04-19
Removing two unused variables (two warnings from #639)

JackDunaway:
Related to #639, this commit removes two unused variables caught by the MSVC compiler on AppVeyor: https://ci.appveyor.com/project/gdamore/nanomsg/build/0.8.410/job/09bx1jujs4bsqnqb#L1421


gdamore:
Looks good.  I think you should do this as your first integration directly into the repo - I will touch base on gitter to work it through.

Issue #588, opened 2016-01-13
Dolt breaks libtool PIC options


Snaipe:
Currently, the Dolt macros completely break the behaviour of libtool PIC options (`--(with|without)-pic` and `pic-only`). This makes it impossible to link the static nanomsg library into shared libraries.

(I remember discussing with @gdamore on gitter that the autotools build system was to be replaced by cmake on all platforms eventually, so I'm not sure if I should submit a patch for dolt.m4)

wirebirdlabs:
Hi @Snaipe -- I don't have an answer, but can offer some related information.

When working on bringing CMake chain up to the same capability as the Autotools build chain, I noticed something similar -- that nanomsg was not yet designed to switch back/forth to shared/static library at the configuration stage.

Here's a commit that fixes this for Windows: https://github.com/wirebirdlabs/featherweight-nanomsg/commit/ce4c5d353b9e22f4e7f0e7620095724cc72c8b61

That said, would you be willing to switch to CMake instead of Autotools and submit any patches there? If so, I would greatly appreciate your help so we could mainline that work. Here are more details: https://github.com/nanomsg/nanomsg/issues/574

And here's a preview of the CMake script: https://github.com/wirebirdlabs/featherweight-nanomsg/compare/pr-ws-mega-squash...wirebirdlabs:ftw-kernel#files_bucket

Snaipe:
This seems fair. A CMake setup would be easier to deal with on my end, so I see no issue switching to it at the moment.

wirebirdlabs:
Cool! What is your timeline for needing this? I will try to create a cogent PR within the next couple days that puts together this CMake work, from which you can work.

If you want to start before that, feel free to cherry-pick and continue making forward progress, but be forewarned that the PR will have squashed most of those commits into a single commit (meaning, if you work from these, your branch wouldn't cleanly fast-forward from what would eventually be mainlined, though your work should be able to be cherry-picked cleanly onto mainline)

Snaipe:
I can wait a couple of days, no worries.

gdamore:
So definitely we're gonna do CMake.  Probably this can be closed when we have it.

gdamore:
Initial cmake work is done, and we're moving towards that as fast as we can.  I'm closing this issue as "WILL NOT FIX" -- because just use cmake. :-)

JackDunaway:
See: https://github.com/nanomsg/nanomsg/issues/648

PR #652, opened 2016-04-20
Added cmake option to build a static library using position-independent code

Snaipe:
This is a follow-up of #588. Autotools has the concept of "convenience libraries", i.e. static library built with position-independent code; and unfortunately, CMake doesn't. This is a problem, because it becomes impossible to link a the static nanomsg library into a shared library.

This PR fixes this by introducing a compiler check for `-fPIC` and adding it to the CFLAGS if found.

Because position-independent code can have some overhead, the PR also introduces the `NN_ENABLE_PIC` option to enable/disable this behaviour. The option is also on by default, because I think that being able to Just Work™ is more important by default than having an almost not noticeable overhead, and applications that absolutely *need* to disable that can still do so in their build.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,6 +29,7 @@ include (CheckSymbolExists)
 include (CheckStructHasMember)
 include (CheckLibraryExists)
 include (CheckCSourceCompiles)
+include (CheckCCompilerFlags)
 
 project (nanomsg C)
 enable_testing ()
@@ -85,6 +86,7 @@ endif()
 option (NN_ENABLE_NANOCAT "Enable building nanocat utility." ON)
 option (NN_STATIC_LIB "Build static library instead of shared library." OFF)
 option (NN_ENABLE_GETADDRINFO_A "Enable/disable use of getaddrinfo_a in place of getaddrinfo." ON)
+option (NN_ENABLE_PIC "Enable position-independent code when building static library" ON)
 
 #  Platform checks.
 
@@ -171,6 +173,8 @@ nn_check_sym (CLOCK_MONOTONIC time.h NN_HAVE_CLOCK_MONOTONIC)
 nn_check_sym (atomic_cas_32 atomic.h NN_HAVE_ATOMIC_SOLARIS)
 
 nn_check_struct_member(msghdr msg_control sys/socket.h NN_HAVE_MSG_CONTROL)
+
+check_c_compiler_flag ("-fPIC -Werror" NN_HAVE_FPIC)
 
 if (NN_HAVE_SEMAPHORE_RT OR NN_HAVE_SEMAPHORE_PTHREAD)
     add_definitions (-DNN_HAVE_SEMAPHORE)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -340,6 +340,10 @@ foreach (f ${NN_SOURCES})
 endforeach ()
 
 if (NN_STATIC_LIB)
+    if (NN_HAVE_FPIC AND NN_ENABLE_PIC)
+        set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fPIC")
+    endif ()
+
     add_library (${PROJECT_NAME} STATIC ${NN_SOURCES})
     add_definitions (-DNN_STATIC_LIB)
 else ()


JackDunaway:
I recall ironing out some of the same issues at one point, but ended up with a simpler solution using a single line with `CMAKE_POSITION_INDEPENDENT_CODE`: https://github.com/featherweight/ftw-kernel/blob/master/CMakeLists.txt#L31

Thoughts between this macro, versus explicitly specifying flags? I've not studied the differences recently or deeply enough to have an opinion on what's best here.

JackDunaway:
> CMAKE_POSITION_INDEPENDENT_CODE

A very brief search suggests this macro may have been added in 2.8.9 (citation needed), and therefore is not compatible with our >=2.8.7 goal.

Snaipe:
As you mentioned, it's mostly a compatibility issue. I could set `CMAKE_POSITION_INDEPENDENT_CODE` iff cmake >= 2.8.9, and fallback to the current `-fPIC` check otherwise, though.

JackDunaway:
I'm ambivalent on casing based on CMake version with the polyfill. Is the complexity worth it? Perhaps. The question is what `CMAKE_POSITION_INDEPENDENT_CODE` actually buys us beyond this one compiler flag `-fPIC`, especially in terms of forward compatibility and CMake's ability to smooth over subtle platform/compiler differences. (I don't know the answer to this question)

A new point of feedback while looking at the code again ... rather than setting `CMAKE_C_FLAGS`, would it be better to set compile flags with `set_target_properties` for the library target only?

gdamore:
I'd argue in favor of using the CMAKE flag, and basically insist that folks who want that capability have to use "modern" cmake. 

The 2.8.7 goal for cmake always felt really really arbitrary to me, and we're not going to be stuck there forever.

CMakeLists.txt
```diff
@@ -172,6 +174,8 @@ nn_check_sym (atomic_cas_32 atomic.h NN_HAVE_ATOMIC_SOLARIS)
 
 nn_check_struct_member(msghdr msg_control sys/socket.h NN_HAVE_MSG_CONTROL)
 
+check_c_compiler_flag ("-fPIC -Werror" NN_HAVE_FPIC)
+
```
gdamore: This starts a precedent I want to avoid.  The compiler flags knowledge can get really hairy -- especially with some of the flags for optimizations, different debugging formats, etc.  And there are reasonable compilers that have different flags for PIC.  E.g. SunPro uses -KPIC.  So, let's just use the cmake magic to enable this instead, please, rather than carrying this baggage around ourselves.

gdamore:
CMAKE_POSITION_INDEPENDENT_CODE supports e.g. SunPro compilers with -KPIC, etc.  Its not desirable for us to start duplicating the effort that the cmake folks have put (and continue to put) into these things -- our job is big enough.

JackDunaway:
> I'd argue in favor of using the CMAKE flag, and basically insist that folks who want that capability have to use "modern" cmake.

This seems incredibly reasonable. I like it the best of all the options so far, actually.

> The 2.8.7 goal for cmake always felt really really arbitrary to me, and we're not going to be stuck there forever.

It's not arbitrary at all, from our perspective -- it's what Travis CI ships with out of the box, because it's what Ubuntu 12.04 LTS is using. OK, perhaps it's "somewhat" arbitrary for us to abide by this, but at least a BIG stake in the ground -- an important service provider in our CI infrastructure depends upon it.

Another not-terrible option is to re-introduce code that downloaded CMake 2.8.12 a pre-build step. The cost here is a few lines in `.travis.yml`, about +15sec per build, and another potential point of maintenance/failure.

gdamore:
Travis won’t stay at 2.8.7 forever, and 12.04 LTS is was released in 2012 —
thats four years ago.  It has another year until EOL, but 14.04LTS is now
two years old and 16.04LTS is due out tomorrow.

I don’t feel particularly constrained by the older releases — especially
given that upgrading CMake is pretty darn easy.   Here’s the status of
cmake in ubuntu: https://launchpad.net/ubuntu/+source/cmake

We have not, for example, done any analysis of what’s in centos or redhat,
or suse.  Never mind the fact that CMake isn’t even installed by default on
MacOS X or Windows.

gdamore:
Btw, I’m in favor of delivering official pre-packaged binaries (.DEB, .RPM,
etc.) using cpack to build, so that Joe Developer can just install without
having to worry about cmake etc.

JackDunaway:
> delivering official pre-packaged binaries (.DEB, .RPM, etc.) using cpack to build

Hearty +1. Further, this should be a gate before 1.0, and take precedence over the "here's how to build from source" instructions.

JackDunaway:
> We have not, for example, done any analysis of what’s in centos or redhat, or suse

Arguably, we have indeed done the analysis -- by omission. Those environments are not represented in the officially-supported automated CI, therefore they aren't quite as big a stake in the ground. By not officially building these environments, that alone is a statement whether they should/shouldn't be a dominant driving force for dependencies such as CMake version.

Should they be built? I'd love to see more environments added to the automated build; continuiing to expand in this manner has flushed many unique potential issues, given a widening variety of compilers and environments.

How easy is it to expand here? There's http://build.nanomsg.org/waterfall, which is as successful as its maintainers' maintenance (contrasted against, the outsourced maintenance of AppVeyor and Travis).

So, the analysis has been done, where 2.8.7 and 2.8.12 are neck-and-neck today, both appreciably above 3.x -- but as you mention, time elapses, and it's a moving target.

JackDunaway:
> Here’s the status of cmake in ubuntu: https://launchpad.net/ubuntu/+source/cmake

Very true ... but again, the limitation here is Travis CI. Can we install that package without `sudo` in a Travis environment today? Perhaps tomorrow; but today, we can't. (Would love to be proven otherwise!)

gdamore:
Should install just fine - just in a different install directory because you lack permission to override the existing one. 

Also we might want to see what circle CI has.  There's been some suggestion that they can be a better choice.   I've not looked.

Snaipe:
> Travis won’t stay at 2.8.7 forever, and 12.04 LTS is was released in 2012 — thats four years ago. It has another year until EOL, but 14.04LTS is now two years old and 16.04LTS is due out tomorrow.

Fair enough. I would argue that Travis is probably the most popular CI service out there, so 2.8.7 compatibility was a huge convenience for the majority of users that build nanomsg from source, but Travis also has a switch to use their beta 14.04 LTS worker instead of the default 12.04 LTS one, so it still should be possible to use the CMAKE switch.

In that case, there is no need to merge this. CMAKE_POSITION_INDEPENDENT_CODE is mostly meant to be passed as an option at configure time directly -- adding another round of indirection is unecessary.

PR #654, opened 2016-04-21
Added more builds to the travis matrix and enabled warnings

Snaipe:
Enabling all "reasonable" warnings is mostly here for the long-term quality of the project, and should be the default for all automated builds.

I also added per-configuration entries to the build matrix to catch any mishaps with and without optimisation settings.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,43 @@
 language: c
 sudo: false
+
+_environments:
+  - &gcc-warnings
+    - CFLAGS="-std=c89 -Wall -Wextra -pedantic"
+
 matrix:
   include:
+  # Linux
+    # GCC
     - os: linux
       compiler: gcc
+      env: [CONFIGURATION=Debug, *gcc-warnings]
+    - os: linux
+      compiler: gcc
+      env: [CONFIGURATION=Release, DEPLOY=true, *gcc-warnings]
+    # Clang
     - os: linux
       compiler: clang
-#    - os: osx
-#      compiler: gcc
+      env: [CONFIGURATION=Debug, *gcc-warnings]
+    - os: linux
+      compiler: clang
+      env: [CONFIGURATION=Release, *gcc-warnings]
+  # OS X
+    # GCC 4.9
+    - os: osx
+      compiler: gcc-4.9
+      env: [CONFIGURATION=Debug, *gcc-warnings]
+    - os: osx
+      compiler: gcc-4.9
+      env: [CONFIGURATION=Release, *gcc-warnings]
+    # AppleClang
     - os: osx
       compiler: clang
+      env: [CONFIGURATION=Debug, *gcc-warnings]
+    - os: osx
+      compiler: clang
+      env: [CONFIGURATION=Release, *gcc-warnings]
+
 before_script:
   - gem install asciidoctor
 script:
@@ -22,7 +50,7 @@ script:
   - mkdir build
   - cd build
   # Perform CMake backend generation, build, and test
-  - cmake ..
+  - cmake -DCMAKE_BUILD_TYPE=${CONFIGURATION} ${CMAKE_OPTS} ..
   - cmake --build . -- -j4
   - ctest --output-on-failure -C Debug -j4
 #deploy:
@@ -35,5 +63,6 @@ script:
 #    - nanomsg-${TRAVIS_TAG}.tar.bz2
 #  skip_cleanup: true
 #  on:
+#    condition: $DEPLOY = true
 #    tags: true
 #    repo: nanomsg/nanomsg
--- a/.travis.yml
+++ b/.travis.yml
@@ -74,13 +74,14 @@ script:
   # Print version and available CMake generators to aid in CI development
   - cmake --version
   - cmake --help
-  # Perform out-of-source build
-  - mkdir build
-  - cd build
-  # Perform CMake backend generation, build, and test
-  - cmake -DCMAKE_BUILD_TYPE=${CONFIGURATION} ${CMAKE_OPTS} ..
-  - cmake --build . -- -j4
-  - ctest --output-on-failure -C Debug -j4
+
+  # Perform out-of-source CMake backend generation, build, and test
+  - |
+      mkdir build && cd $_ &&
+      cmake -DCMAKE_BUILD_TYPE=${CONFIGURATION} ${CMAKE_OPTS} .. &&
+      cmake --build . -- -j4 &&
+      ctest --output-on-failure -C Debug -j4 ${CTEST_OPTS}
+
 #deploy:
 #  provider: releases
 #  api_key:
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,21 @@ _environments:
   - &gcc-warnings
     - CFLAGS="-std=c89 -Wall -Wextra -pedantic"
 
+_scripts:
+  - &install_cmake35
+    - |
+        if [ ! -f $HOME/bin/cmake ]; then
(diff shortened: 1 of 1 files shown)


gdamore:
In the past I made a conscious decision not to enable so many builds -- mostly because it greatly increases the time to get a result.  Eventually we will want these all, but I don't think we are quite there yet...

Snaipe:
@gdamore Builds gets processed 6 by 6 (except for OS X ones), so It shouldn't matter here. I can remove on the OS X side RelWithDebInfo if you don't plan to release binaries with debug symbols, or Release if you don't care about release binaries without those symbols.

Snaipe:
Also, if speed is a problem, I could allow failures on all OS X builds and put a `fast_finish: true` in the matrix, since they are the ones mostly slowing the build down.

Of course, the main issue being that if OS X builds fail, the build still gets reported as passing (if the linux builds are passing).

JackDunaway:
One thing that should definitely happen -- if the build fails, we should skip the `ctest` (and everything else downstream). It's not working like this right now -- e.g., https://travis-ci.org/nanomsg/nanomsg/jobs/124774212#L384

Snaipe:
@JackDunaway regarding your comment about aborting if any command fails, I tried inserting a `set -e`, but it seems to mess up with the build results (with the job marked as errored rather than failing), so it seems that in order to abort the script early, you need to either put your actual build commands in a script with `set -e`, or group everything as one `sh -ec`, or chain everything with `&&`.

JackDunaway:
`&&` chaining might be the way to go -- here's one precedent: https://github.com/travis-ci/travis-ci/issues/1066

Such syntax doesn't look too bad:
```
   # Perform CMake backend generation, build, and test
   - cmake -DCMAKE_BUILD_TYPE=${CONFIGURATION} ${CMAKE_OPTS} .. &&
     cmake --build . -- -j4 &&
     ctest --output-on-failure -C Debug -j4 ${CTEST_OPTS}
```

.travis.yml
```diff
@@ -5,7 +5,15 @@ _environments:
   - &gcc-warnings
     - CFLAGS="-std=c89 -Wall -Wextra -pedantic"
 
+_scripts:
+  - &install_cmake35
+    - >
+        wget https://cmake.org/files/v3.5/cmake-3.5.2-Linux-x86_64.tar.gz &&
+        sudo tar xfC $(basename $_) / --strip-components=1
+
```
JackDunaway: Use of `sudo` here doesn't come with zero cost, falling back to standard infrastructure rather than what is reported/observed as [faster container-based VM](https://docs.travis-ci.com/user/workers/container-based-infrastructure/).

Is it worth it? It seems so, especially with this new use of `repeat-until-fail`

We can configure some environments to respond quickly (<1min feedback is common in Travis on a Pull Request), and then allow additional environments configured as below to continue with more strenuous tests.

A curiosity:

3.5 downloaded: https://travis-ci.org/nanomsg/nanomsg/jobs/124985033#L168
3.2 reported: https://travis-ci.org/nanomsg/nanomsg/jobs/124985033#L275

Snaipe: Right -- speed is a problem, which is why I set those two tests in `allow_failures`, and set `fast_finish` to true. This has the effect that fast, container-based builds determine the global result of the test, and doesn't undermine the ability to get builds results fast -- slower builds can work in the background without clogging the job queue.

JackDunaway: Wow, good call all around. I see even now that the overall test has been reported, even as these stress-testers are still cranking away. Best of all worlds.

.travis.yml
```diff
@@ -38,6 +52,20 @@ matrix:
...
+      cache: *cache_homedir
+    - compiler: gcc
+      env: CONFIGURATION=Release  CTEST_OPTS="--repeat-until-fail 25 --schedule-random"
+      install: *install_cmake35
+      cache: *cache_homedir
+
+  allow_failures:
+    - env: CONFIGURATION=Debug    CTEST_OPTS="--repeat-until-fail 25 --schedule-random"
+    - env: CONFIGURATION=Release  CTEST_OPTS="--repeat-until-fail 25 --schedule-random"
+
```
JackDunaway: There must be a better way to constrain `env` and pass parameters to `cmake` and `ctest`; any ideas?

Snaipe: I'm not sure to get what you're saying, are you asking if there's a way to avoid duplicating the full `env` entry?

If so, I don't think it's currently possible unfortunately. I originally added a `ALLOW_FAIL=yes` variable in both builds and put it as a condition for allowed failures, but it wasn't recognized. It seems you have to put the whole environment array as a condition.

What I could do, in fact, is cheat a little. I could specify a `compiler: gcc-failure` key in both jobs, then re-export `CC` as `gcc` rather than `gcc-failure`, and use the compiler key as the condition.

This is unfortunate hackery for what should be normal behaviour.

JackDunaway: Yes, the intent was to be able to avoid duplicating the full `env` entry. Is it possible to introduce a new K/V pair into the matrix, say called `stress_test: true` or `failure_allowed: true`, that then becomes a definition for `allow_failures`? I've not used/studied Travis enough to know what options are available for specifying pattern matching for `allow_failures`.

JackDunaway: (...searching...) This is two years outdated, but supports your current method: http://stackoverflow.com/a/20909996 Ideally, it's outdated and better now??

Snaipe: > Is it possible to introduce a new K/V pair into the matrix, say called stress_test: true or failure_allowed: true, that then becomes a definition for allow_failures?

AFAIK, no -- hence the hack using `compiler`.

It might be interesting to submit an issue ticket on travis-ci to ask for such a feature, or even better, to allow for a per-job `allow_failure: true` key.

JackDunaway: The per-job is exactly a path I was searching for. This makes the most sense, but it doesn't look like it exists right now.

What about pre-processor expansion? For example, is there such a syntax where we could declare `STRESSTEST="CONFIGURATION=Debug    CTEST_OPTS=\"--repeat-until-fail 25 --schedule-random\""` at the top of the YML definition, and then define lines 57, 61, 66, and 67 above as `$STRESSTEST`? (I'm not such a pre-processor or expansion exists; just thinnking aloud)

JackDunaway: To answer my own question -- it appears you've already figured out how to do this with the `scripts` -- yet I'm unable to find any documentation or similar uses of this trick on the web -- could the same concept be used for these `env` variables?

Snaipe: Oh, right. Those are named yaml anchors, and you could in fact define two anchors for the env keys.

Effectively, the result would be:

```yml
  # Sporadic failure builds
    - compiler: gcc
      env: &env_sporadic_dbg CONFIGURATION=Debug    CTEST_OPTS="--repeat-until-fail 25 --schedule-random"
      install: *install_cmake35
      cache: *cache_homedir
    - compiler: gcc
      env: &env_sporadic_rel CONFIGURATION=Release  CTEST_OPTS="--repeat-until-fail 25 --schedule-random"
      install: *install_cmake35
      cache: *cache_homedir

  allow_failures:
    - env: *env_sporadic_dbg
    - env: *env_sporadic_rel
```

Sounds good enough?

JackDunaway: That's much nicer! Also, I learned something new.

gdamore:
Can we get an update on this, and I will review?

gdamore:
nng has a much richer test matrix here... i'm probably going to punt on this for now, as I'm winding down investing on nanomsg itself.

gdamore:
At this point, I've decided to punt on this pending nng.  Sorry.

PR #659, opened 2016-04-21
fixes #603 crash if REP closes connection after sending protocol header

### LINKED ISSUE #603: crash if REP closes connection after sending protocol header
Observed behavior: `nanocat --req` crashes if the REP endpoint closes the socket right after sending the 8-byte protocol header.

Expected behavior: `nanocat --req` should open a new connection and send the request again.

Environment: Ubuntu Trusty, Linux 3.13, nanomsg 0.8-beta. Happens to both `tcp://` and `ipc://` addresses.

How to reproduce:

```
$ rm -f faulty.sock && python -c 'if 1:
    import socket
    ssock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM, 0)
    ssock.bind("faulty.sock")
    ssock.listen(16)
    while 1:
      sock, addr = ssock.accept()
      hdr = sock.recv(64)
      assert hdr == "\0SP\0\0000\0\0", repr(hdr)
      sock.sendall("\0SP\0\0001\0\0")
      sock.close()' &
$ while sleep .1; do nanocat --req --data d --connect ipc://faulty.sock; done
```

This makes `nanocat --req` abort with either of these error messages:
```
Unexpected source: state=5 source=1 action=5 (src/transports/utils/streamhdr.c:306)
Aborted (core dumped)
```
or:
```
Assertion failed: 7 == NN_USOCK_STATE_ACTIVE (src/aio/usock_posix.inc:420)
Aborted (core dumped)
```

The Python code above implements a custom REP server which closes the connection immediately after sending the correct 8-byte protocol header.

I can also reproduce the 2nd (NN_USOCK_STATE_ACTIVE) error message with `tcp://127.0.0.1:33221` (and a corresponding update to the REP server Python code).

gdamore:
Note that there is probably a similar error in the Windows code... I've not checked yet.

### DIFF
--- a/src/aio/usock_posix.inc
+++ b/src/aio/usock_posix.inc
@@ -1,7 +1,7 @@
 /*
     Copyright (c) 2013 Martin Sustrik  All rights reserved.
     Copyright (c) 2013 GoPivotal, Inc.  All rights reserved.
-    Copyright 2015 Garrett D'Amore <garrett@damore.org>
+    Copyright 2016 Garrett D'Amore <garrett@damore.org>
 
     Permission is hereby granted, free of charge, to any person obtaining a copy
     of this software and associated documentation files (the "Software"),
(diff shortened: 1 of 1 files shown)


gdamore:
rebased and merged: 
commit 98ee1c5795620767c2dc32b214737f9d7fb3c498
Date:   Thu Apr 21 14:31:48 2016 -0700

    fixes #603 crash if REP closes connection after sending protocol header

src/aio/usock_posix.inc
```diff
@@ -832,6 +838,13 @@ error:
         case NN_USOCK_SRC_FD:
             return;
 
+        case NN_FSM_ACTION:
+            switch (type) {
+            case NN_USOCK_ACTION_ERROR:
+                return;
```
JackDunaway: This raises a flag when reading the code; perhaps a comment, or even better `nn_assert()`, would help by explaining why an error condition results in a no-op?

PR #665, opened 2016-04-22
fixes #660 enable asciidoctor documenation generation in cmake

### LINKED ISSUE #660: enable asciidoctor documenation generation in cmake
As part of converting to cmake, we need to support generating the documentation via asciidoctor.

gdamore:
While here I did a couple of things:

1) Doc generation enabled by default if asciidoctor is present
2) relocate the definition of the project so that installdirs is correct
3) stop forcing NSIS -- CPack generator can be chosen as needed

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -2,19 +2,27 @@ version: 0.8.{build}
 environment:
   matrix:
     # array of all environments used to test builds
+    - GENERATOR: NMake Makefiles
+      CFG: Debug
+      VS_VERSION: 12.0
     - GENERATOR: Visual Studio 14 2015
+      VS_VERSION: 14.0
       CFG: Debug
     - GENERATOR: Visual Studio 12 2013
+      VS_VERSION: 12.0
       CFG: Debug
     - GENERATOR: Visual Studio 14 2015 Win64
       CFG: Debug
+      VS_VERSION: 14.0
     - GENERATOR: Visual Studio 12 2013 Win64
       CFG: Debug
+      VS_VERSION: 12.0
 install:
   - gem install asciidoctor
 build:
   parallel: true
 build_script:
+  - cmd: IF NOT %VS_VERSION% == NONE call "C:/Program Files (x86)/Microsoft Visual Studio %VS_VERSION%/Common7/Tools/vsvars32.bat"
   - cmd: cmake --version
   - cmd: md build
   - cmd: cd build
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -10,6 +10,8 @@ environment:
       CFG: Debug
     - GENERATOR: Visual Studio 12 2013 Win64
       CFG: Debug
+install:
+  - gem install asciidoctor
 build:
   parallel: true
 build_script:
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -84,11 +84,11 @@ endif()
 # User-defined options.
 
 option (NN_STATIC_LIB "Build static library instead of shared library." OFF)
+option (NN_ENABLE_DOC "Enable building documentation." ON)
 option (NN_ENABLE_GETADDRINFO_A "Enable/disable use of getaddrinfo_a in place of getaddrinfo." ON)
 option (NN_TESTS "Build and run nanomsg tests" ON)
 option (NN_TOOLS "Build nanomsg tools" ON)
 option (NN_ENABLE_NANOCAT "Enable building nanocat utility." ${NN_TOOLS})
-option (NN_ENABLE_DOC "Enable building documentation." ON)
 
 #  Platform checks.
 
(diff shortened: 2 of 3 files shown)


CMakeLists.txt
```diff
@@ -205,6 +207,92 @@ if (NN_ENABLE_NANOCAT)
     target_link_libraries (nanocat ${PROJECT_NAME})
 endif ()
 
+# Build the documenation
+if (NN_ENABLE_DOC)
+    find_program (ASCIIDOCTOR_EXE asciidoctor)
+    if (NOT ASCIIDOCTOR_EXE)
+        message (SEND_ERROR "asciidoctor not found, cannot create docs")
+    endif ()
```
Snaipe: `SEND_ERROR` continues execution, so the whole block below should be skipped if asciidoctor isn't found.

gdamore: According to CMake docs,

CMake Error, continue processing,
                              but skip generation

So I think this means that further operation inside cmake will continue,
but it won’t actually emit a Makefile (or whatever).  This way we can check
for other errors too.  Am I mistaken?

CMakeLists.txt
```diff
@@ -205,6 +207,98 @@ if (NN_ENABLE_NANOCAT)
...
+    add_libnanomsg_man (nn_pipeline 7)
+    add_libnanomsg_man (nn_bus 7)
+    add_libnanomsg_man (nn_inproc 7)
+    add_libnanomsg_man (nn_ipc 7)
+    add_libnanomsg_man (nn_tcp 7)
+    add_libnanomsg_man (nn_ws 7)
+    add_libnanomsg_man (nn_env 7)
+
+     add_custom_target (man ALL DEPENDS ${NN_MANS})
+     add_custom_target (html ALL DEPENDS ${NN_HTMLS})
```
JackDunaway: Indentation? Or is this a rendering issue?

CMakeLists.txt
```diff
@@ -205,6 +207,98 @@ if (NN_ENABLE_NANOCAT)
     target_link_libraries (nanocat ${PROJECT_NAME})
 endif ()
 
+find_program (ASCIIDOCTOR_EXE asciidoctor)
```
JackDunaway: Should this be inside the conditional block starting at 221? In other words, don't even check if it's installed unless doc build is enabled. Then, if this is moved inside that block, a fatal error could be issued, rather than just a `WARNING`

CMakeLists.txt
```diff
@@ -205,6 +207,98 @@ if (NN_ENABLE_NANOCAT)
...
+if (NOT ASCIIDOCTOR_EXE)
+    message (WARNING "asciidoctor not found, skipping docs")
+    set (NN_ENABLE_DOC OFF)
+else ()
+    if (NN_ENABLE_DOC)
+        message (STATUS "Using asciidoctor at ${ASCIIDOCTOR_EXE}")
```
JackDunaway: With regards to above comment, this line could go away, as being the implicit "success" path.

CMakeLists.txt
gdamore: No its real.  I fixed it.  I *hate hate hate* space based indentation.  8-cell tabs (with real tab characters) are *soo* much easier to see and use.  (This is a topic we may explore later.)

.appveyor.yml
```diff
@@ -2,17 +2,27 @@ version: 0.8.{build}
...
+      VS_VERSION: 12.0
       CFG: Debug
     - GENERATOR: Visual Studio 14 2015 Win64
       CFG: Debug
+      VS_VERSION: 14.0
     - GENERATOR: Visual Studio 12 2013 Win64
       CFG: Debug
+      VS_VERSION: 12.0
+install:
+  - gem install asciidoctor
```
JackDunaway: How much time does this add to the build? Probably not much, but it's worth monitoring.

CMakeLists.txt
gdamore: It was very helpful .. I can remove it now if you like.  (Problem was that I couldn't tell whether this was working or not on the VS builds because VS output sucks.)

CMakeLists.txt
gdamore: originally I had a plan to make the default value of using asciidoctor depend on this.  I couldn't quite achieve this with a fatal error, so by checking it here, I can disable it.  arguably I could enclose this in an if branch, but apart from saving the time to check (few dozen microsends of the build time I think), it doesn't buy me anything else.  I prefer to keep it simple as it is for now.

gdamore: Actually, maybe I can improve slightly, stay tuned.

JackDunaway: Psuedocode, starting at line 221 below:

IF NN_ENABLE_DOC
check for asciidoctor
if asciidoctor doesn't exist, fatal error
ELSEIF (not NN_ENABLE_DOC)
# do nothing; no check for asciidoctor needed, since docs not requested

.appveyor.yml
gdamore: Its tiny so far as I’ve seen.  There are ways to cache the results, but
that adds a lot of complexity so I’m going to avoid that until we see this
is a problem.

CMakeLists.txt
gdamore: The above pseudo code is wrong, since we want docs enabled by default, but don't want default options to fatal error.  So I've found and delivered a different solution to this already. :-)

JackDunaway:
Closing this PR; work has been mainlined as separate, squashed commits https://github.com/nanomsg/nanomsg/commit/6e14fe9af0622a88312eeb2a40d578e19f73cf6c and https://github.com/nanomsg/nanomsg/commit/b5224d57d46414d789045a6c0e5dc2bb7da11351

gdamore:
Oh yeah, thanks for that.

PR #671, opened 2016-04-25
Fixing issues building on Win x64

JackDunaway:


tests/win_sec_attr.c
```diff
@@ -50,27 +50,27 @@ int main ()
     PACL pACL;
     DWORD ret2;
     int ret3;
-    void * void_ret_value = NULL;
-    size_t void_ret_size = sizeof(void_ret_value);
+    void *void_ret_value = NULL;
+    size_t void_ret_size = sizeof (void_ret_value);
```
gdamore: do we need a scratch variable here?  would sizeof (void *) work equally well without the void_ret_value declaration?

tests/win_sec_attr.c
```diff
@@ -50,27 +50,27 @@ int main ()
...
 
     sc = test_socket (AF_SP, NN_PAIR);
     test_connect (sc, SOCKET_ADDRESS);
 
     sb = test_socket (AF_SP, NN_PAIR);
 
-    memset (&sec, 0, sizeof(sec));
-    sec.lpSecurityDescriptor = (PSECURITY_DESCRIPTOR)malloc (SECURITY_DESCRIPTOR_MIN_LENGTH);
+    memset (&sec, 0, sizeof (sec));
+    sec.lpSecurityDescriptor = (PSECURITY_DESCRIPTOR) malloc (SECURITY_DESCRIPTOR_MIN_LENGTH);
```
gdamore: This cast *should* be unnecessary.

gdamore: Because malloc should return void *... which should auto cast to anything.

tests/win_sec_attr.c
```diff
@@ -102,44 +102,36 @@ int main ()
...
-    the_sid = (PSID)&(allowed_ace->SidStart);
+    allowed_ace = (PACCESS_ALLOWED_ACE) ace;
+    the_sid = (PSID) &(allowed_ace->SidStart);
 
-    nn_assert(IsValidSid(the_sid));
+    nn_assert (IsValidSid (the_sid));
 
-    equal = EqualSid((PSID)&(allowed_ace->SidStart), &SIDAuthUsers);
-    nn_assert(equal);
-    LocalFree(dacl);
```
gdamore: Is this the source of the crash?

Snaipe: I think the original author misinterpreted the [doc for GetSecurityInfo](https://msdn.microsoft.com/en-us/library/windows/desktop/aa446654(v=vs.85).aspx) and meant `LocalFree(sd)`

JackDunaway: Yes; and @Snaipe, good call; I'm going to audit and see how many `LocalFree` need to be called ...

tests/win_sec_attr.c
JackDunaway: Good catch; will remove this too.

tests/win_sec_attr.c
JackDunaway: We need both variables for `nn_getsockopt`. Also, `sizeof (void_ret_value)` carries more semantic meaning that `sizeof (void *)` here.

tests/win_sec_attr.c
JackDunaway: > think the original author misinterpreted the doc for GetSecurityInfo and meant LocalFree(sd)

Glad you caught this. I had drawn the same conclusion about intent, but provided the wrong fix; new commit coming soon ...

PR #677, opened 2016-04-26
fixes #554 weird hang/crash & race against nn_term

### LINKED ISSUE #554: weird hang/crash & race against nn_term

gdamore:
I *hope* this will solve the hangs and crashes against nn_term.  I've done a few things.

nn_recvmsg and nn_sendmsg check explicitly for the nn_term case on error paths (fixes up error returns from some EBADFD cases.)

nn_term() actually *closes* the sockets on exit -- this arranges for all their resources to reaped.  Before we leaked like a sieve.  And, worse, we didn't always shut things down, and we didn't wait at all for threads to shut down.  As a consequence of this, nn_term() may take longer than it did before.  This shouldn't matter because nn_term() ought *never* be on your hot code path, or you're doing something badly wrong.

To make this work, I had to create a custom version of nn_close() that proceeds even in the face of the zombification.  This is used internally by nn_term.  Also the nn_global_hold_socket_locked() got a new argument.

nn_device had some bugs here, and the use of of the poll fds to loop was pretty bad, and the code was non-portable.  Its been replaced with a simpler design using two threads, one in each direction.  In addition to much easier to read and validate code, there should be scalability improvements, and the code should actually run quite a lot faster as we have half as many system calls to execute for each message.

This is totally untested at this point -- I'm hoping to get some CI test coverage with this PR.

### DIFF
--- a/src/core/global.c
+++ b/src/core/global.c
@@ -154,9 +154,13 @@ static int nn_global_create_ep (struct nn_sock *, const char *addr, int bind);
 static int nn_global_create_socket (int domain, int protocol);
 
 /*  Socket holds. */
-static int nn_global_hold_socket(struct nn_sock **sockp, int s);
-static int nn_global_hold_socket_locked(struct nn_sock **sockp, int s);
+static int nn_global_hold_socket (struct nn_sock **sockp, int s);
+static int nn_global_hold_socket_locked (struct nn_sock **sockp, int s,
+    int is_term);
 static void nn_global_rele_socket(struct nn_sock *);
+
+/*  Close helper. */
+static int nn_close_impl (int s, int is_term);
 
 int nn_errno (void)
 {
@@ -303,6 +307,7 @@ static void nn_global_term (void)
 void nn_term (void)
 {
     int i;
+    struct nn_sock *sock;
 
     nn_glock_lock ();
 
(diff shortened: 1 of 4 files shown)


src/core/global.c
```diff
@@ -303,20 +307,28 @@ static void nn_global_term (void)
...
+            }
     }
 
     nn_glock_unlock ();
+
+    /* Make sure we really close resources, this will cause global
+       resources to be freed too when the last socket is closed. */
+    for (i = 0; i < NN_MAX_SOCKETS; i++) {
+        (void) nn_close_impl (i, 1);
+    }
```
JackDunaway: In this loop, each iteration obtains/releases the glock. Is this semantic correct, compared to holding the glock for the entirety of looping through all sockets?

JackDunaway: That is, compared to the strategy just above this line, where the glock is held around the loop that zombifies all sockets.

JackDunaway: PS -- I understand that if the glock needs to be held around the loop, then `nn_close_impl` would need to be refactored (to not glock), which would cascade these same considerations anywhere `nn_close_impl` is called.

tests/device.c
```diff
@@ -51,7 +51,7 @@ void device1 (NN_UNUSED void *arg)
 
     /*  Run the device. */
     rc = nn_device (deva, devb);
-    nn_assert (rc < 0 && (nn_errno () == ETERM || nn_errno () == EBADF));
+    nn_assert (rc < 0 && (nn_errno () == ETERM));
```
JackDunaway: Good to see the test tighten up here! If the previous pass criteria would switch between the two error codes, yet now it's just one, that seems like a step in a good direction!

src/core/global.c
gdamore: Yes this is the right way. We're shutting down so the grab and release is fine here.

PR #678, opened 2016-04-26
fixes #666 Unix domain socket files not unlinked when closed

### LINKED ISSUE #666: Unix domain socket files not unlinked when closed
IPC sockets on linux doesn't seem to unlink the socket file when the nanomsg socket is closed/terminated.

This is consistently happenning for all IPC tests.

Tested on Linux 4.4.5-1 (ArchLinux) and Linux 3.13.0-85 (Ubuntu 14.04).

Snaipe:


src/transports/ipc/bipc.c
```diff
@@ -168,6 +168,11 @@ static void nn_bipc_destroy (struct nn_epbase *self)
 static void nn_bipc_shutdown (struct nn_fsm *self, int src, int type,
     void *srcptr)
 {
+#if !defined NN_HAVE_WINDOWS
```
JackDunaway: It would be better to see not not-logic here -- and also, ideally a symbol that represents a feature, not a platform. Based on our current symbol definitions from CMake, would this be reasonable? Do we need more symbols? Or even another CMake option such as `NN_ENABLE_RELEASE_PIPE`  (that defaults to `ON`) ?

Snaipe: I used the same test as the matching [unlinking code below](https://github.com/Snaipe/nanomsg/blob/ipc-unlink/src/transports/ipc/bipc.c#L415-L429) to be consistent, but you're right: it should definitely be inside a feature test rather than a platform test.

There is currently no way to check if unix domain sockets are supported, other than supposing that all platforms other than windows support them. Maybe a separate commit should address this specific issue, but checking for availability of domain socket is not as trivial as checking for `AF_UNIX`, because windows actually provide a definition of it without using it in practice.

Snaipe: Actually, nevermind: `AF_UNIX` on windows is defined in `winsock2.h`, so we can feature test with `sys/socket.h`.

gdamore: Its stil a reasonable question — the presence of the #define doesn’t mean
that IPC is actually implemented.  Indeed, I’ve some evidence that IPC
doesn’t work quite the same everywhere — it can, for example, depend on the
filesystem code supporting it.  For example, I believe it doesn’t work in
all directories on QNX (or so I’ve heard).  As another example, it doesn’t
work on FAT filesystems where you need to have support for pipe special
files.

That said, I think we can leave this question up to the platform code — if
the code *compiles* then that’s all we can do.  If it fails at runtime,
then the user is going to have to work around it.

For the case of Windows, we don’t use AF_UNIX at all, and I’m happy to have
a per-platform feature test for it — Windows is such a special snowflake
here (with its named pipes) that it deserves special handling.

 - Garrett

gdamore:
Looks good.  You want to rebase this or should I?  Also you want to update copyright?

Issue #561, opened 2015-11-28
includedir goes too deep


donpdonp:
the includedir pkgconfig has /nanomsg in the path so #include <nanomsg/bus.h> is not found in include/nanomsg/nanomsg/bus.h

```
pkg-config --cflags libnanomsg
-I/opt/nanomsg/0.8-beta/include/nanomsg 
```

```
 go get github.com/op/go-nanomsg
# github.com/op/go-nanomsg
src/github.com/op/go-nanomsg/bus.go:5:26: fatal error: nanomsg/bus.h: No such file or directory
 // #include <nanomsg/bus.h>
                          ^
```

editing libnanomsg.pc.in to be
```
includedir=@includedir@
```
fixes the problem.

gdamore:
This looks like the change issued with #487   Sadly I'm not sure what is correct right now... it seems like we have some disagreements about pkgconfig.  I don't actually use that tool.

gdamore:
So it looks like your pkg-config flags for -I are wrong.  Don't say "include/nanomsg", but just "include".  Did you supply the -I option to pkg-config by hand?

gdamore:
I'm closing this for now.  Feel free to reopen when you're prepared to provide more info.

gdamore:
Huh.  In retrospect I think we got this wrong.  Applications should indeed include <nanomsg/nn.h> or whatever.

PR #686, opened 2016-04-30
fixes #685 cmake needs to deliver libnanomsg.pc (pkg-config data file)

### LINKED ISSUE #685: cmake needs to deliver libnanomsg.pc (pkg-config data file)
We prepare this file with autoconf, but we need to change the cmake code to prepare it, and deliver it in the same way that autoconf does.

gdamore:
This creates a file, nanomsg.pc  -- (note the absence of the leading lib prefix) for nanomsg.)

This new file behaves as specified in #561 -- and furthermore, simplifies Windows library detection works -- checking using the same tests we use for other platforms.  (Hopefully it will work!)

As a  bonus, it corrects the Libs.private entry so that users wanting to link statically can get the right answers with pkg-config --static --libs nanomsg

The version emitted is the library version, not the ABI version.

This pkg-config file departs rather significantly from the version used in autotools; folks relying on it will have to change things.  But this is a big improvement IMO.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -356,21 +356,23 @@ set_target_properties (${PROJECT_NAME} PROPERTIES LIBRARY_OUTPUT_DIRECTORY ${PRO
 set_target_properties (${PROJECT_NAME} PROPERTIES ARCHIVE_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR})
 
 target_link_libraries (${PROJECT_NAME} ${NN_REQUIRED_LIBRARIES})
-if (WIN32)
-    target_link_libraries (${PROJECT_NAME} ws2_32)
-    target_link_libraries (${PROJECT_NAME} mswsock)
-    target_link_libraries (${PROJECT_NAME} advapi32)
-elseif (UNIX)
-    if(THREADS_HAVE_PTHREAD_ARG)
-        add_definitions (-pthread)
-    endif()
-    if(CMAKE_THREAD_LIBS_INIT)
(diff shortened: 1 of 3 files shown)


gdamore:
Well, my attempt to simplify windows checks for libraries failed miserably.  I'll revert that portion.

src/CMakeLists.txt
```diff
@@ -356,21 +356,23 @@ set_target_properties (${PROJECT_NAME} PROPERTIES LIBRARY_OUTPUT_DIRECTORY ${PRO
...
+    add_definitions (-pthread)
+endif()
+if(CMAKE_THREAD_LIBS_INIT)
+    target_link_libraries (${PROJECT_NAME} "${CMAKE_THREAD_LIBS_INIT}")
+endif()
 
+# pkg-config file
+if(NN_REQUIRED_LIBRARIES)
+    foreach (lib ${NN_REQUIRED_LIBRARIES})
+        set (NN_REQUIRED_LFLAGS "${NN_REQUIRED_LFLAGS} -l${lib}")
```
Snaipe: Following the current CMakeLists.txt, this is going to have `-lpthread` rather than `-pthread`. I don't think it matters in practice though (I've never seen any pthread linker flag other than `-lpthread`), but it might be desirable for compliance/portability of user programs.

gdamore: There are compilers that do -pthread -- this flag usually sets (on Solaris) some other flags -- generally -D_REENTRANT and possibly -D_POSIX_PTHREAD_SEMANTICS, and adding -lpthread.  SunPro (legacy) had -mt for a similar purpose -- adding -lthread and adding the -D_REENTRANT.

The link-time stuff is just -lpthread, which ought to be sufficient.

Issue #692, opened 2016-05-09
nn_term semantics are busted


gdamore:
As documented, nn_term() is unusable.  Its also the source of a lot of problems. 

Specifically the API calls for nn_term() to be a non-blocking function, that wakes all other workers, which are supposed to clean up, and all outstanding blocking calls return -ETERM.  The problem is that without blocking, it is impossible to know when it is safe to clean up, or even reinitialize the library -- worker threads can still be running, etc.   Further, the specific requirements that functions in flight return -ETERM immediately upon completion creates a significant headache of race conditions to clean up, and is the source of quite a few bugs.  Having a secondary way to for sockets to terminate beyond just normal close is unwarranted complexity.  Its also very unlike POSIX.  Finally, once performed, there is no way for a caller to reinitialize the library for subsequent use.

As specified, nn_term() is mostly useless.

We therefore propose the following changes:

* nn_term() shall become a blocking call.
* it shall operate by setting a global flag (as it does today), and by closing all open sockets.
* nn_socket() shall continue to check the flag and return -ETERM if it is set
* nn_init() shall be a new function which clears the global flag
* all other functions shall operate as normal.  This means that after nn_term() returns (and before
 nn_init() is called) they shall return -EBADF.

gdamore:
This should address the problems noted in #562 as well.  nn_term will remain useful for some consumers, but in general its behavior will be far more predictable.

Note that nn_term should also be considered optional -- that is it shouldn't release any resources that would not also be released by a call to nn_close() of the last socket.

PR #694, opened 2016-05-10
Newterm

gdamore:
This should address both #692 and the various shutdown races.  The latter is only fixed on POSIX -- I'm not sure yet whether the same fix is required for Windows or not.

### DIFF
--- a/src/core/global.c
+++ b/src/core/global.c
@@ -308,10 +308,7 @@ void nn_term (void)
     int i;
 
     nn_glock_lock ();
-
-    /*  Switch the global state into the zombie state. */
     self.flags |= NN_CTX_FLAG_TERMING;
-
     nn_glock_unlock ();
 
     /* Make sure we really close resources, this will cause global
--- a/src/core/global.c
+++ b/src/core/global.c
@@ -43,6 +43,7 @@
 #include "../utils/chunk.h"
 #include "../utils/msg.h"
 #include "../utils/attr.h"
+#include "../utils/sleep.h"
 
 #include "../transports/inproc/inproc.h"
 #include "../transports/ipc/ipc.h"
@@ -92,7 +93,9 @@
     the type should be changed to uint32_t or int. */
 CT_ASSERT (NN_MAX_SOCKETS <= 0x10000);
 
-#define NN_CTX_FLAG_ZOMBIE 1
+#define NN_CTX_FLAG_TERMED 1
+#define NN_CTX_FLAG_TERMING 2
+#define NN_CTX_FLAG_TERM (NN_CTX_FLAG_TERMED | NN_CTX_FLAG_TERMING)
 
 #define NN_GLOBAL_SRC_STAT_TIMER 1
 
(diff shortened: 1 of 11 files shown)


src/core/global.c
```diff
@@ -309,25 +308,33 @@ void nn_term (void)
...
+    nn_glock_lock ();
+    /*  Make sure any nn_term in progress is done. */
+    while (self.flags & NN_CTX_FLAG_TERMING) {
+        nn_glock_unlock ();
+        nn_sleep (100);
+        nn_glock_lock ();
     }
+    self.flags &= ~NN_CTX_FLAG_TERMED;
+    nn_glock_unlock ();
 }
```
JackDunaway: Is `nn_init` actually called anywhere right now? Or is this the TODO from line 323?

src/utils/efd.c
```diff
@@ -44,18 +45,60 @@ int nn_efd_wait (struct nn_efd *self, int timeout)
...
+        if (nn_slow (pfd.fd < 0))
+            return -EBADF;
+
+        switch (expire) {
+        case 0:
+            /* poll once */
+            timeout = 0;
+            break;
+
+        case (uint64_t)-1:
```
JackDunaway: Is this valid? `timeout` is of type `int`.

src/utils/efd.c
```diff
@@ -44,18 +45,60 @@ int nn_efd_wait (struct nn_efd *self, int timeout)
...
-        return -EINTR;
-    errno_assert (rc >= 0);
-    if (nn_slow (rc == 0))
-        return -ETIMEDOUT;
-    return 0;
+    if (timeout > 0) {
+        expire = nn_clock_ms() + timeout;
+    } else {
+        expire = timeout;
+    }
```
JackDunaway: It's not immediately clear what values <-1 would do. I believe these are prevented upstream, and if that's the case, it would make sense to insert an `nn_assert (timeout == -1);` here, perhaps with a comment why.

src/utils/efd.c
gdamore: Yes.  The value here is "expire", not timeout.

src/utils/efd.c
gdamore: I would be happy to make this more explicit -- a timeout of -1 is valid, and results in expire = (uint64_t)-1.  That means an infinite timeout.  This is not a place for an nn_assert().

The unfortunate thing (and it does work this way) is that its not obvious that assigning (int)-1 to a uint64_t results in a value of 0xffffffffffffffffULL (aka (uint64_t)-1).  A more clear way to write this, although slightly less efficient, would be to explicitly check for timeout < 0 and setting expire = (uint64_t)-1.

If you want me to make that change for clarity (at a tiny expense in terms of performance) I'm happy to do so.

gdamore:
Fixed and integrated:

commit 994a99f7180017218fe3866fffd70a9990dfb7f6
Date:   Tue May 10 13:07:16 2016 -0700

    fixes #692 nn_term semantics are busted

Issue #583, opened 2015-12-17
Occasional deadlock during nn_shutdown in websocket


gdamore:

JackDunaway:
Fixed with patches introduced by #695

Issue #581, opened 2015-12-15
WebSocket transport lagging features of TCP Transport


JackDunaway:
The WebSocket transport is currently lagging features of the TCP transport, including reconnect backoff timers for both bind and connect, and also a socket receive maximum bytes option.

reqshark:
on the browser-side of the connection, is there a way to configure nanomsg features?

wirebirdlabs:
Would you explain?

reqshark:
sure, let's say i set up an NN_PAIR or some other socket type. 

On the *browser-side*, setting no options, I'm actually able to get msgs into a web browser's JavaScript like this:

```js
      var ws = new WebSocket('ws://127.0.0.1:7789', [
      // see the rfc on sp websocket mapping:
      // raw.githubusercontent.com/nanomsg/nanomsg/master/rfc/sp-websocket-mapping-01.txt
        'pair.sp.nanomsg.org'
      ]);
      ws.onopen = function open() { ws.send('Ping'); }
      ws.onerror = function error(err) { console.log(err); }
      ws.onmessage = function message(e) {
        var reader = new FileReader();
        reader.addEventListener('loadend', function() {
          document.querySelector('span').textContent = reader.result;
        });
        reader.readAsText(e.data);
      }

```

Would any socket options I set on the other end *pass-through* or is there any plan to be able to set options from the browser-side? Does that make sense?

reqshark:
btw, when i say "get msgs into a web browser's JavaScript " I also mean send msgs back! :+1:

wirebirdlabs:
What kinds of options are you wanting to set?

reqshark:
Really just `NN_RCVMAXSIZE`, for larger data sets or images bigger than 1024kB..

But at some point if `NN_SNDBUF` or `NN_RCVBUF` could improve perf I would want to set those

wirebirdlabs:
That'd be a security hole for any transport type (not just WebSocket) -- enabling a socket peer to change behavior like this.

If you want to be able to tune those parameters remotely, you can introduce your own application-level protocol (but it doesn't seem like a good idea)

reqshark:
Ok I could see that, thanks!

PR #695, opened 2016-05-11
fixes #584, #583, #581

### LINKED ISSUE #584: Want async shutdown test for websocket
As this has been a challenge for us in the past, @wirebirdlabs has written a test case for async shutdown to check for races against issues elsewhere in the library.

JackDunaway:


src/transports/ws/sws.c
```diff
@@ -678,7 +686,76 @@ static void nn_sws_validate_utf8_chunk (struct nn_sws *self)
...
+    pos = self->inmsg_current_chunk_buf;
+
+    /*  Peer did not provide a Close Code, so choose our own here. */
+    if (len == 0) {
+        nn_sws_fail_conn (self, NN_SWS_CLOSE_NORMAL, "");
+        return;
+    }
+
+    /*  If the payload is not even long enough for the required 2-octet
+        Close Code, the connection should have been failed upstream. */
```
gdamore: What does "upstream" mean in this context?

src/transports/ws/sws.c
```diff
@@ -678,7 +686,76 @@ static void nn_sws_validate_utf8_chunk (struct nn_sws *self)
...
+    nn_assert (len >= NN_SWS_CLOSE_CODE_LEN);
+    len -= NN_SWS_CLOSE_CODE_LEN;
+    pos += NN_SWS_CLOSE_CODE_LEN;
+
+    /*  As per RFC 6455 7.1.6, the Close Reason following the Close Code
```
gdamore: While this is required for generation, do we care about actually checking this?  What happens if we just ignore the close code whether well formed or not?  (Be generous in what you accept.... )

src/transports/ws/sws.c
```diff
@@ -678,7 +686,76 @@ static void nn_sws_validate_utf8_chunk (struct nn_sws *self)
...
+        close_code == NN_SWS_CLOSE_ERR_EXTENSION ||
+        close_code == NN_SWS_CLOSE_ERR_SERVER ||
+        (close_code >= 3000 && close_code <= 3999) ||
+        (close_code >= 4000 && close_code <= 4999)) {
+        /*  Repeat close code, per RFC 6455 7.4.1 and 7.4.2 */
+        nn_sws_fail_conn (self, (int) close_code, "");
+    }
+    else {
+        nn_sws_fail_conn (self, NN_SWS_CLOSE_ERR_PROTO,
+            "Unrecognized close code.");
```
gdamore: TBH, I'm not sure what value passing the close code here gives.  How is this useful to the application?

src/transports/ws/sws.c
```diff
@@ -759,7 +835,7 @@ static int nn_sws_fail_conn (struct nn_sws *self, int code, char *reason)
         nn_fsm_raise (&self->fsm, &self->done, NN_SWS_RETURN_CLOSE_HANDSHAKE);
     }
 
-    return 0;
+    return;
```
gdamore: Return statement here not needed.

src/transports/ws/sws.c
```diff
@@ -1200,13 +1272,25 @@ static void nn_sws_handler (struct nn_fsm *self, int src, int type,
                             sws->inmsg_current_chunk_buf = sws->inmsg_control;
                         }
                         else {
-                            sws->inmsg_chunks++;
                             sws->inmsg_total_size += sws->inmsg_current_chunk_len;
+                            /*  Protect non-control messages against the
```
gdamore: Yay.  This is a critical stopper bug you are fixing. :-)

src/transports/ws/sws.c
```diff
@@ -1298,8 +1379,19 @@ static void nn_sws_handler (struct nn_fsm *self, int src, int type,
...
                         sws->inmsg_total_size += sws->inmsg_current_chunk_len;
+                        /*  Protect non-control messages against the
+                            NN_RCVMAXSIZE threshold; control messages already
+                            have a small pre-allocated buffer, and therefore
+                            are not subject to this limit. */
+                        nn_pipebase_getopt (&sws->pipebase, NN_SOL_SOCKET,
+                            NN_RCVMAXSIZE, &opt, &opt_sz);
+                        if (opt >= 0 && sws->inmsg_total_size > (size_t) opt) {
+                            nn_sws_fail_conn (sws, NN_SWS_CLOSE_ERR_TOOBIG,
+                                "Message larger than application allows.");
```
gdamore: This message, if exposed to users, could be confusing.  "Message larger than NN_RCVMAXSIZE" might be clearer.  Or just "message size exceeds limit"?

tests/ws_async_shutdown.c
```diff
@@ -0,0 +1,106 @@
...
+        if (nn_errno () == ETIMEDOUT) {
+            continue;
+        }
+        break;
+    }
+    /*  Socket is expected to be closed by caller.  */
+    errno_assert (nn_errno () == EBADF);
+}
+
+int main ()
```
gdamore: I'd prefer to see argc, argv arguments (even if unused) to main().

src/transports/ws/bws.c
```diff
@@ -314,6 +344,71 @@ static void nn_bws_handler (struct nn_fsm *self, int src, int type,
         }
 
 /******************************************************************************/
+/*  CLOSING_USOCK state.                                                      */
```
gdamore: Just a little note.  I hate the use of enclosing asterisks in block comments.  It becomes a pain to maintain them, and it looks horrible if ever used with a non-monospaced font (I know don't do that!) I used to use the same comment style when I was young, but in my years since I've learned that they add little value -- the ones on the right side are especially annoying to have to maintain.  Usually just a single multiline comment with left aligned asterisks is sufficient to catch attention -- especially if surrounded by whitespace.

gdamore: That said, I think Martin used these block comments extensively elsewhere in the code.  They still annoy me, and I celebrate a little when I can remove them. :-)

JackDunaway: Agree here. I'll leave the format for now for conformance, then we can come back later and change them all in bulk.

src/transports/ws/sws.c
JackDunaway: Upstream logic in the currently-processing message. Not upstream in the sense of "further up the chain" on a block diagram of the system.

src/transports/ws/sws.c
JackDunaway: The Close Reason itself it optional, but if it exists, it **MUST BE** valid UTF-8. If it is not, RFC 6455 states that we (local socket) can immediately fail the connection without best-effort response. If it's well-formed, RFC 6455 states we must give a best-effort response. I see where you're coming from, but relaxing the spec here in this manner would be incorrect.

gdamore: So it would be wrong to give a best effort response to this, even if the reason was not valid UTF-8?

src/transports/ws/sws.c
JackDunaway: It's just RFC 6455 conformance. A well-behaved server (like this logic) should provide best-effort hints as to why it's unexpectedly closing a connection when it does, based on the malconformance mode of the client.

src/transports/ws/sws.c
gdamore: So lets change the wording here, because its unclear.  Perhaps say "should already have been failed."

src/transports/ws/sws.c
JackDunaway: IIRC, it's something that pedantic compilers prefer; either way, it doesn't hurt to be explicit. Do you have strong opionions otherwise? Else, we'll keep it for now.

src/transports/ws/sws.c
JackDunaway: Good suggestion; will change.

tests/ws_async_shutdown.c
JackDunaway: Even further, this test needs the full-blown upgrade made by @Snaipe to dynamically retrieve its port. (This test predates that upgrade.) Stay tuned ...

src/transports/ws/sws.c
JackDunaway: Will change.

src/transports/ws/sws.c
JackDunaway: No, since a best-effort response is still given for all other failure modes. It's a paltry effort, but it tries at least once prior to closing the connection.

JackDunaway: Also -- it's very helpful to see how all this nanomsg websocket code interacts with the Autobahn Test Suite. It's simple to run A/B tests on question about behavior like this one, since it does an excellent job of probing all corner cases.

gdamore: Ok, lets check this at some point -- I'd rather not have the burden of validating UTF-8-ishness if we can skip it.  Seems to be of very little merit here, IMO.

src/transports/ws/sws.c
gdamore: I see, this response is sent back to the far end, so its for peer consumption rather than app consumption.  Right?

src/transports/ws/sws.c
gdamore: No "pedantic" compiler that I'm aware of "prefers" an explicit return at the end of a void function.  If there is a returned value, then that's different.   (Implicit return values of zero are what compilers bitch about.)  I'd prefer to remove it, actually.

PR #696, opened 2016-05-11
Compiler nits

JackDunaway:


src/aio/poller_kqueue.inc
```diff
@@ -107,6 +105,7 @@ void nn_poller_reset_in (struct nn_poller *self, struct nn_poller_hndl *hndl)
     if (hndl->events & NN_POLLER_EVENT_IN) {
         EV_SET (&ev, hndl->fd, EVFILT_READ, EV_DELETE, 0, 0, 0);
         rc = kevent (self->kq, &ev, 1, NULL, 0, NULL);
+        nn_assert (rc == 0);
```
gdamore: This feels wrong.  What is this fixing?

tests/testutil.h
```diff
@@ -201,12 +201,12 @@ static void NN_UNUSED test_drop_impl (char *file, int line, int sock, int err)
     }
 }
 
-static int get_test_port(int argc, const char *argv[])
+static int NN_UNUSED get_test_port (int argc, const char *argv[])
```
gdamore: What is unused here?  The return value?  The arguments?  Which ones?  I think you need to apply NN_UNUSED in front of specific arguments if that's what they are for.

tests/testutil.h
```diff
@@ -201,12 +201,12 @@ static void NN_UNUSED test_drop_impl (char *file, int line, int sock, int err)
...
 {
     return atoi(argc < 2 ? "5555" : argv[1]);
 }
 
-static void test_addr_from(char *out, const char *proto,
+static void NN_UNUSED test_addr_from (char *out, const char *proto,
```
gdamore: Same as above.  I think NN_USED is in the wrong place.

src/aio/poller_kqueue.inc
JackDunaway: It's definitely wrong; it's not a good fix, but a placeholder for something that needs to be further investigated. The compiler warned of an unused assignment to `rc` here -- and for `kevent`, we don't even expect `rc == 0`

tools/options.c
```diff
@@ -333,8 +333,8 @@ static void nn_append_string (struct nn_parse_context *ctx,
 {
     struct nn_string_list *lst;
 
-    lst = (struct nn_string_list *)(
-        ((char *)ctx->target) + opt->offset);
+    lst = (struct nn_string_list *)(((char *)ctx->target) + opt->offset);
+    nn_assert (lst);
```
gdamore: maybe better to nn_assert (ctx != NULL) before the above assignment?

tests/testutil.h
JackDunaway: This follows the precedent set by all other functions in this file. I believe it's some artifact of how the macros are called, not these functions themselves.

tools/options.c
```diff
@@ -345,7 +345,9 @@ static void nn_append_string (struct nn_parse_context *ctx,
     if (!lst->items) {
         nn_memory_error (ctx);
     }
-    lst->items[lst->num-1] = str;
```
gdamore: This is a logic change.  It shouldn't matter because nn_memory_error (ctx) should probably abort the program.  Maybe that needs a NORETURN (or whatever) statement?  I actually don't like adding the extra indentation here, it makes the flow harder to grok.

tests/testutil.h
JackDunaway: Example of precedent: https://github.com/nanomsg/nanomsg/pull/696/files/cdcb73a7dbebfaf9bb98190baef4372df1720937#diff-650a958e586e7f65e22e0174b4bac596R186

tools/options.c
```diff
@@ -355,6 +357,7 @@ static void nn_append_string_to_free (struct nn_parse_context *ctx,
 
     lst = (struct nn_string_list *)(
         ((char *)ctx->target) + opt->offset);
+    nn_assert (lst);
```
gdamore: As above, probably better to assert the ctx rather than lst.  The compiler should be able to figure it out from there.

tools/options.c
```diff
@@ -366,7 +369,8 @@ static void nn_append_string_to_free (struct nn_parse_context *ctx,
     if (!lst->items) {
         nn_memory_error (ctx);
     }
-    lst->to_free[lst->to_free_num-1] = str;
+    nn_assert (lst->to_free);
```
gdamore: This might be an assert against the str at the start of the function?

tools/options.c
```diff
@@ -789,6 +793,7 @@ void nn_free_options (struct nn_commandline *cline, void *target) {
         case NN_OPT_LIST_APPEND:
         case NN_OPT_LIST_APPEND_FMT:
             lst = (struct nn_string_list *)(((char *)target) + opt->offset);
+            nn_assert (lst);
```
gdamore: Similar comments as above.

tools/options.c
JackDunaway: Look again -- it's `lst->to_free` that needs the assert because it's indexed with `[lst->to_free_num-1]`

tests/testutil.h
gdamore: Ok, I see now.  I think this is goofy, and but its doing what was intended, because testutil.h is compiled inline and these are static functions.  So I'm fine with that for now.  We ought to fix this to be less "weird" in the future.  (A separate header and .c file would be more typical.)

tests/testutil.h
gdamore: As above, fine for now.

tools/options.c
JackDunaway: This doesn't quite work, because the thing we must guarantee is that `lst` is indeed non-NULL, yet it's calculated from two other function parameters.

tools/options.c
JackDunaway: Not quite enough to constrain -- see: https://github.com/nanomsg/nanomsg/pull/696/files#r62933992

tools/options.c
JackDunaway: How about just adding `nn_assert (lst && lst->items);`? I agree the logic change obfuscates.

src/protocols/pubsub/trie.c
```diff
@@ -299,7 +298,6 @@ int nn_trie_subscribe (struct nn_trie *self, const uint8_t *data, size_t size)
     memmove (ch->prefix, ch->prefix + pos + 1, ch->prefix_len);
     ch = nn_node_compact (ch);
     *nn_node_child (*node, 0) = ch;
-    pos = (*node)->prefix_len;
```
JackDunaway: @gdamore This looks like it may have been important at one time, but it's dead code now. I've tried to guess its intended meaning, but can't figure out anything it should obviously be. Shall we move forward with removing it?

Asked another way, is this orphaned code that should be removed, or an unfinished thought that should be finished?

tools/options.c
gdamore: Ok, so the more I think about this, the more vociferously I believe this is all wrong.  We're adding in a specific check for exactly one form of error (NULL pointers).  There are a bunch of others, stemming from e.g. use-after-free, uninitialized (and hence random) pointers on the stack, pointers to objects of the wrong data type, etc. etc.

The analyzer that is triggering this is crap, and I'm *not* interested in doing runtime parameter validation for all of these pointers.  There may be specific cases where such an assertion adds value to us (by documenting an assumption that we are making which isn't otherwise immediately obvious).

I'm not interested in doing the equivalent of adding assertions against NULL to function arguments everywhere.  You won't find libc does this (e.g. strcpy does not), and neither should we.

src/protocols/pubsub/trie.c
gdamore: I haven't actually tried to understand the code.  Its probably dead from an earlier implementation, but I had nothing to do with it.

tools/options.c
gdamore: I'm withdrawing all my exceptions to this usage in this file.  This is in options.c -- so not in the mainline library, and the assertions therefore do no harm to any end users.  The need for them underscores pretty strongly to me that the options design here is somewhat sketchy, but it works for now.

PR #700, opened 2016-05-11
fixes #692 nn_term semantics are busted

### LINKED ISSUE #692: nn_term semantics are busted

### LINKED ISSUE #699: tcp_shutdown hangs
We see many tcp_shutdown fixes in Windows on AppVeyor.  Further, I've demonstrated a 100% reliable hang in test code as part of the changes for #692 that shows a hang in nn_close() when other thread(s) are running with nn_recv() on the same socket.

The problem is that poll (and I hypothesize also select() on windows) does not wake up when one of the file descriptors is closed locally.

The band-aid workaround is to wake up at least once every 100 ms to check to make sure the socket is still open.

In the longer term, we'll replace the poll/select polling for notification with a simpler condition variable scheme that will be immune to these errors (and coincidentally also quite a bit more efficient / faster).

gdamore:
fixes #699 tcp_shutdown hangs

### DIFF
--- a/src/core/global.c
+++ b/src/core/global.c
@@ -43,6 +43,7 @@
 #include "../utils/chunk.h"
 #include "../utils/msg.h"
 #include "../utils/attr.h"
+#include "../utils/sleep.h"
 
 #include "../transports/inproc/inproc.h"
 #include "../transports/ipc/ipc.h"
@@ -92,7 +93,9 @@
     the type should be changed to uint32_t or int. */
 CT_ASSERT (NN_MAX_SOCKETS <= 0x10000);
 
-#define NN_CTX_FLAG_ZOMBIE 1
+#define NN_CTX_FLAG_TERMED 1
+#define NN_CTX_FLAG_TERMING 2
+#define NN_CTX_FLAG_TERM (NN_CTX_FLAG_TERMED | NN_CTX_FLAG_TERMING)
 
 #define NN_GLOBAL_SRC_STAT_TIMER 1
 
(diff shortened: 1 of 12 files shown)


gdamore:
Rebased and integrated:

commit 994a99f7180017218fe3866fffd70a9990dfb7f6
Date:   Tue May 10 13:07:16 2016 -0700

    fixes #692 nn_term semantics are busted

The fix for #699 and tcp starvation is left for later.

PR #703, opened 2016-05-11
another swag at fixing tcp_shutdown on windows.

gdamore:


src/utils/efd.c
```diff
@@ -119,6 +119,7 @@ int nn_efd_wait (struct nn_efd *self, int timeout)
     }
 
     for (;;) {
+        fd = self->r;
```
JackDunaway: Good catch; even if doesn't fix issue, this is a good change to keep.

gdamore: Agreed.  I still haven’t figured it out.  This is going to take time.

gdamore:
All fixed up now.  Closing.

PR #708, opened 2016-05-13
Add support for mutex debugging (mutex recursion, unowned lock release.)

gdamore:
This adds more extensive checks for mutex correctness, but only if
the environment variable NN_MUTEX_DEBUG is set.  This additional
level of debugging can have a small performance impact, but the
checks may help us detect and diagnose programming errors involving
mutexes sooner.

### DIFF
--- a/src/utils/mutex.c
+++ b/src/utils/mutex.c
@@ -23,25 +23,44 @@
 #include "mutex.h"
 #include "err.h"
 
+#include <stdlib.h>
+
 #ifdef NN_HAVE_WINDOWS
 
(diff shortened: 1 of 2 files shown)


gdamore:
If the initial build works, I'll add NN_MUTEX_DEBUG to the CI environment files.

JackDunaway:
I like this idea a lot; but as discussed on Gitter, we may be able to get away with doing it unconditionally rather than needing an env var, since it's so cheap: https://gitter.im/nanomsg/nanomsg?at=573605ce64dbdadc7debd519

PR #718, opened 2016-05-16
fixes #717 NN_STATIC_LIB not properly handled in MSVC

### LINKED ISSUE #717: NN_STATIC_LIB not properly handled in MSVC
Currently, `cmake -DNN_STATIC_LIB=ON ..` is not handled properly compiling with MSVC.

On build, many LNK2019 errors are generated by the tests/tools built as executable, since there is no symbol defined for these targets that governs `NN_EXPORT`. Based on the current behavior,`NN_EXPORT` then falls into the default case of `__declspec(dllimport)`.

The correct behavior when linking a static library from an application, when it is not desired to re-export those functions, is to ensure `NN_EXPORT` resolves to an empty string (i.e., no explicit specification for export visibility).

A PR is forthcoming to fix this issue.

JackDunaway:
Note that also the definition of `NN_EXPORT` within `nn.h` has been tightened from `defined(_WIN32)` to `defined(NN_HAVE_MSVC)` for explicitness and correctness.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -157,6 +157,14 @@ if (WIN32)
     set(NN_REQUIRED_LIBRARIES ${NN_REQUIRED_LIBRARIES} ws2_32)
     set(NN_REQUIRED_LIBRARIES ${NN_REQUIRED_LIBRARIES} mswsock)
     set(NN_REQUIRED_LIBRARIES ${NN_REQUIRED_LIBRARIES} advapi32)
+
+    # Ensure tests/tools built as EXE choose symbol visibility appropriate for
+    # linking the library, whether static or dynamic.
+    if (NN_STATIC_LIB)
+        add_definitions (-DNN_WIN32_LIB_LINK)
+    else ()
+        add_definitions (-DNN_WIN32_DLL_LINK)
+    endif ()
 else ()
     # Unconditionally declare the following feature test macros.  These are
     # needed for some platforms (glibc and SunOS/illumos) and should be harmless
--- a/src/nn.h
+++ b/src/nn.h
@@ -36,10 +36,14 @@ extern "C" {
 
 /*  Handle DSO symbol visibility. */
 #if !defined(NN_EXPORT)
-#    if defined(_WIN32) && !defined(NN_STATIC_LIB)
-#        if defined NN_SHARED_LIB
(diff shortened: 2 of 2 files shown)


CMakeLists.txt
```diff
@@ -123,6 +123,18 @@ else ()
     message (AUTHOR_WARNING "${ISSUE_REPORT_MSG}")
 endif ()
 
+# MSVC symbol visibility and linker behavior is sufficiently different from all
+# other platforms. Here, we ensure tests/tools built as EXE choose symbol
+# visibility appropriate for linking the library, whether static or dynamic.
+if (MSVC)
```
gdamore: What happens with MINGW or other compilers (Borland?) on Win32 platforms?  Put another way, is the issue here the *compiler* (MSVC) or the *platform* (Windows)?  I don't know the answer, but I hope you or someone else here might.

JackDunaway: I cannot say with certainty, but my understanding is that this is MSVC-specific, and therefore tightening works in our favor.

My understanding is that MinGW64 would export all symbols; I'm not sure about Borland.

In this regard, MinGW64 correctness would be improved by this move (I have not confirmed this)

The rationale is to tighten the definition as much as possible prior to 1.0, rather than being too gratuitous.

Also of note: the definition for `NN_EXPORT` has been incorrect for a long time for `NN_STATIC_LIB` on MSVC. It would have been broken altogether for users prior to https://github.com/nanomsg/nanomsg/commit/5b34eb6467975ff4c444e2f8e1267bf01aba4b36#diff-2051e1a7c073348e8ba0043ac9b5c77c, and that commit was not even exactly correct (even though, it "worked")

Snaipe: This would actually break mingw-w64 (& non-msvc in general) builds because NN_EXPORT isn't defined to `__declspec(dllimport)`/`__declspec(dllexport)` on non-msvc compilers anymore. CMake solves this problem when generating an export header, but I would rather have it leave NN_EXPORT as a conditional of `_WIN32`, not `NN_HAVE_MSVC` (or maybe let it use `__attribute__((dllimport))` on GNU compilers).

JackDunaway: Thank you for chiming in! My assumptions were incorrect. I'll revert back to using `_WIN32`

PR #720, opened 2016-05-16
Staticlib

gdamore:


CMakeLists.txt
```diff
@@ -123,6 +123,10 @@ else ()
     message (AUTHOR_WARNING "${ISSUE_REPORT_MSG}")
 endif ()
 
+if (NN_STATIC_LIB)
+    add_definitions (-DNN_STATIC_LIB)
+endif ()
+
```
JackDunaway: It's not particularly obvious why this fixes the issue; the reason is because the definition is applied prior to adding tests/tools. The reason it didn't work properly before is because this defintion only applied to the `src` directory -- meaning, effective only the library itself, but none of the companion executables within this project.

CMakeLists.txt
```diff
@@ -157,14 +161,6 @@ if (WIN32)
...
     set(NN_REQUIRED_LIBRARIES ${NN_REQUIRED_LIBRARIES} mswsock)
     set(NN_REQUIRED_LIBRARIES ${NN_REQUIRED_LIBRARIES} advapi32)
-
-    # Ensure tests/tools built as EXE choose symbol visibility appropriate for
-    # linking the library, whether static or dynamic.
-    if (NN_STATIC_LIB)
-        add_definitions (-DNN_WIN32_LIB_LINK)
-    else ()
-        add_definitions (-DNN_WIN32_DLL_LINK)
-    endif ()
```
JackDunaway: I agree removing these new symbols.

PR #726, opened 2016-05-17
fixes #725 docs silently fail if asciidoctor is not present

### LINKED ISSUE #725: docs silently fail if asciidoctor is not present
Currently, if `asciidoctor` is not present, the default behavior is to throw a warning. `NN_ENABLE_DOC` is a boolean option with two valid values: `ON` or `OFF`

I propose to add a third option, `DEFAULT`, which would inherit the current semantics of the current default `ON` -- a warning on failure.

The new `ON` would throw a fatal error if `asciidoctor` is not present.

The considerations here are:
- provide the same default behavior as current -- a best-effort try to build the docs ...
- ... yet provide a new option that will abort on error

The main reason is to alert CI to any such problems building the documentation, such as if/when something about `asciidoctor` is changed.

JackDunaway:


CMakeLists.txt
```diff
@@ -90,12 +90,14 @@ endif()
...
 
 option (NN_STATIC_LIB "Build static library instead of shared library." OFF)
-option (NN_ENABLE_DOC "Enable building documentation." ON)
 option (NN_ENABLE_GETADDRINFO_A "Enable/disable use of getaddrinfo_a in place of getaddrinfo." ON)
 option (NN_TESTS "Build and run nanomsg tests" ON)
 option (NN_TOOLS "Build nanomsg tools" ON)
 option (NN_ENABLE_NANOCAT "Enable building nanocat utility." ${NN_TOOLS})
 
+set (NN_ENABLE_DOC "DEFAULT" CACHE STRING "Enable building documentation.")
+set_property (CACHE NN_ENABLE_DOC PROPERTY STRINGS "ON" "OFF" "DEFAULT")
```
gdamore: does this artificially limit valid values by any chance?  (So for example, can I do cmake -DNN_ENABLE_DOC=yes today  instead of =ON ?

CMakeLists.txt
```diff
@@ -224,19 +226,26 @@ if (NN_ENABLE_NANOCAT)
...
+        set (NN_BUILD_DOC ON)
+    elseif (NN_ENABLE_DOC STREQUAL "DEFAULT")
+        message (WARNING "Could not find asciidoctor; skipping docs.")
+        set (NN_BUILD_DOC OFF)
+    else ()
+        message (FATAL_ERROR "asciidoctor is not installed. (hint: use -DNN_ENABLE_DOC=DEFAULT to suppress this error)")
     endif ()
+else ()
+    message (FATAL_ERROR "Unknown option: NN_ENABLE_DOC=${NN_ENABLE_DOC} (hint: use ON|OFF|DEFAULT)")
 endif ()
```
gdamore: Ah, I see, you have explicit checks here. Are ON/OFF sufficient to emulate what CMake does today?

README
```diff
@@ -30,7 +30,7 @@ Prerequests
 3. Documentation (optional)
    * asciidoctor (http://asciidoctor.org/) available as `asciidoctor`
    * If not present, docs are not formatted, but left in readable ASCII
-   * Also available on-line at http://nanomsg.org/documentation
+   * Also available online at http://nanomsg.org/documentation
```
gdamore: Huh.  I prefer with the dash, but I'm stodgy apparently:  http://grammarist.com/spelling/online-on-line/

README
```diff
@@ -12,10 +12,10 @@ applications communicate to form a single distributed application.
```
gdamore: The changes in README look good.  Can we push this in a separate bug fix/commit?

.travis.yml
```diff
@@ -22,7 +22,7 @@ script:
   - mkdir build
```
gdamore: Changes in travis and appveyor are good.  Some questions pending about the CMakeLists changes though.  (See below).

CMakeLists.txt
JackDunaway: `yes` *might* work, but the nominal values are `OFF` or `ON` for options: https://cmake.org/cmake/help/v2.8.7/cmake.html#command:option

So, it *might* be limiting to anyone who has used non-standard boolean values, but the key here is that we're not boolean anymore -- it's an enumerated set of values.

The failure mode for such case mutating forward is fail-safe by throwing a fatal error with a hint.

gdamore: Actually those docs say, and I quote:

True if the constant is 1, ON, YES, TRUE, Y, or a non-zero number. False if the constant is 0, OFF, NO, FALSE, N, IGNORE, "", or ends in the suffix '-NOTFOUND'. Named boolean constants are case-insensitive. If the argument is not one of these constants, it is treated as a variable:

gdamore: I'd only treat the positive cases there, and let the negative cases fall through to false if *anything* else (i.e. it doesn't concretely match one of the positive values)

README
reqshark: unhyphenated removes a character without meaning, so ya I'm gonna have to vote for `online`

gdamore:
With the conflicts and such, I'm going to close this.  Please resubmit if this is important, but honestly I'm much more focused on nng at present.

PR #755, opened 2016-06-04
fixes #754 CMake may not find InitializeConditionVariable

### LINKED ISSUE #754: CMake may not find InitializeConditionVariable
It seems that on some environments, CMake with MinGW-w64 fails to find InitializeConditionVariable, even though the host supports it.

This seems to happen because the CMake distributed with MSYS2 lies when providing the WINNT version, which makes MinGW-w64 believe it's running on windows XP.

Snaipe:
This explicitely tells the compiler that we target Windows Vista and later. Symbol checks on earlier versions of windows will fail regardless, since the linker won't be able to find the symbol even if it's declared.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -108,6 +108,10 @@ elseif (CMAKE_SYSTEM_NAME MATCHES "Windows")
     set (NN_HAVE_WINSOCK 1)
     add_definitions (-DNN_HAVE_WINDOWS)
     add_definitions (-D_CRT_SECURE_NO_WARNINGS)
+
+    # Target Windows Vista and later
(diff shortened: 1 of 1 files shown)


gdamore:
can you point me at relevant information from MSDN?

Snaipe:
See https://msdn.microsoft.com/en-us/library/aa383745(v=vs.100).aspx and https://msdn.microsoft.com/en-us/library/6sehtctf.aspx

Explicitely setting a WINNT version is encouraged when targetting specific features.

gdamore:
thanks. this was helpful. will merge when back at home. :)

PR #768, opened 2016-06-13
NN_MAXTTL missed in nn_symbol

borisz:
Looks like NN_MAXTTL is not exported


gdamore:
Good fix. I'll rebase and merge.

Issue #756, opened 2016-06-04
REP socket corrupts itself when REQ socket crashes


Snaipe:
I have been chasing this bug for a while now, and finally have been able to isolate it out of my project.

In a context where multiple REQ clients connect to one REP server and send data back and forth, if data is flowing from one of the clients to the server while another client dies unexpectedly, the REP socket corrupts itself and starts sending nonsense back to the other REQ clients.

In a situation with 2 clients and 1 server, this would look like this:

```
Server spawns, REP socket bound to tcp://127.0.0.1:4224
Client 1 spawns, REQ socket connects to tcp://127.0.0.1:4224
Client 2 spawns, REQ socket connects to tcp://127.0.0.1:4224
Server calls recv
Client 1 calls send
Client 2 calls send
Server receives the message from Client 1, responds
Client 2 calls recv
Client 1 calls recv
Client 1 receives the response from Server
Server calls recv
Client 1 crashes
Server receives the message from Client 2
Server calls send
Client 2 receives garbage, and aborts.
```

This seems to manifest itself on Windows 10 with MSVC 2015 Update 2, built with both MSVC and MinGW-w64 (it's reproducible on Appveyor, notably). Otherwise, the issue may only happen sporadically (for instance, I cannot reproduce it on my Windows 10 VM with MSVC 2015 Update 2 for some reason).

Here is a [self-contained example that triggers the bug in question](https://ptpb.pw/jdx2).

The example itself could be minimized -- I only happened to get it to work to prove that the issue exists on nanomsg's side, and not on mine.

More tests might be needed to determine whether the socket can be recovered afterwards, or if the sent data might contain fragments of the original payload, but as I can only work on it over RDP on Appveyor because I otherwise cannot reproduce it locally, this is excruciatingly painful to do.

Snaipe:
To add more info since the situation described above is approximative, here's the [log of a reproduction of the bug](https://ptpb.pw/-jbF) with 5 clients (4 process + 1 for the child sink) and 10 messages per client. I would have liked to be able to reproduce it with even less verbosity, but this seems to be the best that I can provide at the moment.

The above log was obtained with a [slightly modified version of the example ligned in the first post](https://ptpb.pw/W7tE)

gdamore:
I have not been able to reproduce inside my Windows 8.1 VM.  Which makes debug hard.  Doing anything inside Windows is a substantial PITA.  Any chance at repro using POSIX code?

gdamore:
[...]

PR #778, opened 2016-06-27
demonstrating `nn_close` bug with cancelled requests; applying `test_setsockopt` to existing tests

JackDunaway:
This test demonstrates an issue which could be related to #756

### DIFF
--- a/src/transports/tcp/btcp.c
+++ b/src/transports/tcp/btcp.c
@@ -118,11 +118,13 @@ int nn_btcp_create (struct nn_ep *ep)
     end = addr + strlen (addr);
     pos = strrchr (addr, ':');
     if (pos == NULL) {
+        nn_free (self);
         return -EINVAL;
     }
     ++pos;
     rc = nn_port_resolve (pos, end - pos);
     if (rc < 0) {
+        nn_free (self);
         return -EINVAL;
     }
 
@@ -134,6 +136,7 @@ int nn_btcp_create (struct nn_ep *ep)
     /*  Parse the address. */
     rc = nn_iface_resolve (addr, pos - addr - 1, ipv4only, &ss, &sslen);
     if (nn_slow (rc < 0)) {
+        nn_free (self);
         return -ENODEV;
     }
 
@@ -151,6 +154,7 @@ int nn_btcp_create (struct nn_ep *ep)
 
     rc = nn_btcp_listen (self);
     if (rc != 0) {
+        // I suspect we might need to do nn_free here.
         return rc;
     }
 
--- a/src/transports/tcp/ctcp.c
+++ b/src/transports/tcp/ctcp.c
@@ -150,10 +150,12 @@ int nn_ctcp_create (struct nn_ep *ep)
 
     /*  Parse the port. */
     if (!colon) {
+        nn_free (self);
         return -EINVAL;
     }
     rc = nn_port_resolve (colon + 1, end - colon - 1);
     if (rc < 0) {
+        nn_free (self);
         return -EINVAL;
     }
 
@@ -162,6 +164,7 @@ int nn_ctcp_create (struct nn_ep *ep)
     if (nn_dns_check_hostname (hostname, colon - hostname) < 0 &&
           nn_literal_resolve (hostname, colon - hostname, ipv4only,
           &ss, &sslen) < 0) {
+        nn_free (self);
         return -EINVAL;
     }
 
@@ -169,6 +172,7 @@ int nn_ctcp_create (struct nn_ep *ep)
     if (semicolon) {
         rc = nn_iface_resolve (addr, semicolon - addr, ipv4only, &ss, &sslen);
         if (rc < 0) {
+            nn_free (self);
             return -ENODEV;
         }
     }


JackDunaway:
Here's a typical stack trace of the crash itself:

````
>	reqrep.exe!nn_fsm_feed(nn_fsm * self=0xdddddddd, int src=0xdddddddd, int type=0xdddddddd, void * srcptr=0xdddddddd) Line 82	C
 	reqrep.exe!nn_fsm_event_process(nn_fsm_event * self=0x00e955d8) Line 77	C
 	reqrep.exe!nn_ctx_leave(nn_ctx * self=0x00e8aa10) Line 91	C
 	reqrep.exe!nn_ctx_leave(nn_ctx * self=0x00e8c3c8) Line 92	C
 	reqrep.exe!nn_worker_routine(void * arg=0x00b38220) Line 219	C
 	reqrep.exe!nn_thread_main_routine(void * arg=0x00b3822c) Line 30	C
```


and here's the parallel thread that's the source of the crash:

```
>	reqrep.exe!nn_sem_wait(nn_sem * self=0x00e8c618) Line 122	C
 	reqrep.exe!nn_sock_term(nn_sock * self=0x00e8c388) Line 216	C
 	reqrep.exe!nn_close(int s=0x00000001) Line 531	C
 	reqrep.exe!test_close_impl(char * file=0x00b15de0, int line=0x000000ba, int sock=0x00000001) Line 120	C
 	reqrep.exe!main(...) Line 186	C
```

Snaipe:
I'll try to reproduce this crash using separate processes.

JackDunaway:
Are you able to repro as-submitted? If not, try increasing `i` beyond 100 here: https://github.com/nanomsg/nanomsg/pull/778/files#diff-dd8deeacecfc1adc4bb20193085968acR177

JackDunaway:
The initial push passed on both Travis and AppVeyor, but I've restarted both.

On the first retry, here's the failure demonstrated on Travis with GCC on Ubuntu ([link to Travis log](https://travis-ci.org/nanomsg/nanomsg/jobs/140529979#L760)):
```
9/42 Test #11: reqrep ...........................***Exception: Other  0.23 sec
/home/travis/build/nanomsg/nanomsg/build/libnanomsg.so.5.0.0(+0x2f30c)[0x7f5c6460c30c]
/home/travis/build/nanomsg/nanomsg/build/libnanomsg.so.5.0.0(nn_fsm_feed+0x46)[0x7f5c645f366a]
/home/travis/build/nanomsg/nanomsg/build/libnanomsg.so.5.0.0(nn_fsm_event_process+0x67)[0x7f5c645f3622]
/home/travis/build/nanomsg/nanomsg/build/libnanomsg.so.5.0.0(nn_ctx_leave+0x11c)[0x7f5c645f34aa]
/home/travis/build/nanomsg/nanomsg/build/libnanomsg.so.5.0.0(nn_ctx_leave+0x12f)[0x7f5c645f34bd]
/home/travis/build/nanomsg/nanomsg/build/libnanomsg.so.5.0.0(+0x1b717)[0x7f5c645f8717]
/home/travis/build/nanomsg/nanomsg/build/libnanomsg.so.5.0.0(+0x1e508)[0x7f5c645fb508]
/lib/x86_64-linux-gnu/libpthread.so.0(+0x7e9a)[0x7f5c643c7e9a]
/lib/x86_64-linux-gnu/libc.so.6(clone+0x6d)[0x7f5c640f536d]
Unexpected state: state=7 source=27713 action=5 (/home/travis/build/nanomsg/nanomsg/src/transports/inproc/sinproc.c:486)
```

MASSIVE thank you to @gdamore for the new stack trace feature; test output on failure is now useful!

For convenience, here is a search of this repository for `27713` which is definitely going to be useful debugging this issue: https://github.com/nanomsg/nanomsg/search?utf8=%E2%9C%93&q=27713

JackDunaway:
15 builds later, restarted at arbitrary times throughout the day for the past few days to try to rule out cluster utilization skewing results, and still no repro on AppVeyor.

I'm investigating another repro scenario that's more repeatable, and perhaps even deterministic.

@gdamore, are you able to repro on your machine?

gdamore:
I think I've seen several issues like this around inproc.  The linking of state machines on the sender and receiver is really problematic; probably what I need to do is just rewrite inproc from the ground up.  I've come close to this several times (started a few of them).  Each pair of connected inproc sockets would then be serviced by either one or two worker threads, that would handle everything.

If this reproduces in any of the other transports I would like to know.  At the moment I'm just considering it another one of the various inproc related problems.

JackDunaway:
I've actually just today had time to debug this, and am narrowing in on a fix, once and for all. It is currently looking promising to solve the `NN_QUEUE_NOTINQUEUE` as well.

The key points seems to be:

1. Enable `nn_mutex` to be recursive (hold more than one lock if nested in the same thread)
2. Ensure `nn_fsm_raiseto` is only called with a lock on its destination `ctx`, and finally
3. Ensure external events in the `eventsto` queue are handled while the owning socket remains locked.

Stay tuned ...

gdamore:
I'm pretty opposed to making nn_mutex recursive.  This will cause an entire new class of bugs to be possible.

JackDunaway:
Perhaps, but it provides the ability to solve many bugs that still plague us.

For that reason, I've already changed the function prototype to be `void nn_mutex_init (nn_mutex_t *self, int recursive)` (it used to be `void nn_mutex_init (nn_mutex_t *self)`), where every callsite in the codebase to `nn_mutex_init` sets `recursive=0` except for the mutex that protects `ctx` objects -- it can be recursive.

I'm still in early testing, but plan to have some more concrete commits tomorrow. This PR will remain a sandbox for testing.

JackDunaway:
https://github.com/nanomsg/nanomsg/pull/778/commits/2102d0616aee38d5e4d1b711526518d1af65f112 is a new test `reqrep_cancellation` dedicated to this bug (rather than previously tacking on to the `reqrep` test). Here's a link to its first repro on Travis CI: https://travis-ci.org/nanomsg/nanomsg/jobs/142891644#L829

gdamore:
almost every case i have seen for recursive mutex use over the years has lead to deadlocks and harder to debug problems. 

recursive mutexes here wont solve our problems. they may give a bit of immediate palliative relief at best.

PR #786, opened 2016-07-30
nanocat: implement to launch an executable for responder

jiazhang0:
-E option allows to launch an executable for responder to handle the input from nn_recv() and output the resulting contents through nn_send().

Internally, a subprocess is used to launch the executable and its standard input and output are bound to the NN socket over a pipe.

Note this option currently is only avialable for Linux environment.

Signed-off-by: Lans Zhang <jia.zhang@windriver.com>


jiazhang0:
Will fix the CI build failure in V2. Closed.

PR #788, opened 2016-08-06
fix inconsistent indentation

QXSoftware:
When I open the src code via SublimeText, I find some inconsistent indentation in src/aio/worker_posix.inc, so I reformatted a little piece of it. I hope this will be useful.


gdamore:
Thanks.  Will probably merge this soon.

PR #794, opened 2016-08-13
Fix FQDN

drunk-owl:
nn_dns_check_hostname returns EINVAL for domain names with the dot at the end.
e.g. "example.org." were not accepted, although it is valid FQDN.

### DIFF
--- a/src/transports/utils/dns.c
+++ b/src/transports/utils/dns.c
@@ -44,11 +44,6 @@ int nn_dns_check_hostname (const char *name, size_t namelen)
 
         /*  End of the hostname. */
         if (namelen == 0) {
-
-            /*  The last label cannot be empty. */
-            if (labelsz == 0)
-                return -EINVAL;
-
             /*  Success! */
             return 0;
         }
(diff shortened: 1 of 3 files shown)


gdamore:
Looking at these changes, I don't understand why we are bothering to validate hostnames at all.  We should be relying upon the resolver to do that for us.  Doing this ourselves in library space is actually rather poor form.  I'll probably just gut all this outright.

gdamore:
So the issue here is that we don't resolve the name up front -- because DNS can be down or a name might not have been registered yet or something.  This lets the software overcome temporary name resolution problems.

Arguably reliance on this is is somewhat unfortunate -- really for the vast majority of cases if an unresolvable DNS name is given to nn_connect(), you'd really just like to have a hard failure, rather than trying to muddle on.  In the vast vast majority of cases, immediate fail-fast is superior to a situation where the thing never works.

This actually relates to the desire for an early failure at connect time too.   Probably this needs to be configurable, but I need to think about this somewhat.

gdamore:
For now, I"m going to just accept this, because clearly a . at the end is a valid FQDN.  There may be other considerations, but this PR is good.

PR #801, opened 2016-09-08
test case for bug796

gdamore:
This is not a fix, but just a test case to hopefully demonstrate bug796 on windows.


gdamore:
Its unclear that the root problem exists, so I'm closing this for now.

PR #807, opened 2016-09-20
Fix pkg-config Cflags definition

mdcb:


Snaipe:
I don't think this fixes anything. Rather, this conflicts with the [documentation](http://nanomsg.org/v1.0.0/nn_socket.3.html). Includes should be `<nanomsg/nn.h>`, not `<nn.h>`.

mdcb:
Ok!

PR #826, opened 2016-10-26
update from nanomsg

wwjiang007:


gdamore:
THere's nothing here.

PR #834, opened 2016-11-22
Update poll.c to check correct revents for POLLOUT

matthewaveryusa:
Update poll.c to check correct revents for POLLOUT


gdamore:
I think you may be somewhat confused, and understandably so.  While we express a "natural" poll interface, under the hood we *always* use readable polls on the file descriptor, because we write to the file descriptor (from other threads) to wake this.  Really it would be much nicer & cleaner if we used condition variables instead here, and that is in my plan of action.

At any rate, your suggested change is actually quite incorrect, as you can see by the failures in the test suite.  (Windows uses a different mechanism entirely, and that's why it passed.)

PR #851, opened 2016-11-29
Make the pkgconfig.in libdir variable reference the real installed library dir

andrew-buckley:
Currently, the `libdir` is hard-coded to be "${prefix}/lib". However depending on the system onto which the library is installed, the `CMAKE_INSTALL_LIBDIR` may not be "lib". This sets the `libdir` to be the same directory where the libraries are actually installed.

If there is a better fix for this, or if there's some reason this shouldn't go through, please let me know.

Thanks!

### DIFF
--- a/src/pkgconfig.in
+++ b/src/pkgconfig.in
@@ -1,7 +1,7 @@
 prefix=@CMAKE_INSTALL_PREFIX@
 exec_prefix=${prefix}
 includedir=${prefix}/include
-libdir=${prefix}/lib
+libdir=${prefix}/@CMAKE_INSTALL_LIBDIR@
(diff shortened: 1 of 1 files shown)


gdamore:
Looks like a good fix... I'll merge this soon.

andrew-buckley:
Great, thanks

andrew-buckley:
Hi @gdamore, just wanted to check if this is still on track to be checked in. 

Thanks again.

andrew-buckley:
@damonbarry if you would also like to see the status of this

gdamore:
Sorry didn't do this earlier...

Issue #774, opened 2016-06-22
Memory leak when a socket fail to bind


takuyozora:
### ISSUE

In some case, the memory allocated here in `src/transport/tcp/btcp.c` : 

``` c
/* Allocate the new endpoint object. */ 
self = nn_alloc (sizeof (struct nn_btcp), "btcp"); 
alloc_assert (self);
```

is not freed. Here is the valgrind report : 

``` c
16: ==13441== Thread 1:
16: ==13441== 747,720 bytes in 1,005 blocks are definitely lost in loss record 56 of 56
16: ==13441==    at 0x4C2ABD0: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
16: ==13441==    by 0x5360572: nn_alloc_ (in /usr/lib/libnanomsg.so.0969988-dirty)
16: ==13441==    by 0x5378D53: nn_btcp_create (in /usr/lib/libnanomsg.so.0969988-dirty)
16: ==13441==    by 0x537CA2D: nn_tcp_bind (in /usr/lib/libnanomsg.so.0969988-dirty)
16: ==13441==    by 0x5354848: nn_ep_init (in /usr/lib/libnanomsg.so.0969988-dirty)
16: ==13441==    by 0x5359121: nn_sock_add_ep (in /usr/lib/libnanomsg.so.0969988-dirty)
16: ==13441==    by 0x5356FF2: nn_global_create_ep (in /usr/lib/libnanomsg.so.0969988-dirty)
16: ==13441==    by 0x5355F01: nn_bind (in /usr/lib/libnanomsg.so.0969988-dirty)
```
### REPRODUCE

``` c
int s1 = nn_socket(AF_SP, NN_PULL);
int s2 = nn_socket(AF_SP, NN_PULL);

int rc1 = nn_bind(s1, "tcp://127.0.0.1:5010");
int rc2 = nn_bind(s2, "tcp://127.0.0.1:5010");

nn_close(s1);
nn_close(s2);
```

I was trying to repair myself [here](https://github.com/nanomsg/nanomsg/pull/773) but I really don't know the internal logic of the lib so I wasn"t able to find the right fix. 

O. Radisson

gdamore:
Confirmed.  It appears that the parent listening structure is not freed.  I'm analyzing now.

hwanren:
I think the cppcheck report matches with the valgrind report w.r.t. nn_alloc_:

[...]

PR #852, opened 2016-12-06
Fix memory leak when binding or connecting to bad URL

bill1600:
In TCP, when connecting or binding to an invalid URL, you get a memory leak, because nn_close not called after editting of the URL fails.

### DIFF
--- a/src/transports/tcp/btcp.c
+++ b/src/transports/tcp/btcp.c
@@ -118,11 +118,13 @@ int nn_btcp_create (struct nn_ep *ep)
     end = addr + strlen (addr);
     pos = strrchr (addr, ':');
     if (pos == NULL) {
+        nn_free (self);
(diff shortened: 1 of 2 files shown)


gdamore:
I will review and update... stay tuned.

takuyozora:
I confirm that the PR fix my issue #774

gdamore:
Rebased onto master:

commit ee99202f84f8d0e8ebd1c1722aa8a682f54919d0
Author: Bill Williams <bill_williams@cable.comcast.com>
Date:   Thu Dec 1 15:35:44 2016 -0800

    Fix memory leak when binding or connecting to bad URL

diff --git a/src/transports/tcp/btcp.c b/src/transports/tcp/btcp.c
index 37541f1..4fe8b58 100644

PR #857, opened 2016-12-19
drop a typo

lu-zero:


gdamore:
I'm actually pleased someone is reading that work...

PR #865, opened 2017-03-08
fix memory leak on nn_bind when error EADDRINUSE returned

bill1600:
Fixes memory leak that happens when nn_bind fails returning EADDRINUSE.


gdamore:
This is crashing (segfaulting) in the tcp test.

bill1600:
Sorry.  Tests passed in my environment (Fedora 19).

bill1600:
Will resubmit later.

bill1600:
Don't commit this.  I just need to use travis to test.

bill1600:
Tests pass.  This is ready to be merged now.

gdamore:
So, the state machine changes look good, but I'd like to not include the debugging stuff; I realize it was useful to you, the debug stuff needs to be more completely thought out if we're gonna do that, and should be handled as a separate PR.

Ultimately, we would like to get this to just one commit as well.

bill1600:
OK. I will remove the debug.

bill1600:
Removed debug. Should be able to merge now.

src/aio/ctx.c
```diff
@@ -64,9 +64,9 @@ void nn_ctx_leave (struct nn_ctx *self)
     }
 
     /*  Notify the owner that we are leaving the context. */
-    if (nn_fast (self->onleave != NULL))
+    if (nn_fast (self->onleave != NULL)) {
```
gdamore: This is a style change.  While I personally like having the extra explicit braces, it isn't intrinsic to this change.  We should remove this for now.  (I'm avoiding style changes to the code base, even though I hate the current style.)

src/utils/err.c
```diff
@@ -210,3 +210,4 @@ void nn_win_error (int err, char *buf, size_t bufsize)
 }
 
 #endif
+
```
gdamore: Please remove this -- we don't want blank lines at the end of files.

tests/tcp.c
```diff
@@ -26,9 +26,11 @@
 #include "../src/pair.h"
 #include "../src/pubsub.h"
 #include "../src/tcp.h"
+#include "../src/utils/err.h"
```
gdamore: This file has only whitespace changes now.  These are not intrinsic to you changes, so lets undo this.

gdamore:
So I've asked for cleanup of some unrelated changes to files -- probably these are things you changed and then changed back as you were trying things out.  We want to boil this to just the change that we want.

Also, I'd like you to submit this as a new PR, with the commits squashed into a single commit, with the commit comment as "Fixes memory leak that happens when nn_bind fails returning EADDRINUSE"  - that way we can see the single issue in one swoop.

(Even better, if we have a github issue already filed for this, the single commit comment would be a line of "fixes #<issue number> <issue synopsis>"   -- see the git commit history for examples of that.

bill1600:
OK, I undertand.
I will clean it up a resubmit as a new PR.

bill1600:
Will resubmit clean.

PR #868, opened 2017-03-16
Allow nanomsg to compile under Visual Studio 2008

crule42:
I had a need to use nanomsg to compile with Visual Studio 2008. The only issue was with the stdint.h include file that does not exist in VS2008. I found an appropriately licensed version and have included that. It only uses the internal version if compiled using VS2008 or below though I have only tested with Visual Studio 2008.

This also includes a change so that when the Visual Studio project file for nanomsg is created using CMake, the files directly in src are shown in a "Source" folder in Visual Studio. This prevents seeing all the header files in their own folder (e.g. nn.h/nn.h).


skyformat99:
Change too much momentum, against such changes,
and affect compile process on the Linux under.

Simple processing method:   copy stdint.h  to include folder of VS2008.

gdamore:
So I'm unlikely to accept this.  stdint.h is provided in Visual Studio 2012 and later, and these compilers are *free*, and support the same platforms as VS2008.  I frankly can't see any reason to worry about supporting these older compilers.

If you have a problem where you have to solve this, as indicated, just copy a stdint.h from some likely place into your include directories.

crule42:
<shrug> It's your repo. 
I have a requirement to run Visual Studio 2008 and wanted to use the library. Thought others might want the capability also. 

As far as simply copying stdint.h to the Visual Studio directory, that's fine for a one-off machine, but having to remember to constantly do that for new machines is not worth the hassle when it can be accomplished in a self contained library.  The way this was implemented it also compiles fine under linux.

And Visual Studio 2012 is not free, it's only free for non-commercial use which is not my case.

I can remove the pull request if that's what you want. Just let me know.

gdamore:
So no, I'm not going to accept this PR, sorry.  I would however like to understand why using older VS 2008 (its pretty darn old now) is a requirement?  These days I'm using VS 15 (2017) Community edition.  It's free for non-enterprise users (including individual commercial users).  Enterprise users (those with > 250 PCs or $1M in revenue) have to pay for it if they want to create commercial apps, but ... that's not a typical concern.  If you work for a big company, you can afford to pay for your compiler to build your commercial apps.  (I'm not interested in spending my effort sustaining code to enable corporations who want to cheap skate their commercial development process.)

Again this doesn't affect *any* open source development, nor does it affect smaller companies or individuals.  Those folks should just pick up VS community and stop whining. :-)

Issue #777, opened 2016-06-26
SIGABRT on connecting to inproc with type PAIR


djc:
In

```
0   libnanomsg.5.0.0.dylib              0x0000000103cbbf8c nn_sinproc_handler + 1279
1   libnanomsg.5.0.0.dylib              0x0000000103cad227 nn_ctx_leave + 164
2   libnanomsg.5.0.0.dylib              0x0000000103cad234 nn_ctx_leave + 177
3   libnanomsg.5.0.0.dylib              0x0000000103cad234 nn_ctx_leave + 177
4   libnanomsg.5.0.0.dylib              0x0000000103cac441 nn_sock_add_ep + 163
5   libnanomsg.5.0.0.dylib              0x0000000103ca9ed2 nn_connect + 72
```

See https://github.com/nanomsg/nnpy/issues/17.

gdamore:
So this is PAIR.  Basically we are saying that the socket is already paired, and we can't add additional ones.  Sadly we are aborting here rather than just returning a sane error code.  Stay tuned.

gdamore:
This seems to be tied to inproc as well.

gdamore:
I'm pretty sure the above fixes the issue.  Please feel free to verify for yourself.  I'll probably merge this shortly.

Issue #328, opened 2014-11-03
IOCP doesn't report disconnections sometimes


sustrik:
This causes unexpected behavour in the cases like this one:

    sb = test_socket (AF_SP, NN_PAIR);
    test_bind (sb, SOCKET_ADDRESS);
    s1 = test_socket (AF_SP, NN_PAIR);
    test_bind (s1, SOCKET_ADDRESS);
    sc = test_socket (AF_SP, NN_PAIR);
    test_connect (sc, SOCKET_ADDRESS);
    nn_sleep (100);
    test_send (sb, "ABC");
    test_recv (sc, "ABC");
    test_close (sb);
    nn_sleep (300);
    // Following line block because re-connection haven't happened  because
    // failure of the original connection wasn't reported by IOCP.
    test_send (s1, "ABC");
    test_recv (sc, "ABC");
    test_close (sc);
    test_close (s1);

The possible solution is to have an outstanding request to read 0 bytes of OOB data at all times. It will get triggered only when disconneciton happens. See here:

http://stackoverflow.com/questions/11210995/how-to-detect-disconnection-rst-fin-using-windows-iocp

gdamore:
This is true for POSIX systems too -- if you don't attempt to do I/O, you might never notice a disconnect or loss of a connection.  I found the read() (of any data) to be required on POSIX systems with mangos.  We should probably do this in the protocol handers -- even write-only protocol handlers should probably set up a reader on the underlying transport, just to catch this condition.

gdamore:
I think I've found workarounds for this.  IOCP fails, but we can close the remote peer, which causes a POLLHUP condition to arise, which makes everyone happy.  So far this seems to work.

Martin, do you have a PR with the formal test case for this?

gdamore:
So I made one, from your code, obviously. :-)

gdamore:
So, actually, the test so far looks pretty good; I think PR #525 that I integrated actually solves this.  The test code is passing at least.

gdamore:
(You'll note that the only thing I merged here was your test case.)

PR #874, opened 2017-04-17
Moving 'tests/bug328.c' to 'pair_tcp.c'. Moving 'tests/bug777.c' into 'inproc_noclose.c'.

metadings:


metadings:
Now if readded 'bug777.c', so I did just move this into 'inproc_noclose.c'...

gdamore:
What purpose does this PR serve?

metadings:
Just to move names `bug777` (to `inproc_close.c`) and `bug328`(to `pair_tcp.c`).

gdamore:
but what was wrong with the old names.  these were tests for specific bugs described in the given bugids.

metadings:
Well, I thought it is bad, if you have a `build` directory with new commands `bug328`, `bug777`.
I don't know about bugids, where are these bugids defined?

Snaipe:
Those are just tests, the binaries are not installed on the target system. The bug IDs refer to past github issues, in this case #328 and #777. If the problem is about polluting the build directory with seemingly unrelated binaries, I'd suggest putting the binaries produced by `add_libnanomsg_test` into a `test` subdirectory.

metadings:
well... Should I name them `issue*.c`?

PR #875, opened 2017-04-21
Invalidates src chunkref in nn_chunkref_mv.

thisco-de:
Assuming that the chunk hold by src exceeds NN_CHUNKREF_MAX bytes,
i.e. uref[0] == 0xff, then only a reference (resp. pointer) to the
piece of memory storing this chunk is moved to dst. Then, whenever
one calls nn_chunkref_term on one of the instances (src or dst) and
later on the other, the application will crash due to the request
of freeing an unallocated piece of memory.
For this reason, it is safer to somehow invalidate src in a call to
nn_chunkref_mv by setting its u.ref[0] to 0x00, such that a call to
nn_chunkref_term will not attempt to free memory.

### DIFF
--- a/src/utils/chunkref.c
+++ b/src/utils/chunkref.c
@@ -97,6 +97,7 @@ void nn_chunkref_mv (struct nn_chunkref *dst, struct nn_chunkref *src)
 {
     memcpy (dst, src, src->u.ref [0] == 0xff ?
         (int)sizeof (struct nn_chunkref_chunk) : src->u.ref [0] + 1);
+    src->u.ref[0] = 0x00;
 }
 
 void nn_chunkref_cp (struct nn_chunkref *dst, struct nn_chunkref *src)


thisco-de:
Sorry, can you please be a little more specific on what you mean by saying "unconditionally"? Perhaps the context from which I am looking at this piece of code it is a bit to narrow, but as far as I understand it, the tag as well as the pointed to address get copied to dest. By setting the tag (resp. u.ref[0]) to 0, this semantically would be equivalent to "there is no data", shouldn't it? nn_chunkref_getchunk does exactly the same thing.

skyformat99:
By setting the tag (resp. u.ref[0]) to 0, Were it to do so, it easily would have made a memory leak。

gdamore:
Right -- so the problem here is that no validation is done beforehand, and yes, this means that the data pointed to can get lost, leading to a leak.

Do you actually have a reproducible crash that this fixes?

gdamore:
So I've spent a bunch more time looking at this.

The issue here is a confusion about the semantics of nn_msg_mv(), which is the only consumer of nn_chunkref_mv.

nn_msg_mv() *moves* the contents of the message from one struct to another.  It *leaves* the old message structure in an intdeterminate state.  As you note, it would be a gross error to call nn_msg_free() on the source message after this. 

The existing nn_msg_mv() callers simply reinitialize the source message structure.

Your change, while not precisely ... perfect, would actually make the source structure "safer", leaving a zero size behind, so that nn_msg_free() would be safe.  Having said that though, none of this API is public (for external consumption) -- it isn't documented anywhere at all, and while the existing semantic is probably confusing, it's not really worth fixing at this point.

The reason it isn't worth "fixing", is that the next generation library, nng, doesn't have this problem at all (it doesn't use chunks or chunkrefs at all!)  and I'm trying hard to minimize further investment in the stable version of libnanomsg.

So, I'm closing this as will not fix -- it isn't necessary to fix and doesn't represent an actual bug.

thisco-de:
I really have to say I am sorry for not giving any answers on this issue, as I first became too busy writing the project I based on nanomsg during the first half of the year and had a lot of other things in mind during the last week.

However, as far as I remember, it was exactly the semantic meaning of 'mv', which caused me some headache while implementing my project, until I digged deeper into the implementation after getting your feedback. I cannot really tell anymore, how I circumvented the problems I had while implementing the protocol for my project (you are right, nn_msg_msg() is not in the public interface, but you have to deal with it, when writing new protocols).

Soon I also noticed your re-implementation of nanomsg in nng, hence the issue was already of the table from my anyway. Hence, thanks for taking the time to research the issue and give me feedback. Looking forward to use nng in my projects some day! :+1:

PR #876, opened 2017-04-21
Sets pipebase's data-pointer to NULL during initialization.

thisco-de:
A freshly initialized pipe(base) does not have any additional data
assigned to it. However, a receiver of a pipe might not necessarily have
the knowledge, whether or not someone else might have set the pointer to
point to additional data or not. Since the pipe is not set to NULL
during initialization of itself, the pointer is likely to contain an
arbitrary value and thus may cause later checks by the receivers of the
pipe to fail. Therefore, the data-pointer should be initialized with
value NULL in nn_pipebase_init to prevent undefined behaviour or even
crashes.

### DIFF
--- a/src/core/pipe.c
+++ b/src/core/pipe.c
@@ -58,6 +58,7 @@ void nn_pipebase_init (struct nn_pipebase *self,
     self->instate = NN_PIPEBASE_INSTATE_DEACTIVATED;
     self->outstate = NN_PIPEBASE_OUTSTATE_DEACTIVATED;
     self->sock = ep->sock;
+    self->data = NULL;
     memcpy (&self->options, &ep->options, sizeof (struct nn_ep_options));
     nn_fsm_event_init (&self->in);
     nn_fsm_event_init (&self->out);


gdamore:
so the actual value of the data field at this point should be undefined; that is the meaning of the pipes data member is not set until something else sets this.  normally that should be done by the caller of this routine.  (code should not assume that an unassigned pipe has any value here - even NULL - unless they set it.)

Do you have specific bug cases where this is not being done?

thisco-de:
Simply think of two indepentent components where one has to check whether the other has already attached data to it or not. For the sake of a concrete example, take an implementation of a transport that might conditionally attach data or not (resp. different transport, such that some attach data others do not) and the protocol has to check for this. The TCP transport, for instance, does not nullify the data member after initialization (cf. nn_stcp_init), which - according to your argumentation - it should, if I got you right. However, I do not see the point in not describing the state of the object explicitly/accurately at initialization time (i.e. there is *no* data attached) and by this taking potentially error-prone assumptions out of the game entirely. For what reason? To save a single instruction?

gdamore:
You're missing the point here.  Undefined state is actually a valid situation here.  If a protocol needs to depend on this data member being NULL, then it should set it so.  This is protocol private data, and it isn't the framework's job to assume or impart any meaning to this; in this context NULL is no more meaningful an initialization value than say 0xFFFFFFFF or 354, or a pointer to a string containing the value "NULL".

If a protocol does not *use* this field, then it is free to leave it uninitialized.  Of course, if it *does*, then it had better do so.  Frankly, I want transports doing this themselves, then relying on the framework to do so for them.

This is more about design correctness than about saving an instruction.  (That said, yes, it can save a slight bit of work.)

Note that we are talking about *protocols*.  It is incorrect for a *transport* to set this member.  Transports are things like TCP.  Protocols are things like REQ/REP.

This member is entirely protocol-private.  Any protocol that calls nn_pipe_getdata() should first have called nn_pipe_setdata().

gdamore:
Note that all the protocols, except PAIR, both set and use the data field properly.  PAIR makes no reference this whatsoever, since it has no need to store protocol private data there.

gdamore:
I'm closing this pull request per the discussion above.

PR #882, opened 2017-05-11
fixes #118 NN_TCP_NODELAY doesn't cause TCP_NODELAY to be set

### LINKED ISSUE #118: NN_TCP_NODELAY doesn't cause TCP_NODELAY to be set.

urdh:


src/transports/tcp/atcp.c
```diff
@@ -192,6 +201,11 @@ static void nn_atcp_handler (struct nn_fsm *self, int src, int type,
                 nn_assert (sz == sizeof (val));
                 nn_usock_setsockopt (&atcp->usock, SOL_SOCKET, SO_RCVBUF,
                     &val, sizeof (val));
+                sz = sizeof (val);
+                nn_ep_getopt (atcp->ep, NN_TCP, NN_TCP_NODELAY, &val, &sz);
+                nn_assert (sz == sizeof (val));
+                nn_usock_setsockopt (&atcp->usock, IPPROTO_TCP, SO_RCVBUF,
```
gdamore: This is incorrect.  It is setting the SO_RCVBUF, not the TCP_NODELAY option.   I think you made a copy-pasta error.

gdamore:
Client side code (connect side) looks good.  Server side is using the wrong socket option.  Please fix and retest, then resubmit.  I'd appreciate an understanding of how you validated the test.

urdh:
I wrote and tested (manually observing traffic in Wireshark) the patch for work but reimplemented it at home, that's how the error snuck in there.

It would perhaps be appropriate to have a unit test that inspects the socket options (or something), but I'm not familiar with the nanomsg unit tests, and I don't see any obvious way to do that.

src/transports/tcp/atcp.c
urdh: Fixed.

PR #897, opened 2017-09-15
Fix buffer and memory leak issue in rpc_demo.c of demo directory

davesliu:
Fix bug of undefined array end and memory leak of greeting.


gdamore:
Looks good, thanks!

PR #898, opened 2017-09-18
fix memory leak issue of nn_bws_create

davesliu:


gdamore:
This is failing with segfaults on Windows.   It needs work before it can be merged.

src/transports/ws/bws.c
```diff
@@ -150,6 +153,7 @@ int nn_bws_create (struct nn_ep *ep)
 
     rc = nn_bws_listen (self);
     if (rc != 0) {
+        nn_free(self);
```
gdamore: This is almost surely incorrect.  The problem here is that we have the state machine setup and running, and a usock set up to use it as well.

Simply freeing is inadequate here, and leads to a segfault.

gdamore:
In fact, reading the entire code flow, this is all all wrong.

Because we've established the ep state, our bws_stop() entry point should be called eventually, and the endpoint should actually terminate cleanly.

The current approach in this PR is completely wrong.  I'm closing it.

PR #916, opened 2017-10-13
fixes #719 Need expanded AppVeyor build environment matrix

### LINKED ISSUE #719: Need expanded AppVeyor build environment matrix
In order to demonstrate support of earliest officially supported Visual Studio environment:

`Visual Studio 10 2010`
and
`Visual Studio 10 2010 Win64`

In order to test MinGW-64 support, potentially with `allow_failures` set on this one:

`MSYS Makefiles`

In order to test building static libraries, we need at least one environment with the following:

`cmake -DNN_STATIC_LIB=ON ..`

gdamore:
This incorporates changes that I think will cover both mingw-w64
and support for the very latest 2017 compiler and all the way back
to 2010.  We also test Release and Static builds.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -5,21 +5,89 @@ environment:
     CFLAGS: /MP
   matrix:
     # array of all environments used to test builds
+    # The latest Visual Studio 2017 is our "preferred" environment.
+    # We expect that it runs fastest too.
+
+    # MSys2, 64-bit
+    - GENERATOR: MSYS Makefiles
+      APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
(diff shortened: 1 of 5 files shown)


gdamore:
So it would appear that the Visual Studio 2017 image is unreliable when running tests.  I think the problem is likely that it is being run on different infrastructure, and consequently may have fewer resources available or other constraints.

I could spend a bunch of time tracking this down, but at this point, I'm inclined to just punt on it for now.  We will need to get this right for nng, and that may mean taking more care about designing tests that don't have these problems -- I think part of the problem is high thread counts don't work out so well there.

But for nanomsg, I'm going to just punt.

PR #925, opened 2017-10-29
add wait before closing socket in getting started

bitmage:
1. add wait before closing socket in getting started
2. add asciidoctor install and build notes to readme

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 _site
+.DS_Store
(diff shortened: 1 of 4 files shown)


gdamore:
So I don't understand the purpose of the changes in the development-bindings.html document.

I also notice -- and this isn't your fault -- that most of the docs got updated dates, but that's the only thing in them that change.  The build script is just too dumb at present and I have a fix in mind (that I'm using in nng actually) to fix that work flow.

In the meantime, I'd appreciate it if you could resubmit with *only* the README and gettingstarted/index.adoc files.

If that's too much trouble, I can cherry-pick and clean the PR myself.

bitmage:
It should be cleaned up now.  Sorry about that.

PR #934, opened 2017-11-17
CMake: Add nanomsgConfig.cmake and CMakeLists.txt for the the demos

maddouri:
Hello,

Thank you for providing a CMake-based build solution for nanomsg.

The `cmake --build . --target install` works well.
However, it doesn't provide a package configuration file which would make using the library in CMake projects easy.

This PR is an attempt at providing such a file using the guidelines from https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html

I have also took the liberty of showing how to use this new file with the demos by writing a simple `CMakeLists.txt` file.


gdamore:
This has been low on my radar for a while, partly because I'm pretty focused on the new style nng library.  I will have a closer look at this though.

One thing that stands out to me is that the current code (and your changes to it) rely pretty strongly on Travis CI-isms in our build.  I'm pretty sure I want to move away from Travis for most things (its just gotten too unreliable), which really means I need to decouple our build from Travis as much as possible.  (We are looking to move to Circle CI.)

gdamore:
I'm creating a demo/CMakefile similar, but different what you did, to make it easier to follow:

```
#
# Demonstration CMakeLists.txt for nanomsg demos.
#
# This file shows how one might use nanomsg from a another project
# that is also CMake-driven.
#
# Thanks for the idea goes to @maddouri.
#
cmake_minimum_required (VERSION 2.8.7)

project(nanomsg-demo)

# Call this from your own project's makefile.
find_package(nanomsg CONFIG REQUIRED)

add_executable(async_demo async_demo.c)
target_link_libraries(async_demo nanomsg)

add_executable(device_demo device_demo.c)
target_link_libraries(device_demo nanomsg)

add_executable(pthread_demo pthread_demo.c)
target_link_libraries(pthread_demo nanomsg)

add_executable(pubsub_demo pubsub_demo.c)
target_link_libraries(pubsub_demo nanomsg)

add_executable(rpc_demo rpc_demo.c)
target_link_libraries(rpc_demo nanomsg)
```

I think this is clearer than the macro/loop driven one you wrote.  As the purpose is to make it clear how to use CMake, I think this is the better option.  I'll be submitting a separate PR with that change.

I'm closing this one for now.

PR #949, opened 2018-02-01
export cmake target

qsodev:
export cmake target and install to `cmake/nanomsg-target.config` to enable `find_package` in other cmake based projects.

~~~
find_package(nanomsg)
...
target_link_libraries(some_target nanomsg)
~~~


gdamore:
I need to research this a bit more.  It may be appropriate to do the same thing for nng btw.

lilianmoraru:
Things like `add_definitions (-pthread)` should probably also be switched to:
```
target_compile_definitions(... PUBLIC ...)
```
and linking in this particular case?:
```
# find_package(Threads) - "Threads::Threads" not available in this version?
target_link_libraries(... PUBLIC pthread)
```

gdamore:
@lilianmoraru Huh?   I don't see any explicit additions of -pthreads in these CMakefiles, instead we use  find_package(Threads REQUIRED) which I thought was the correct way to do this for CMake.  I don't know how that is supposed to work with layered dependencies and CMake, admittedly.

qsodev:
`find_package(Threads)` together with 
~~~
if(THREADS_HAVE_PTHREAD_ARG)
    add_definitions (-pthread)
endif()
if(CMAKE_THREAD_LIBS_INIT)
    target_link_libraries (${PROJECT_NAME} "${CMAKE_THREAD_LIBS_INIT}")
endif()
~~~
as it is defined in `src/CMakeLists.txt` works correctly. I am not sure what `target_compile_options` handles differently compared to `add_definitions`. I am not a cmake expert..

But the auto generated cmake config files contain the pthread dependency. In case of Debug build:
~~~
...
# Import target "nanomsg" for configuration "Debug"
set_property(TARGET nanomsg APPEND PROPERTY IMPORTED_CONFIGURATIONS DEBUG)
set_target_properties(nanomsg PROPERTIES
  IMPORTED_LINK_INTERFACE_LIBRARIES_DEBUG "anl;rt;pthread;nsl;-lpthread"
  IMPORTED_LOCATION_DEBUG "${_IMPORT_PREFIX}/lib64/libnanomsg.so.5.1.0"
  IMPORTED_SONAME_DEBUG "libnanomsg.so.5"
  )
...
~~~
So a project using `find_package(nanomsg)` `target_link_libraries(some_target nanomsg)` will also have `-lpthread` linker option added automatically.

Is there something wrong with that? It seems correct to me.

qsodev:
`NN_STATIC_LIB` definition is now added implicitly to depending targets when using the static library

PR #953, opened 2018-02-07
update HTML and make the documentation building portable

chickenkiller:
The build.sh script does not work under linux/bash, and it seems to be difficult to make it portable between macos and linux.

So, I rewrote its logic in a ruby script, and build.sh calls it.


_adoc/build.sh
```diff
@@ -73,7 +74,7 @@ do
 	if [[ -n "$frontmatter" ]]
 	then
 		echo "---"
-		echo "$frontmatter"
+		echo -e "$frontmatter"
```
gdamore: What does -e do?  It isn't defined on macOS at least....

gdamore:
Can you please describe whatever the problem is that you think you're fixing with build.sh?

_adoc/build.sh
chickenkiller: -e enables interpretation of backslash-escape sequences, like the `\n` in the frontmatter. Under linux/bash, without this flag, the frontmatter results in such a line:

````
---
version: 1.1.2\nlayout: default
---
````

If the 'echo' command behaves differently on macos, maybe the solution is to keep the frontmatter in 2 variables and 'echo' them one after the other to avoid the `\n` escaping problem.

chickenkiller:
I rewrote my P/R entirely to use a Ruby script, which IMHO will be more portable across platforms.

gdamore:
Well, I wish you had asked me in advance.  I'm not really happy to have a ruby build script, because I don't actually *know* ruby.  As the person who will use or modify this script more often than anyone else, I can't really support this.  (Yes, I know that asciidoctor itself is a ruby program.  But I don't have to modify *that*.)

I don't want to have to learn Ruby just to update the documentation.

Building the documentation on Windows is not something that I ever need to do.

chickenkiller:
No worries, I thought afterwards that you would not like a ruby script. Forget it.

PS. I was generating on Linux, not Windows

gdamore:
Furthermore, I just ran the build script under Ubuntu Linux, no problem.  I passed it to bash.

You do need to be aware that on Debian Linux, "/bin/sh" is *not* bash, and is not ksh.  So it lacks the "[[" built-in command.  If you call it as "bash ./build.sh" it works fine.  (It turns out that "[[" is not part of the POSIX shell specification -- something I didn't realize.  Making it use "[" instead of "[[" (and also changing "]]" to "]") would solve the portability problem, although it's really a complete non-issue.

I'm closing this PR (rejecting it).

chickenkiller:
You also probably know that bash is not the default shell on MacOS, and as such, maybe it's worth using the same interpreter on all platforms. Can I suggest you these:

- switch the shebang line to use bash explicitely
- use the `-e` option for echo command, since it is mandatory for the script to run successfully under bash
- fix the typo on the `frontmatter` variable which you mispelled on line 52
- fix the `aargs` variable initialization at the loop start

I fixed all these previously (except the explicit bash invocation) because actually the current script is not functional under bash. A 'git diff' after calling it shows that the frontmatter are incorrectly formatted and that non-manpage documentation is rendered using the manpage backend which is not what you expect.

Can I suggest a P/R again so that the generation works under Linux at least?

gdamore:
I've rewritten the script to be portable.  I've also verified the results.

PR #967, opened 2018-04-18
Added resources for arm-linux-gnueabi- cross-compilation

avistel:
This introduces a directory to store cross-compilation-related resources. For now, I've added a toolchain file for ARM linux, as well as a brief how-to on setting up the build for cross-compilation and installation in the relevant toolchain folders.


gdamore:
So, CMAKE_TOOLCHAIN_FILE is fairly standard.  I don't think we need to provide this, anyone cross compiling other cmake driven software should already know how to deal with this.  (The reason not to do this ourselves is that if CMake changes in the future, or if people want to support other cross-build environments, then it gets into a very slipper support slope.)

I think its sufficient to merely state that we don't do anything magical for this,  that use of the CMAKE_TOOLCHAIN_FILE as documented on the CMake wiki is known to work.

avistel:
Fair enough. I'll withdraw.

PR #989, opened 2018-08-28
Enable version handling in exported CMake package

dennisklein:
This implements the version constraint signature of the `find_package` CMake command, e.g. one can find a minimum version or a specific version:

```cmake
find_package(nanomsg 1.1.4 [EXACT])
```

---

The bump of the minimum required CMake version might not be necessary, but I only know for sure, it will work with CMake `2.8.12+`. It might work even earlier, but the upstream git repo does not reflect any earlier versions. If anyone knows better, would be great.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -3,6 +3,7 @@
 #   Copyright (c) 2013 GoPivotal, Inc.  All rights reserved.
 #   Copyright (c) 2015-2016 Jack R. Dunaway. All rights reserved.
 #   Copyright 2016 Garrett D'Amore <garrett@damore.org>
+#   Copyright (c) 2018 Dennis Klein <d.klein@gsi.de>
 #
 #   Permission is hereby granted, free of charge, to any person obtaining a copy
 #   of this software and associated documentation files (the "Software"),
@@ -377,4 +378,28 @@ install (TARGETS ${PROJECT_NAME}
     LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
     RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
 )
-install (EXPORT ${PROJECT_NAME}-target DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/${PROJECT_NAME} FILE ${PROJECT_NAME}-config.cmake)
+
+# Generate and install CMake package
+include(CMakePackageConfigHelpers)
+set(PACKAGE_INSTALL_DESTINATION
+    ${CMAKE_INSTALL_LIBDIR}/cmake/${PROJECT_NAME}-${NN_PACKAGE_VERSION}
+)
+install(EXPORT ${PROJECT_NAME}-target
+    DESTINATION ${PACKAGE_INSTALL_DESTINATION}
+)
+write_basic_package_version_file(
+    ${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}-config-version.cmake
+    VERSION ${NN_PACKAGE_VERSION}
+    COMPATIBILITY SameMajorVersion
+)
+configure_package_config_file(
+    ${CMAKE_SOURCE_DIR}/cmake/${PROJECT_NAME}-config.cmake.in
+    ${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}-config.cmake
+    INSTALL_DESTINATION ${PACKAGE_INSTALL_DESTINATION}
+    PATH_VARS CMAKE_INSTALL_PREFIX
+)
+install(FILES
+    ${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}-config.cmake
+    ${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}-config-version.cmake
+    DESTINATION ${PACKAGE_INSTALL_DESTINATION}
+)
--- a/AUTHORS
+++ b/AUTHORS
@@ -13,6 +13,7 @@ Bruce Mitchener <bruce.mitchener@gmail.com>
 Bruno Bigras <bigras.bruno@gmail.com>
 Chip Salzenberg <chip@pobox.com>
 David Beck <dbeck@beckground.hu>
+Dennis Klein <d.klein@gsi.de>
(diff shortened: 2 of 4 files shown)


gdamore:
Thanks for this!  I'm going to take a little bit of time before I integrate it, as I want to make sure I understand what it is doing.  This is another corner of CMake that is heretofore unexplored (by me).

src/CMakeLists.txt
```diff
@@ -377,4 +378,28 @@ install (TARGETS ${PROJECT_NAME}
...
+set(PACKAGE_INSTALL_DESTINATION
+    ${CMAKE_INSTALL_LIBDIR}/cmake/${PROJECT_NAME}-${NN_PACKAGE_VERSION}
+)
+install(EXPORT ${PROJECT_NAME}-target
+    DESTINATION ${PACKAGE_INSTALL_DESTINATION}
+)
+write_basic_package_version_file(
+    ${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}-config-version.cmake
+    VERSION ${NN_PACKAGE_VERSION}
+    COMPATIBILITY AnyNewerVersion
```
gdamore: This should be SameMajorVersion instead of AnyNewVersion I think.

src/CMakeLists.txt
```diff
@@ -377,4 +378,28 @@ install (TARGETS ${PROJECT_NAME}
...
+)
+configure_package_config_file(
+    ${CMAKE_SOURCE_DIR}/cmake/${PROJECT_NAME}-config.cmake.in
+    ${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}-config.cmake
+    INSTALL_DESTINATION ${PACKAGE_INSTALL_DESTINATION}
+    PATH_VARS CMAKE_INSTALL_PREFIX
+)
+install(FILES
```
gdamore: This no longer has an EXPORT associated with it.  Is that correct?  I'm still trying to come to grips with the relationships within CMake here.

src/CMakeLists.txt
dennisklein: It is definitely your decision, what it should be, I am not familiar with the version semantics of nanomsg, but consider the following example:

The msgpack project also decided to use `SameMajorVersion` and if I want to express in my project, that I need version `2.1.5+` and that msgpack is required, I ended up writing this code:

```cmake
if(NOT msgpack_FOUND)
  find_package(msgpack 3.0.0)
  if(NOT msgpack_FOUND)
    find_package(msgpack 2.1.5 REQUIRED)
  endif()
endif()
```

You can see, that I will need to modify this code for every major release of msgpack.

While I agree with you that choosing `SameMajorVersion` is semantically more correct, I would recommend against it, because of the shortcomings of the `find_package` command on the user end.

src/CMakeLists.txt
dennisklein: Yes, the `EXPORT` property is only needed when installing targets for which an export set should be generated and installed (a file written in cmake that lists all exported targets and is meant to be imported by the consumer).

Next to the export sets, you can have these two special _top-level_ files in your CMake package. `*-config.cmake` is meant to export all sorts of information that might be useful to your user. `*-config-version.cmake` implements how  the user-supplied version (e.g. `find_package(nanomsg 1.1.4)`) is evaluated. In this PR I use the standard implementation shipped with CMake (it relies on the CMake built-in operator `VERSION_GREATER|EQUAL|LESS`). But if you wanted to implement a custom order on your versions, you could replace the call to `write_basic_package_version_file(...)` with your own file.

I tested that the default implementation from CMake works with all possible `NN_PACKAGE_VERSION`s, They can have the format `1.1.4` or `1.1.4-3-g45h5kj5` or `1.1.4-3-g45h5kj5-dirty` as far as I could see.

dennisklein:
> I think the AnyNewerVersion needs to change. 

I replied to your in-line comment, if you confirm again, that you wish this parameter to be changed, I will of course change it.

> I presume you've tested this with importing the package as well?

Yes, the code is based on what we successfully use in our projects for quite a while (on various Linux and MacOS platforms). And I specifically tested this PR on my machine (Fedora 28) when installing nanomsg to system-path `/usr/local` and to non-system-path `$HOME/nanomsg`. Both worked for me.

dennisklein:
If you are curious, you can find more comprehensive and authoritative information on [CMake package layout](https://cmake.org/cmake/help/v3.0/manual/cmake-packages.7.html#package-layout) and the [`CMakePackageConfigHelpers` module](https://cmake.org/cmake/help/v3.0/module/CMakePackageConfigHelpers.html).

gdamore:
So, nanomsg uses strict SemVer versioning, and the major number would only increase if we introduce changes that break backwards compatibility.  This is something we try pretty hard to avoid, because breaking applications is pretty undesirable.  (In fact, we don't think there will ever be a nanomsg 2.0 -- we have NNG as the replacement for the future, and nanomsg 1.x is pretty stable.)

Ultimately, if we *did* bump the major version, we would want to break automatic configuration because that's a sign that applications probably need to review their usage, because some kind of breaking change occurred.

Therefore, I will reiterate my request that we use the SameMajorVer approach. 

(Note that projects which don't follow Semantic Versioning should not use that approach.  It looks like MsgPack doesn't really follow Semantic Versioning, and has been a bit loose with breaking changes actually...)

src/CMakeLists.txt
gdamore: I stand by my request here.  MsgPack is a different project, and does not seem to follow Semantic Versioning.  We do, and are very careful to avoid bumping the major number unless we have breaking changes, which we try hard to avoid.   I believe the "shortcomings" you refer to are not shortcomings at all, but a feature.  The shortcomings are instead in the imported package, which seems not to place much value on semantic versioning.

dennisklein:
Rebased and applied requested change.

masalap:
Aww I S

Il 29 ago 2018 12:49, "Dennis Klein" <notifications@github.com> ha scritto:

Rebased and applied requested change.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/nanomsg/nanomsg/pull/989#issuecomment-416909889>, or mute
the thread

dennisklein:
Any updates here?

gdamore:
I’ll take a look at it tonight.

PR #1030, opened 2019-07-09
Mangos doc fixes

seoester:
This pull request aims to improve the godoc documentation of mangos/v2. The PR addresses nanomsg/mangos#91. It contains three changes:

 * Fix of the mangos repo URL to point to the new location.
 * Re-generation of the entire directory structure required by godoc.
 * The script with which the directory structure was generated.

I placed the script in the `./go` directory. If you would prefer a different place, please let me know.

These changes should not disrupt users working with the mangos repository as checked out by `go get`:

 * The change is practically invisible when using Go Modules. Cached modules are identified by the combination of the module path and the version. Upgrades or initial downloads lead to a lookup of the source repo using the remote import path and then a check-out.
 * Gopath users can continue to work with the checked out code.
   When running `go get -u` to update the module, `go get` shows a warning similar to this:
   ```
   package nanomsg.org/go/mangos/v2: nanomsg.org/go/mangos/v2 is a custom import path for https://github.com/nanomsg/mangos.git, but /go/src/nanomsg.org/go/mangos/v2 is checked out from https://github.com/nanomsg/mangos-v2.git
   ```
   The user has to delete `$GOPATH/nanomsg.org/go/mangos/v2` and run the command again to check out the latest version.

In addition:

 * Users without a local clone of the repository / cached module can now import any package successfully.
   Before, these users received an error message and were then required to download the repository using `go get`.
 * After merging, the complete documentation will be available on GoDoc.org at the expected URLs.
   Before, many packages were missing and lead to "Not found" messages.

## Go Modules

The remote import path behaviour is explained in the [go command's documentation][go-rip].
This is the relevant section:

> The meta tag has the form:
>
> ```html
> <meta name="go-import" content="import-prefix vcs repo-root">
> ```
>
> The import-prefix is the import path corresponding to the repository root. It must be a prefix or an exact match of the package being fetched with "go get". If it's not an exact match, another http request is made at the prefix to verify the <meta> tags match.

The open question is, whether the import-prefix should include the version postfix (`/v?`) or not. Currently, it is included for mangos.
I searched for other projects using remote import paths as well as Go Modules with a release > v1. All projects I found used an import-prefix without the version postfix. The biggest repos are:

 * https://sigs.k8s.io/kustomize/v3?go-get=1
 * https://mvdan.cc/sh/v3?go-get=1

For mangos, however, stripping the version postfix is not possible: The import-prefix is used to determine the location within the GOPATH into which the repository is checked out. Stripping the version postfix would break imports for GOPATH code.
Fortunately, "go get" seems to be robust enough to fetch the correct repo nonetheless.

[go-rip]: https://golang.org/cmd/go/#hdr-Remote_import_paths

## GOPATH: Pre Go Modules Behaviour

Code not using Go Modules yet ("GOPATH code") will check out all repositories into the GOPATH directory. "go get" will check out the mangos v2 repo to `$GOPATH/nanomsg.org/go/mangos/v2` as specified via the import-prefix.

"go get" will always check out the `master` branch. This leads to inconsistencies between Go Modules code and GOPATH code if any commit in the `master` branch is not yet released in the latest tag.

## GoDoc.org

GoDoc.org ([gddo][gddo]) does not support Go Modules yet. Remote import paths are supported via an additional `<meta>` header called [Source Code Links][gddo-rip].
The current behaviour & the changes introduced with this PR will show documentation for the `master` branch for all packages in mangos v2. I think this behaviour is "good enough".

I believe it would be possible to show documentation for multiple versions on gddo, even though support is not implemented. I have not seen this yet, but the proper documentation should be displayed by specifying an appropriate branch or tag in the Source Code Link header. This is similar to what gopkg.in does.

[gddo]: https://github.com/golang/gddo/
[gddo-rip]: https://github.com/golang/gddo/wiki/Source-Code-Links

## Going Forward

This issue is rather complicated for mangos v2. In part, this is due to using the same import path for Go Modules code and GOPATH code.
I have seen another project (don't recall which one atm) using a different approach: Go Modules code uses the proper import path with a version postfix. Developers of GOPATH code are recommended to use gopkg.in for imports. gopkg.in resolves to the latest tag with the same major version. That is the same behaviour as "go get" when using Go Modules.
This would result in the following import paths for mangos v3.

 * Go Modules code: `nanomsg.org/go/mangos/v3`
 * GOPATH code: `gopkg.in/nanomsg/mangos.v3`

I would recommend this approach. A major advantage is that GOPATH users will never check out unreleased code in the `master` branch.
The script in this PR can be used to generate the necessary directory structure for the Go Modules remote import path.


seoester:
@gdamore, what are your thoughts on this PR? Can we get this merged?

gdamore:
And yes... we can.

PR #1041, opened 2020-01-19
Update pthread_demo.c

ariels1996:
Thank you for sharing this wonderful demo file.
It was very useful for study with nanomsg.

It's nothing important, but I found some issues like this:
1. header file for ntohl, htonl.
2. Compile issue in 'int server(const char *url)' with comparison between pointer and integer.

Thank you.

### DIFF
--- a/demo/pthread_demo.c
+++ b/demo/pthread_demo.c
@@ -47,6 +47,7 @@
 #include <poll.h>
 #include <sys/time.h>
 #include <pthread.h>
+#include <arpa/inet.h>
 
 #include <nanomsg/nn.h>
 #include <nanomsg/reqrep.h>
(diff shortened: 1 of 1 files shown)


demo/pthread_demo.c
```diff
@@ -170,8 +171,8 @@ int server(const char *url)
 
     /*  Now wait on them to finish. */
     for (i = 0; i < MAXWORKERS; i++) {
-        if (tids[i] != NULL) {
-            pthread_join (tids[i], NULL);
+        if (tids[i] != '\0') {
```
gdamore: This is wrong.  NULL is the correct value here -- i.e. its a (void *)0   Not a character.

gdamore: As far as the other change (the htonl thing) -- maybe it would be better to simply eliminate the use of the macro altogether for this example.  Real programs should use endian safe operations if they want to be portable, but this example is about showing the use of threads, not about showing how to be portable.

ariels1996:
Thank you for review!!
There is no problem, so I'll close this.

PR #1055, opened 2020-11-12
add IPv6 link-local support

oengel:
Adds support for IPv6 link-local to nanomsg.
binding/connecting addresses such as 'tcp://[fe80::2%ctrl-iface]:9000' now works.

### DIFF
--- a/src/transports/utils/literal.c
+++ b/src/transports/utils/literal.c
@@ -25,12 +25,31 @@
 #include "../../utils/err.h"
 #include "../../utils/fast.h"
 
+#include <netdb.h>
 #include <string.h>
 
 #ifndef NN_HAVE_WINDOWS
 #include <arpa/inet.h>
 #include <netinet/in.h>
 #endif
+
+void nn_literal_link_local_resolve(struct in6_addr *in6addr, int64_t *sin6_scope_id, char *addr)
+{
+    if (! IN6_IS_ADDR_LINKLOCAL(in6addr)) {
+        return;
+    }
+    struct addrinfo hints, *res;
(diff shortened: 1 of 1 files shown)


gdamore:
I will review this, but be advised that this project is in sustaining mode, and I had not intended to make any new feature releases.  IPv6 link-local support (scoped addresses) represents a new feature.  New activity is occurring in the NNG project, which provides quite a lot more capabilities.

gdamore:
This looks good -- except that I'm worried it might break things for folks who don't have AF_INET6.  I know that's hard to believe in this year of 2021, but there might be older and embedded systems where folks are still running nanomsg.

Could it be conditionalized maybe?

gdamore:
> This looks good -- except that I'm worried it might break things for folks who don't have AF_INET6. I know that's hard to believe in this year of 2021, but there might be older and embedded systems where folks are still running nanomsg.
> 
> Could it be conditionalized maybe?

Actually nvm... we already had v6 there.

src/transports/utils/literal.c
```diff
@@ -61,11 +82,25 @@ int nn_literal_resolve (const char *addr, size_t addrlen,
 
     /*  Try to interpret the literal as an IPv6 address. */
     if (!ipv4only) {
-        rc = inet_pton (AF_INET6, addrz, &in6addr);
-        if (rc == 1) {
+        // check if '%' exists in addrz (i.e. '$addr%$ifname')
+        ifname_dindex = strstr(addrz, "%");
```
gdamore, line 86: strchr is more portable.

PR #1060, opened 2021-04-09
skip rpath during compilation and build.

tongyishu:
modified CMakeList.txt, skip rpath during compilation and build.


tongyishu:
new mr instead

PR #1062, opened 2021-04-22
Avoiding null pointer dereference

binho1991:
This pull request aims to fix null pointer dereference.

[Null pointer dereference](https://cwe.mitre.org/data/definitions/476.html)

### DIFF
--- a/src/core/sock.c
+++ b/src/core/sock.c
@@ -503,6 +503,9 @@ int nn_sock_add_ep (struct nn_sock *self, const struct nn_transport *transport,
 
     /*  Instantiate the endpoint. */
     ep = nn_alloc (sizeof (struct nn_ep), "endpoint");
+    if (!ep){
+        return -ENOMEM;
+    }
     rc = nn_ep_init (ep, NN_SOCK_SRC_EP, self, self->eid, transport,
         bind, addr);
     if (nn_slow (rc < 0)) {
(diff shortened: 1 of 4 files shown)


src/transports/tcp/btcp.c
```diff
@@ -113,6 +113,7 @@ int nn_btcp_create (struct nn_ep *ep)
 
     /*  Allocate the new endpoint object. */
     self = nn_alloc (sizeof (struct nn_btcp), "btcp");
+    alloc_assert (self);
```
gdamore, line 116: You should remove the one below two lines down.

binho1991: Hi @gdamore! Thanks for the feedback. I removed the assert.

src/core/sock.c
```diff
@@ -503,6 +503,9 @@ int nn_sock_add_ep (struct nn_sock *self, const struct nn_transport *transport,
 
     /*  Instantiate the endpoint. */
     ep = nn_alloc (sizeof (struct nn_ep), "endpoint");
+    if (!ep){
```
gdamore, line 506: nit: the style here does not match.  (whitespace between ) and {

binho1991: thanks @gdamore! I added whitespace!

tests/ipc.c
```diff
@@ -81,6 +81,10 @@ int main ()
     /*  Send something large enough to trigger overlapped I/O on Windows. */
     size = 10000;
     buf = malloc (size);
+    if (!buf){
```
gdamore, line 84: I'd normally *not* do this for a test case, since the panic will be sufficient to debug, and this isn't part of the library core.  But I'll leave it alone for now.

PR #1064, opened 2021-06-11
Increase time for some tests

scardracs:
These tests fails due to timeout errors on CPUs like HPPA. Tested on Gentoo Linux.


DerDakon:
This is enough for ws_async_shutdown. I managed to get the tests pass on my HP C8000 (one of the fastest HP machines) with these results:

```
1/2 Test #2: inproc_shutdown ..................   Passed   55.39 sec
2/2 Test #4: ipc_shutdown .....................   Passed   62.18 sec
```

gdamore:
it seems strange that any of the tests would need over a minute to run.

scardracs:
That's because HPPA machines are really slow

PR #1069, opened 2021-11-13
skip rpath during compilation and build

tongyishu:
skip rpath during compilation and build


gdamore:
What's the motivation for this change.  libnanomsg is kind of in sustaining mode, and I'm concerned about changes which would only work with "modern" CMake -- ostensibly if folks can use more modern versions of cmake they might also be using NNG instead of libnanomsg.

gdamore:
Looking at this, I'm disinclined to approve it. 

If you still want this, please provide an explanation of the problem you're trying to solve -- my instinct is that this would make things worse for more than it would make things better for.

PR #1123, opened 2025-07-09
cmake: Set minimum required version to 3.5 for CMake 4+ compatibility

alperak:
Fix:

| CMake Error at CMakeLists.txt:27 (cmake_minimum_required):
|   Compatibility with CMake < 3.5 has been removed from CMake.
|
|   Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
|   to tell CMake that the project requires at least <min> but has been updated
|   to work with policies introduced by <max> or earlier.
|
|   Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.
|
|
| -- Configuring incomplete, errors occurred!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,7 +24,7 @@
 #   IN THE SOFTWARE.
 #
 
-cmake_minimum_required (VERSION 2.8.12)
+cmake_minimum_required (VERSION 3.5)
 
 project (nanomsg C)
 include (CheckFunctionExists)
--- a/demo/CMakeLists.txt
+++ b/demo/CMakeLists.txt
@@ -6,7 +6,7 @@
 #
 # Thanks for the idea goes to @maddouri.
 #
-cmake_minimum_required (VERSION 2.8.7)
+cmake_minimum_required (VERSION 3.5)
 
 project(nanomsg-demo)
 


gdamore (commented):
I'm not sure about this.

While I have no qualms about using a newer CMake (CMake 3.5 is itself quite old), the concern here is that this is a legacy package that people may be using with environments where upgrading CMake is hard.

If you want a modern replacement, please have a look at NNG.

gdamore:
Actually we could do this we just add a ... Max.... I need to see if CMake 2.8 supports that syntax.

gdamore:
More details here: https://cmake.org/cmake/help/v4.0/command/cmake_minimum_required.html

We would need to verify that we can indeed operate with newer cmake.

CMakeLists.txt
```diff
@@ -24,7 +24,7 @@
 #   IN THE SOFTWARE.
 #
 
-cmake_minimum_required (VERSION 2.8.12)
+cmake_minimum_required (VERSION 3.5)
```
gdamore, line 27: This will break anyone still building on older platforms where CMake is < 3.5.

There is a new range syntax that is apparently supported ... cmake_minimum_required(VERSION 2.8.12 .. 3.5) or something.

Please do the research to make sure we don't break 2.8.12 compatibility as this is a legacy project which some people may still be using on older projects.

gdamore (changes requested):

gdamore:
I fixed this myself, in my own PR, as I ran into it myself.  Lol.

